PR #1, opened 2019-11-17
a first design test

sebastian-xyz:


jonathandturner:
Cool!  Yeah, this is a great start at putting something together.

I have a couple comments, but I wanted to say that I mean, this is already lightyears ahead of what's there now :) I'm happy working on the PR until we get it where we want it, or to take it live and use new PRs to continue to evolve it. Either way.

Onto the PR:

I noticed a few things that I think would improve what you have. One is the top bar:

Screenshot from 2019-11-18 18-57-36

The words get cut off a bit so they're harder to read.

First page looks like a good landing page. We could even link from there into parts of the book or project.

For documentation, I'd love to link to https://github.com/nushell/contributor-book and in the future also https://github.com/nushell/cookbook.

I wonder if we could make the command list more of a table.  Right now it looks like an intimidating long list.  I like the command help once you click on them. Have you thought about how you want to keep it up to date?

But yeah, great start.

sebastian-xyz:
Thanks!

I don't know what happened to that top bar but thats definitly not how it was intented to look like :)
Maybe thats caused by a different screen resolution. I'll fix that.

I'm not quite sure what you mean with the contributers-book and the cookbook.  Do you mean it should also be on the documentation page?

Yeah that's a great idea with the table because I couldn't think of a way to make natively two column page in markdown.

No I haven't figured out yet how to keep it up to date. I have to say I've never built a webpage before so maybe you've got an idea :)

jonathandturner:
> I'm not quite sure what you mean with the contributers-book and the cookbook. Do you mean it should also be on the documentation page?

Yeah, exactly. Grow into a kind of bookshelf on the documentation page to point people where they need to read.

> I have to say I've never built a webpage before so maybe you've got an idea :)

No worries, I'm not sure how we would do it either. Maybe there's a capability in jekyll. We don't have to do it now, I think we can use what we have and let someone else figure out how to do the keeping up-to-date part :)

sebastian-xyz:
For the blog we could make another page on the site or we could place a button somewhere at the top which says something like 'visit our blog' and redirects to blog.nushell.sh .

sebastian-xyz:
Any further improvement suggestions?

jonathandturner:
I think this is a good enough to land as our website, with the exception of having a blog link that doesn't work yet

Should I start up a separate repo for the blog so we can create one of those, too?

jonathandturner:
The other option would be to take the blog link out for now, put this site up. Once the blog is ready, we can put the link back on this site.

sebastian-xyz:
I think we can start to create the blog too. This shouldn't take too long and maybe during doing this there will come up some good ideas to improve the site :)

jonathandturner:
Sounds good! Here's the new repo for the blog: https://github.com/nushell/blog

PR #6, opened 2019-11-23
add feedback link and script

sebastian-xyz:
added a little python script which allows us to keep the `Quick References` easily up to date. We just need to copy the documentation of commands in the nushell repo into the commands folder of the website. If we then run the script it cycles through all the files in the folder and adds the correct Front Matter to them. It prints out the new table for the `Quick references` directly into `documentation.md`


jonathandturner:
Great!

PR #12, opened 2019-11-27
Add anchor links to headings, tidy up HTML and Sass

Aloso:
- When a heading is hovered, a 🔗 symbol appears. Click on it to link to the heading
- When an internal link (e.g. https://www.nushell.sh/installation.html#docker-containers) is clicked, the browser now scrolls to the correct position (previously it scrolled ~60px too far)
- Fix a few HTML issues
- Refactor CSS, convert some colors into Sass variables
- Fix broken link

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ _site
 .jekyll-cache
 .jekyll-metadata
 vendor
+*.iml
+.idea/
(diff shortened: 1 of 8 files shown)


jonathandturner:
These are cool, thanks!

PR #13, opened 2019-11-28
Fix variable width characters in `pre`

### LINKED ISSUE #11: Better tables in examples
I noticed some people talking about how tables looked in different browsers on Reddit, and then when I checked nushell.sh I noticed a similar thing.  Here's what I see in Firefox+Windows:

image

I wonder if there's a way to either render this the same across browsers, or if maybe we should just put an image there instead.

Aloso:
This PR adds some JS which iterates over `<code>` blocks within a `<pre>` and replaces all horizontal line characters (`━`, `─`) with a `<u>` elment containing the character.

This element has a CSS rule `width: 1ch` which gives it the width of the `0` character. Unfortunately, this only works in browsers with JS enabled. Furthermore, the `ch` unit is not supported in IE <= 8, Safari <= 6.1, Android Browser <= 4.3 and Opera Mini. In these browsers, this fix has no effect. However, I think this is acceptable, since these browsers are rarely used.

Fixes #11

### DIFF
--- a/_includes/script.html
+++ b/_includes/script.html
@@ -1,8 +1,13 @@
 <script type="text/javascript">
     $('h1, h2, h3, h4, h5, h6').filter('[id]').each(function () {
-        var el = document.createElement('a')
-        el.className = 'anchor'
+        var el = document.createElement('a');
+        el.className = 'anchor';
         el.href = '#' + this.id;
         this.prepend(el)
     });
+    $('pre code').each(function () {
+        this.innerHTML = this.innerHTML
+            .split('').map(c => c === '━' || c === '─' ? `<span>${c}</span>` : c)
+            .join('');
+    })
 </script>
--- a/_includes/script.html
+++ b/_includes/script.html
@@ -6,7 +6,7 @@
         this.prepend(el)
     });
     $('pre code').each(function () {
-        this.innerHTML = this.innerHTML.replace(/\u2501+|\u2500+|\u252f|\u253c|\u2537|\u2502/g, function (s) {
+        this.innerHTML = this.innerHTML.replace(/\u2501{1,8}|\u2500{1,8}|\u252f|\u253c|\u2537|\u2502/g, function (s) {
             return '<u style="width:' + s.length + 'ch">' + s + '</u>'
         });
     })
--- a/_sass/jekyll-theme-midnight.scss
+++ b/_sass/jekyll-theme-midnight.scss
@@ -150,7 +150,7 @@ pre {
   overflow: auto;
   overflow-y: hidden;
   font-size: 80%;
-  line-height: 1.35em;
+  line-height: 1.3em;
 
   code {
     color: #efefef;
@@ -161,9 +161,12 @@ pre {
     padding: 0;
   }
 
+  // table borders:
   u {
     text-decoration: none;
     width: 1ch;
+    height: 1em;
+    color: white;
     overflow: visible;
     display: inline-block;
   }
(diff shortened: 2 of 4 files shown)


waldyrious:
> with a `<u>` elment

You seem to be actually using a <span> in the code. Maybe the description needs to be updated?

_includes/script.html
```diff
@@ -1,8 +1,13 @@
...
-        var el = document.createElement('a')
-        el.className = 'anchor'
+        var el = document.createElement('a');
+        el.className = 'anchor';
         el.href = '#' + this.id;
         this.prepend(el)
     });
+    $('pre code').each(function () {
+        this.innerHTML = this.innerHTML
+            .split('').map(c => c === '━' || c === '─' ? `<span>${c}</span>` : c)
```
waldyrious: I'm not sure this will work for lines containing lots of ┷/┯, etc.

It seems to me that the right solution would be to use a font where these characters have the correct width.

_sass/jekyll-theme-midnight.scss
```diff
@@ -160,6 +160,13 @@ pre {
     margin: 0;
     padding: 0;
   }
+
+  u {
```
waldyrious: In the code above you're using a `<span>`, not a `<u>`. Surely you meant the former, right?

Also, note that the PR description also says `<u>` instead of `<span>`.

Aloso: Thanks, I missed this. It should be `'<u>' + c + '</u>'`.

Also I just noticed that I used a tempalte string. I've gotten too used to TypeScript :D

_includes/script.html
Aloso: `┯ ┷ ┼` have the same width as `│`, so it still aligns properly. I'd like to keep the number of elements low.

Aloso: > It seems to me that the right solution would be to use a font where these characters have the correct width.

I tried several monospaced fonts, and their box-drawing characters were always rendered too wide. This doesn't seem to be the font's fault. I verified that Fira Code includes box-drawing characters, and that they have the same width as other characters. Yet they are rendered too wide on Android. I didn't try it on Windows.

Aloso:
I fixed the issues (I hope). I amended and force pushed, because I made a mistake, but it should be fine now.

Aloso:
I updated the terminal colors. Coincidentally, this PR also allowed me to fix the gray lines after a `#`:

Before:
Old terminal colors

After:
Better terminal colors

_includes/script.html
waldyrious: That clearly looks like Android is overriding the font definition and replacing it with a system font. How is the Source Code Pro font being provided? I noticed that the OpenSans definitions in [_sass/fonts.scss](https://github.com/nushell/nushell.github.io/blob/master/_sass/fonts.scss) point to source files in a `fonts/` directory, but that directory doesn't exist. Maybe we need to actually include the font files in the repo so that they are actually used when they're not immediately available in the system?

Aloso: I tried it even with Google fonts, so I'm pretty sure that the font was loaded.

The fonts directory is in `_site/assets`. I don't know how that was generated, since `_site` is in `.gitignore`, but I'm not that familiar with jekyll.

jonathandturner:
Sounds good. Thanks for that.

jonathandturner:
Now that I was able to try this on Firefox on a laptop and my Android phone, it seems like the combination of both PRs may not be giving us quite what we expected.

For now, I'm going to revert this PR and we can give it another try.

waldyrious:
@jonathandturner did you check the comments in [this thread](https://github.com/nushell/nushell.github.io/pull/13#discussion_r351905071)? It could be that the issue is that the fonts are not being properly included in the built site. From @Aloso's last comment there, it seemed like this PR was not yet ready to merge anyway :)

PR #14, opened 2019-11-28
Adjust for mobile devices

Aloso:
This adjusts font sizes on small screens. It makes sure the header fits in 1 line on screens that are at least 360 CSS-pixels wide.


jonathandturner:
I'm wishing we had an easy way to test this before landing, but it looks good so I'll go ahead and land.

PR #17, opened 2019-11-29
Add monospaced font to align tables, fix table header color

Aloso:
Finally fixes the table border problem by adding `Source Code Pro` as a webfont.

Additionally, it fixes the [gray table headers](http://www.nushell.sh/commands/add.html).


_sass/fonts.scss
```diff
@@ -8,7 +32,6 @@
          url('../fonts/OpenSans-Light-webfont.svg#OpenSansLight') format('svg');
```
waldyrious: Don't these font files also need to be added to the `assets/fonts` directory?

_includes/script.html
```diff
@@ -1,8 +1,11 @@
...
-        var el = document.createElement('a')
-        el.className = 'anchor'
+        var el = document.createElement('a');
+        el.className = 'anchor';
         el.href = '#' + this.id;
-        this.prepend(el)
+        this.prepend(el);
     });
+    $('pre code').each(function () {
+        this.innerHTML = this.innerHTML.replace(/\u2502/g, '<u>\u2502</u>');
```
waldyrious: I'm not sure I understand why you're using an `<u>` element instead of a `<span>`. The latter makes more semantic sense IIUC, and using `<u>` forces you to override the underlining with `text-decoration: none;` in the CSS. What's the rationale? Is it just because it's shorter?

_sass/jekyll-theme-midnight.scss
```diff
@@ -88,7 +88,7 @@ a.anchor {
   }
 
   &::before {
-    content: '🔗';
+    content: '\01F517';
```
waldyrious: Maybe add a comment mentioning this character's name, to make the code more self-explanatory. 🙂

_sass/fonts.scss
Aloso: They are automatically added during the build. I don't know why, though. I can add them if it doesn't work on your machine.

waldyrious: Well, you're adding the Source Code Pro files to that folder in this PR, so it would make sense for these to be there as well, if only for consistency.

_includes/script.html
Aloso: Because it's shorter, but also because `<span>` is already used within `<code>` for syntax highlighting. `<u>` has no _semantic_ meaning in HTML5, so it shouldn't be a problem.

_sass/fonts.scss
Aloso: Done!

sebastian-xyz:
The table formatting looks good now on Windows and Linux.

@Aloso Did you test the new font on a mobile?

On my device the table headers are still gray but since this wasn't our main problem I think we can leave it for now.
table

_includes/script.html
```diff
@@ -1,8 +1,12 @@
...
+        el.className = 'anchor';
         el.href = '#' + this.id;
-        this.prepend(el)
+        this.prepend(el);
     });
-</script>
\ No newline at end of file
+    $('pre code').each(function () {
+        // Box Drawings Light Vertical  https://unicode-table.com/en/2502/
+	this.innerHTML = this.innerHTML.replace(/\u2502/g, '<u>\u2502</u>');
```
waldyrious: Heads-up: this line is indented with a tab rather than spaces 🤓

sebastian-xyz:
Sorry I was on the wrong branch after updating my local files :)

Table headers look good!

We should be able to merge

jonathandturner:
Looks great on Android!

PR #27, opened 2020-01-17
Fix logo font

waldyrious:
Same change as what was done in 076d63e2d for the `<code>` elements, now applied for the logo as well.

Here's what the website looks for me on Firefox 72.0 on macOS 10.15.2 (Catalina):

Before:

Screenshot 2020-01-17 at 17 51 06

After:

> Screenshot 2020-01-17 at 17 51 24


sebastian-xyz:
Thanks!

PR #28, opened 2020-01-18
Improve installation instructions

waldyrious:
- Various wording, punctuation and formatting tweaks
- Add a separate section for Homebrew, since it works for both macOS and Linux
- Break some long lines, especially those with links, to allow for cleaner diffs

PR #30, opened 2020-01-26
Update cargo flags

### LINKED ISSUE #29: Cargo flags on installation page are wrong
On the installation page it says to run `cargo install nu --all --features=stable` for an "all bells and whistles" install. But it seems there is no `--all` flag in cargo. I'm new to Rust so not sure if this should be `--all-features` or `--features=stable` which both seem valid.
EDIT: I have copied the installation instructions from the book

nespera:
I changed them to match what is in the book. Should fix #29

### DIFF
--- a/installation.md
+++ b/installation.md
@@ -135,7 +135,7 @@ From there, we can build and run Nu if we are using `rustup` with:
 
 ```
 > cd nushell
-nushell> cargo build --all-features && cargo run --all-features
+nushell> cargo build --workspace --features=stable && cargo run --features=stable
 ```
 
 You can also build and run Nu in release mode:
--- a/installation.md
+++ b/installation.md
@@ -118,7 +118,7 @@ You can also install Nu with all the bells and whistles (be sure to have install
 [dependencies](https://book.nushell.sh/en/installation#dependencies) for your platform):
 
 ```
-cargo install nu --all --features=stable
+cargo install nu --features=stable
 ```
 
 ## Building from source
@@ -135,7 +135,7 @@ From there, we can build and run Nu if we are using `rustup` with:
 
 ```
 > cd nushell
-nushell> cargo build --all --features=stable && cargo run --all --features=stable
+nushell> cargo build --all-features && cargo run --all-features
 ```
 
 You can also build and run Nu in release mode:


waldyrious:
Cross-referencing [related discussion](https://github.com/nushell/nushell.github.io/pull/28#discussion_r368233299). Note that any changes to the flags should be reflected in the [main repo's README](https://github.com/nushell/nushell#local) as well.

jonathandturner:
I believe the line we should be using going forward is:

```
cargo build --workspace --features=stable
```

nespera:
I don't know much about cargo, but it looks like in the previous change the `--all` flag was added to `cargo install` which does not support it, unlike `cargo build`, which does.

installation.md
```diff
@@ -135,7 +135,7 @@ From there, we can build and run Nu if we are using `rustup` with:
 
 ```
 > cd nushell
-nushell> cargo build --all --features=stable && cargo run --all --features=stable
+nushell> cargo build --all-features && cargo run --all-features
```
andrasio: This won't work correctly. Basically `--all-features` also adds the tracer which we need to fix. For `cargo build` and `cargo run` we better leave them as `--all --features=stable`

installation.md
```diff
@@ -118,7 +118,7 @@ You can also install Nu with all the bells and whistles (be sure to have install
 [dependencies](https://book.nushell.sh/en/installation#dependencies) for your platform):
 
 ```
-cargo install nu --all --features=stable
+cargo install nu --features=stable
```
andrasio: Good.

installation.md
nespera: --all is not supported by `cargo run` as far as I know

andrasio: omg, you are right. Then we may want to have `cargo run --features=stable` (not `--all-features`)

andrasio: For the case of `cargo build` I also agree with @jonathandturner 's comment.

nespera: Ok, updated to use @jonathandturner 's suggestion for `cargo build` and yours for `cargo run`

sebastian-xyz:
Thank you!

PR #36, opened 2020-02-01
more information on main page

sebastian-xyz:
PR adds a few improvements to the main page including installation instructions and book references.

Please take a look at it and let me know if you have ideas for further improvements.

I'm not sure if the backgroundcolor for the instalation instructions are a good choice. There are probably better choices. :)


jonathandturner:
I had a few ideas looking through this.  I think the direction you're going is good: we should make it easier to see why you'd want to use Nu from the start, and having some nice examples right at the top looks like a good way to do that.

A couple things I was wondering how you felt:
* What if the bottom right circle had two words, one on top of the other, so that it felt a bit better. Could use a larger font also
* For the text above the image, I wonder if we should left-align it
* Seeing this new part makes me wonder if we should take the material after the installation instructinos and put it on a separate page. What do you think?

sebastian-xyz:
Yes, we could move those parts after the information section to about. I agree that this would give the main page a clear structure. At the moment there isn't much on the about page and it seems to me that this would be the right place to put it. 
To do this we could move those references to Discord, Twitter und GitHub in another section on the main page. This would also make it a lot easier for people to get in touch with us.
I personally would leave the reference to the book on the main page.

Should we in addition to that remove the green background of the installation section?

Issue #37, opened 2020-02-09
Have installation instructions on the front page


thegedge:
I think having installation instructions directly on the front page would be ideal. Every extra step to do something often lowers conversion.

sebastian-xyz:
@thegedge I like the idea but I wonder how we should implement this. If we add all the possible ways to install nu we will end up with a huge page and people probably won't find the things they are looking for.

thegedge:
I'm thinking one of two approaches:

1. Try to guess the operating system and offer options specifically for that.
2. Have all the options in some kind of tabbed view.

I think (1) is pretty common these days. Either way, it should always be clear how to get to the extensive installation instructions in case nothing works. We definitely do not want anyone feeling like they're at a dead end with no options 🙂

ibraheemdev:
I think this can be closed now that the site has been updated.

PR #41, opened 2020-03-05
add os-specific installation instructions to front page

sebastian-xyz:
PR adds os detection to determine which installation instructions to display on the front page. #37


jonathandturner:
Nice! This is very cool!

PR #43, opened 2020-03-13
Sync command docs with Nu 0.11.0 release

waldyrious:
Similar to #31. Done by running the following commands:

```
cd ~/repos/nushell/ && git checkout 0.11.0
cd ~/repos/nushell.github.io/commands/ && rm -f *.md
cp -rf ~/repos/nushell/docs/commands/* .
python3 update_commands.py
```

@sebastian-xyz thanks for adding this script in #6! It would be nice if these instructions were in the README of this repo (I had to figure them out by digging through the git history of the commands/*.md files), or even better, if they could be incorporated into the Python script 😃 (for example, it could download the nushell sources rather than require manually copying the files over)

Also, it probably would make sense to include the nushell version in the metadata blocks at the top of the md files (and maybe in the rendered view as well). WDYT?


jonathandturner:
Thanks!

sebastian-xyz:
@waldyrious I'm really sorry for the bad documentation.
Downloading the latest documentation is definitely something we could add to the script. I just wonder what the best way would be to do this. It seems like there is a way to only download a specific folder from GitHub (https://github.community/t5/How-to-use-Git-and-GitHub/How-can-I-download-a-specific-folder-from-a-GitHub-repo/m-p/88#M20).

Yes, adding the version to the metadata and the rendered version definitely makes sense so we can check if the documentation is up to date.

Thank you for your good ideas! 
Do you want to implement it? 🙂

waldyrious:
I'm glad you agree with these ideas :) unfortunately I don't have the availability to work on them myself at the moment, but I'd be happy to contribute instructions for using the update script so that it's more discoverable by others. Let me know if I should do that right away (using the steps I outlined above) or if it would be best to do it after the script is updated.

sebastian-xyz:
I think we can leave it for now. There are 3 more weeks to the next release. Should be possible to update the script before we need to update the instructions :)

PR #44, opened 2020-03-23
update script

sebastian-xyz:
PR updates `update_commands` script. It now automatically fetches the latest command documentation from the main repo and performs all necessary steps to update the command list on the website.


jonathandturner:
Looks great!

waldyrious:
@sebastian-xyz what are the usage instructions now? What is different from the sequence I wrote [here](https://github.com/nushell/nushell.github.io/pull/43#issue-387921890)?

sebastian-xyz:
The script does now most of the things automatically. We just need to run `python update_commands.py` (python version 3). The script then waits for the user input of the current nu version (we could probably fetch the version as well in the future). After that it pulls the documentation automatically and adds the metadata.

Two things are important:
- script needs to be run in the `commands` directory
- check Internet connection before running the script (although there shouldn't happen anything if there's no connection but you never know ;) )

So the instructions would be:
```
> python update_commands.py
enter version: 0.11
...
>
```

PR #50, opened 2020-06-15
Remove leftover dash in the `from` and `to` subcommands

kubouch:
Simple removal of leftover dashes in the Documentation for `from` and `to` subcommands in order to address [issue #1864 from the main nushell repository](https://github.com/nushell/nushell/issues/1864).

This is the simplest solution. I think the preferred solution would be to remove the subcommands from the list completely and add * to `from` and `to` commands to indicate that these commands have subcommands.


jonathandturner:
Good catch

PR #52, opened 2020-07-13
Basically just run the update_commands.py

arashout:
Had to change a line that referred to "master" into "main"

https://github.com/nushell/nushell.github.io/pull/52/files#diff-29024de97217e256465bd5c17ed73b67L34


jonathandturner:
Oh good, thanks for doing that

PR #54, opened 2020-07-17
Use auto-generated docs and show usage information in frame layout

arashout:
Summary:
- Use yaml file inside `_data` directory to display documentation for the commands
- The documentation is displayed using a frame layout with a scrolling layout

It's still not perfect... but I think it's an improvement 
Future work could be to use categorization instead of just listing all the commands as @kubouch mentioned on discord
ezgif com-video-to-gif


documentation.html
```diff
@@ -0,0 +1,169 @@
...
+                {% endif %}
+            </div>
+            <div id="{{ command.name }}-subs" class="hidden">
+                <ul>
+                    {% for sub_command in command.subcommands %}
+                    <li class="row">
+                        <button class="row__item button grow"
+                            onclick="toggleFrame('{{ sub_command.name | replace: ' ', '_' }}-usage')">{{ sub_command.name }}</button>
+                    </li>
+                    {% endfor %}
```
thegedge, line 125: I think subcommands should just be shown on the same level as the other commands, without having to click an "expand" button to see them. I'm thinking the fewer clicks one has to make, the better. It also will be easier to work with if subcommands ever have subcommands, which have subcommands... 😅 

What do you think?

arashout: Yeah, that makes sense. There's just a lot of sub-commands.
But I'm good with either and can change it

thegedge: > There's just a lot of sub-commands.

Yeah let's keep it simple for now and not hide them by default, and return to this decision if things get out of hand.

arashout:
Cool, I don't mind doing more work on the website, if I can work off your wireframes.

PR #59, opened 2020-09-02
Update commands to 0.18.1

gillespiecd:
Update commands to 0.18.1, ran the `update_commands.py` script


jonathandturner:
Great! Thanks for doing that.

PR #66, opened 2020-11-16
Add cname to deploy action

ibraheemdev:


ibraheemdev:
Looks like deployment is fine without this

PR #67, opened 2020-11-16
Replace splitting tag in git log

janosimas:
Replacing the (nu) tag allows removing the quotes from the formatting string.
The new command works in both linux and windows systems.
Discussion in https://github.com/nushell/nushell/issues/2730


jonathandturner:
Thanks!

PR #78, opened 2020-12-28
Add groups docs from @jturner

max-sixty:
As discussed on Discourse.

This is imperfect, but on net better than previous, and enough small improvements gets to perfect...


jonathandturner:
Thanks!

PR #82, opened 2021-01-04
Update example

eoinkelly:
Update the example to use `to json` instead of `to-json`. Also clarify that `clip` is a Windows command and add the macOS equivalent.

### DIFF
--- a/cookbook/system.md
+++ b/cookbook/system.md
@@ -114,7 +114,13 @@ Notes:
 Add the output of a pipeline to your clipboard.
 Note, this currently needs to be string output.
 
-`sys | get mem | to-json | clip`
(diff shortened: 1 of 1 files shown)


jonathandturner:
`clip` should work with macOS too, if you install Nushell with the clip feature. Are you seeing issues with it?

eoinkelly:
Ah ok I understand now. I was using the version from homebrew which doesn't include the `extra` set of features. Once I installed from cargo with `--features=extra` then `clip` works as expected.

I think there is still a speed bump in the docs to resolve because many (maybe most?) folks getting started won't install from cargo with `--features=extra` and will get confused.

I'll amend the change to reflect this.

jonathandturner:
Thanks!

PR #86, opened 2021-01-27
Add command reference

ibraheemdev:
The command reference that was on the front page of the old nushell site is now a chapter of the book


jonathandturner:
Awesome! Thanks for putting this together!

PR #103, opened 2021-02-23
Move PowerShell command listed in Bash column

JTurtl3:
`Sort-Object {Get-Random}` was mistakenly listed as a Bash command. It has been moved to the PowerShell column.


jonathandturner:
Thanks!

PR #107, opened 2021-03-06
Update environment.md

stormasm:
This will make it clearer to people who are familiar with bash as well as making it clear for folks who do not understand the word scope in this context.


jonathandturner:
Thanks!

PR #108, opened 2021-03-08
clear up docker installation instructions

jakevossen5:
I made the following changes to the Docker installation instructions.

- Link to Quay.io
- Move the difference of "nu" and "nu-base" to before the run commands
- "dockerfiles provided" -> "provided dockerfiles"
- Changed "Multistage build" to match capitalization I found in this blog post: https://www.docker.com/blog/multi-stage-builds/
- I changed the order of the `pull_prebuilt_container.sh` to match the other instructions on the paragraph (everything else first does an example for "nu-base", then "nu", and now that matches)

I realize a lot of documentation is just personal preferences / styles, so feel free to reject some / all of these changes! I just saw some things that I would have done differently.


jonathandturner:
thanks!

PR #111, opened 2021-03-16
Add list of accepted parameter types

kubouch:
Currently, it seems this information is only available [in the source code](https://github.com/nushell/nushell/blob/56adc7c3c6c8c011c2db9cdcf780d9213a925a8b/crates/nu-parser/src/parse/def/primitives.rs#L162-L185).

The mentioned version 0.28.0 will get obsolete soon but I'd guess better than nothing, unless there is some mechanism to keep this up to date automatically.


jonathandturner:
Thanks!

PR #119, opened 2021-04-10
adds "Working with lists" page to Book

mvolkmann:


jonathandturner:
Great!

mvolkmann:
@jonathandturner Does it seem like my change to `table_of_contents.md` didn't work? I see the new section in the left nav, but not the table of contents.

jonathandturner:
@mvolkmann you may also need to update https://github.com/nushell/nushell.github.io/blob/master/book/README.md, I don't remember off-hand

mvolkmann:
Okay, I'll look in to that. Thanks!

mvolkmann:
It looks like you are correct about the need to also update the `README.md` file. I'll fix that in a new PR. I'd also like to fix the inconsistencies between that file and `table_of_contents.md`. The biggest difference is whether it refers to "nushell" or "Nu". I was thinking that "nushell" is the shell and "Nu" is the scripting language provided by nushell. But maybe your intention is to switch to just saying "Nu" everywhere.

As an example of a difference between these files, for the "Installation" link one says "Installing nushell" and the other says "Installing Nu". What is your preference?

mvolkmann:
@jonathandturner I created a new PR to make these files match.

PR #120, opened 2021-04-13
modified README.md and table_of_contents.md to match

mvolkmann:
I'm not sure while both README.md and table_of_contents.md are needed, but it seems that they should match. This PR changes them so they do.


jonathandturner:
Let's give it a go

PR #123, opened 2021-04-20
update ls -f -> ls -la

ammkrn:
The "coming from bash" page for all three languages (en/es/ja) shows `ls -f` as the nu equivalent of the vanilla `la -la`. As of 0.29.0, `ls -f` doesn't seem to exist, but `ls -la` gets you both the long form and shows hidden files.


jonathandturner:
Thanks!

PR #125, opened 2021-04-23
Update starship prompt integration

ChristopherBiscardi:
`STARSHIP_SHELL` seems to be the magic key to getting rid of the garbled output when using starship prompt in some situations. Add it to the docs so that it is findable by users


jonathandturner:
Thanks!

PR #126, opened 2021-04-28
Fix Dead Links

nalpine:
Found a couple of dead links that I believe were pointing towards old pages.  Updated them to point towards current locations of page contents.


jonathandturner:
Thanks!

PR #130, opened 2021-05-22
replace `date utc` with `date now` in introduction

henriiik:
Hello!

I was going through the introduction and the `date utc` command did not work for me. After some digging i found that the command had been removed in this PR:

https://github.com/nushell/nushell/pull/2780

So this PR updates the introduction page to use `date now` and `date to-table`


jonathandturner:
Thanks!

PR #131, opened 2021-05-23
update update `sys_get_` examples

henriiik:
Hello! 

The type of sys.host.sessions was changed from a string to a table in this
PR: https://github.com/nushell/nushell/pull/2954, this PR updates the examples in the introduction to reflect that. :)


jonathandturner:
Thanks!

PR #139, opened 2021-05-30
Document $() syntax for subcommands instead of ()

IgorMinar:
This looks like a leftover from the past.

As far as I can tell subcommands should use `$()` rather than `()` syntax.

### DIFF
--- a/book/variables_and_subexpressions.md
+++ b/book/variables_and_subexpressions.md
@@ -31,14 +31,14 @@ testuser
 
 ## Subexpressions
 
-You can always evaluate a subexpression and use its result by wrapping the expression with parentheses `()`.
+You can always evaluate a subexpression and use its result by wrapping the expression with dollar-prefixed parentheses `$()`.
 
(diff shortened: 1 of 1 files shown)


jonathandturner:
@IgorMinar we've updated the docs with the syntax coming in the release in two days, so we'll be out of sync, but should be back in sync once 0.32 gets released.

jonathandturner:
I'll update the docs with a note saying that they're updated for 0.32

IgorMinar:
aha, I see! I'll close this PR then.

IgorMinar:
Funny, I also spend quite some time scratching my head about how to enter the math mode and found that `=` was no longer documented but still needed in 0.31. I see that you updated those docs recently as well: https://github.com/nushell/nushell.github.io/commit/514f98848fda686e0e8a6b6588874ad97d1c73d1#diff-7face5f233216368ef734317c1c2d8bfb2f280418db70385abf660ab1a751f48

jonathandturner:
@IgorMinar yeah sorry for the confusion. In hindsight, we could have waited to land the PR to closer to the release

IgorMinar:
@jonathandturner np, I do like what you did in https://github.com/nushell/nushell/pull/3389 👍🏻

PR #140, opened 2021-06-01
0.32 blog

jonathandturner:


fdncred:
hey @jonathandturner  - here are some notes that took while reading this. take it or leave it - no hurt feelings if you don't agree with these changes. :) good job typing all this up. i like the use of the twin script to capture the screenshots, etc... hehe.

### notes for 0.32 release

i think more should be said about `(ls).name`. it's talked about a little bit at the beginning of the blog, but for people who aren't familiar with "ColumnPaths" they may be lost as to what that means when you mention column paths.

For the `Environment loading` section, it may be good to add more "name, value" entries in the table just to codify that there can be tons of name/value pairs. perhaps echo the table in the activate def, again just to demonstrate we made a table? I think it would connect some dots that this can be more than one entry and that the table literal actually makes a table which is really what is being used. As well as saying something about how the table columns have to be named "name" and "value".

The one thing that is missing in `environment loading` is `environment unloading`. it might be good to show how to store the env, modify it with `load-env`, work in it blah blah blah, then restore the environment to the state before you started this work. I guess that's possible?

after the `Locale improvements (fdncred)` section there's a sentence about `load-env`. Looks like it needs to be moved up to be part of the `load-env` section. it starts with "The changes to the environment are still scoped..."

it may be good to mention the rustyline changes that allow alt-backspace to work now, as well as all better modifier (ctrl/cmd, alt/option, shift) support. previously we couldn't do that. one of the things that people traditionally expression frustration with is keybindings. i think it works so much better now and should be fairly intuitive given the `docs/sample_config/keybindings.yml` sample file in the repo.

note - not for blog - we need to make `delete to trashcan` the default on all platforms. this was part of the feedback we received recently in the reddit post that i summarized with a google sheet. we should revisit that to see what we're still missing at some point.

in the `char` section, it would be good to mention that `char --list` now shows all the "special" characters that can be used in the char command. we should do this same `--list` for the `ansi` command too sometime.

link to the nu_scripts repo for examples of the new syntax

link to the youtube video for the release

PR #148, opened 2021-06-20
Remove config load since it doesnt exist

elferherrera:
Removes load file option since it doesn't exist


jonathandturner:
Thanks!

PR #150, opened 2021-06-22
Suggested layout

elferherrera:
With this layout the first page the user sees is the introduction and not the table of content for both the book and the contribution book


jonathandturner:
I like this! I think that's a much better way to jump into the book

PR #151, opened 2021-06-24
Fix missing value types for completion_prompt_limit and true_color

autophagy:
Hi! :sparkles: 

While going through the configuration documentation, I noticed that 2 configuration options were missing a value type annotation: `completion_prompt_limit` and `true_color`. This PR adds them in to be consistent with the other config option documentation.


jonathandturner:
Thanks!

PR #152, opened 2021-06-25
docs(configuration): show `completion_match_method` under `line_editor`

eduardocanellas:
As noted on nushell/nushell#1843, this option shouldn't be on root. I discovered it while I was configuring my shell :laughing: I decided to move from fish and the case-insensitive option feels much better.


jonathandturner:
Thanks!

PR #158, opened 2021-07-05
Main page description changes.

andrasio:
Mostly translation to spanish's main site. 

The first feature description in english feels uncomfortable:

```
Control your system with powerful pipelines

Pipelines let you command your system like never before. Your system belongs to you, and it awaits your command.
```

Rewrote it. Happy to rollback or to change it to something else, please advise. @jonathandturner

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ tagline: A new type of shell.
 actionText: Get Started →
 actionLink: /book/
 features:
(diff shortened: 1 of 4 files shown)


jonathandturner:
@andrasio - it's better. still not quite sure what we're trying to say here, but I think we can just keep improving it

PR #160, opened 2021-07-12
Match `where` clause with table output

phaunt:
The example had "where cpu > 10" in the command line, whereas the two rows in the output both had 5 < cpu < 10


jntrnr:
Thanks!

PR #163, opened 2021-07-14
Fix typo in 2021-07-13-nushell_0_34.md

syndek:


syndek:
Looks like this is already fixed [here](https://github.com/nushell/nushell.github.io/pull/162), never mind!

PR #166, opened 2021-07-15
fix(plugin): missing branch

dyxushuai:
`nushell` had already used the `main` as the default branch instead of `master`.


jntrnr:
Thanks!

PR #171, opened 2021-07-20
Added some clarifying statements about string interpolation

fhalim:


jntrnr:
Looks good - thanks :)

PR #173, opened 2021-07-20
there's no values column anymore

zkat:
I guess this got removed a while back!


jntrnr:
Wowww... yeah a *long* time ago. Nice catch :)

PR #182, opened 2021-08-05
Update plugin discovery section of the doc

hedonihilist:
The actual plugin discovery approach has changed and thus the document should be updated accordingly.


jntrnr:
Thanks!

PR #187, opened 2021-08-10
add doc for `filesize_metric` config

hedonihilist:
It took me quite some time to figure out how to config `Nu` to display file sizes like `ls -lh` does (automatic switch unit in power of 1024 based on how large the file is). So I think it's quite necessary to document the `filesize_metric` config entry.


jntrnr:
Thanks!

PR #190, opened 2021-08-26
Document removing environment variables

filaretov:
* Describe `unlet-env`

* Describe `let-env ENVVAR = $nothing`

* Add `let-env` as example of `export` and `unset` in Bash table

I could only update the Bash tables in the translated books.


filaretov:
I rebased on the main branch and force pushed, the tables in the Japanese and Spanish Bash tables had some merge conflicts (I also just noticed some old `alias` syntax in those, I'll update them in a new PR after this one is resolved).

jntrnr:
Looks good!

PR #193, opened 2021-09-08
remove additional backslash

tw4452852:
It seems github's markdown parser have different behavior for the backslash in code block than the website's:

github's default parser works fine:
screenshot-20210908@152412

but the website shows the additional backslash in the table:

screenshot-20210908@152056

Because the website is our target eventually, so I remove the additional backslash.


jntrnr:
thanks

PR #194, opened 2021-09-18
Add document for `g` command

tw4452852:
Add `g` command for better shell switch.

nushell/nushell#4014

Signed-off-by: Tw <wei.tan@intel.com>


andrasio:
Thank you!

PR #198, opened 2021-09-24
Fix git rm merged

aminya:
- `$it` is not defined when I ran the previous command. Added each to fix it
- added main branch
- added trim


jntrnr:
thanks!

PR #202, opened 2021-10-03
Fix typo in "config set" instructions

thomasaarholt:
I was trying to set configs in nu 0.37 and noticed this error. I can't check, but I suspect (?) the instructions were meant to be as PRed for 0.7.2 (and that these changes haven't been introduced later)?


thomasaarholt:
Edit, I'm now no longer sure if the `--set` -> `set` change was introduced in 0.7.2. 

Regardless, it would be nice if this was added to the main docs :) I'm currently trying to work out how to (elegantly) edit my path to remove an incorrect path, and I'm not sure how to do that.

PR #208, opened 2021-10-19
Add winget as installation method

rashil2000:
Nu is also available on Scoop and Chocolatey, but I've only included Winget for now.


rashil2000:
Bump?

jntrnr:
thanks!

PR #213, opened 2021-12-13
Show link to Groups setion

kyoheiu:
I was looking at this page and was confused by this line cause I couldn't search "Command groups" page neither by "Command" nor by "group".

After that I came across https://www.nushell.sh/book/types_of_data.html#groups, and thought it would be useful to show the link instead of the ambiguous expression.

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -48,7 +48,7 @@ a
 b; c | d
 ```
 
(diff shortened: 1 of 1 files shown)


fdncred:
Good idea. Thanks!

PR #219, opened 2022-02-08
fix: delete duplicate header text

jaeheonji:
Hi! There was duplicate header text while I was viewing Nushell Command Reference, so I open this PR.

Before:
화면 캡처 2022-02-08 113315

After:
화면 캡처 2022-02-08 113335


jntrnr:
thanks!

PR #222, opened 2022-02-16
Fixed typo

Pound-Hash:
`$it` should be `$jt`


fdncred:
I believe `$it` is correct.

PR #236, opened 2022-03-01
Translate table_of_contents to zh_CN from commit #89b037c2d

hustcer:
Translate `table_of_contents` to zh_CN
I will try to translate all English documents into Chinese when the English documents are stable enough.
I will record the commit id in my commit message in order to keep the Chinese document in sync with the English docs, and also to facilitate the discovery of which English documents have been changed after my translation.


fdncred:
I'm wondering, before we land this, if we should copy all the old translations into the "old_book" folder or something so we still have the old translations as a reference if we need them.

PR #253, opened 2022-03-05
Change `without_color` to `use_ansi_coloring`

lucatrv:
Setting `without_color: false` has been changed into `use_ansi_coloring: true`.


fdncred:
Thanks!

PR #254, opened 2022-03-05
Update command input/output docs

rgwood:
## Context

Earlier today I was trying to replace part of a pipeline with a helper function. It took me a little while to figure it out, and somehow I bounced off [the documentation](https://www.nushell.sh/book/custom_commands.html#output) without fully understanding it.

## Changes

- Make it extra clear that the docs are referring to pipeline input/output. This should be obvious on a close read but I think I misunderstood it on a quick skim.
- Rewrite the pipeline input example for `engine-q`; it no longer works on `main` because `each` expects a parameter.

## Considerations

I initially tried to keep the exact functionality of the `my-echo` example but Nushell has changed enough to make that difficult; if you just add an `|it|` parameter it parses but the function incorrectly returns a list of lists:

image

So instead I just copied the first example from `help each`.

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -224,7 +224,7 @@ Flags:
   -a, --age <integer>: The age of the person
 ```
 
-## Output
+## Pipeline Output
 
 Custom commands stream their output just like built-in commands. For example, let's say we wanted to refactor this pipeline:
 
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks!

PR #255, opened 2022-03-05
Add case sensitivity section to operators page

rgwood:
Adding some info to cover the basics of case sensitivity. I think case-insensitive comparisons are a common thing to do and this is worth documenting on the operators page, just let me know if I should put this info elsewhere.


fdncred:
Thanks!!

PR #258, opened 2022-03-07
Update Windows build-from-source requirements

### LINKED ISSUE #63: Asking Windows users to download Visual Studio Community Edition is inappropriate
From [https://www.nushell.sh/book/en/installation.html](https://www.nushell.sh/book/en/installation.html):

> For Rust to work properly, you’ll need to have a compatible compiler suite installed on your system. These are the recommended compiler suites:
> . . .
> . . .
> . Windows: Visual Studio Community Edition
> For Windows, when you install Visual Studio Community Edition, make sure to install the “C++ build tools” as what we need is link.exe which is provided as part of that optional install. With that, we’re ready to move to the next step.

Keep in mind that Visual Studio Community Edition with just the C++ build tools (unless I'm doing it incorrectly) is a **10.4 GB** installation! I'll stick with Vim, thanks.

Since I used scoop.sh to install nu, I already have gcc installed. I'll see if I can just stick with that, but I would recommend that your directions remove such a big ask. It runs very counter to the entire philosophy of nu advertised on the website to ask newcomers to get such a massive dependency.

rgwood:
Closes #63 .

As mentioned in [the `rustup` docs](https://rust-lang.github.io/rustup/installation/windows.html), the VS Build Tools are an OK substitute for anyone who doesn't want all of Visual Studio.

The optional install needed for Rust etc. is called "Desktop development with C++" these days:

image

### DIFF
--- a/book/installation.md
+++ b/book/installation.md
@@ -40,9 +40,9 @@ For Rust to work properly, you'll need to have a compatible compiler suite insta
 
 * Linux: GCC or Clang
 * macOS: Clang (install Xcode)
-* Windows: [Visual Studio Community Edition](https://visualstudio.microsoft.com/vs/community/)
-
-For Windows, when you install Visual Studio Community Edition, make sure to install the "C++ build tools" as what we need is `link.exe` which is provided as part of that optional install.  With that, we're ready to move to the next step.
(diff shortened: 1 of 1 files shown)


jntrnr:
thanks!

PR #264, opened 2022-03-07
Update intro for latest Nu version

rgwood:
The book intro shows a lot of examples, and they had not been updated since `engine-q`. I reran the examples on v0.59.1 and updated the intro accordingly.

### Notes

- I intentionally kept things largely the same as before - I ran the obviously-macOS examples on macOS, and the obviously-Linux examples on Linux. And I tweaked the results so the username + machine name are the same.
- It's a bit weird that virtual memory usage is ~32GB for every process on macOS. Suspect it's a bug related to [this](https://github.com/GuillaumeGomez/sysinfo/issues/428) and [this](https://github.com/nushell/nushell/issues/3003).  It's arguably not ideal to show a bug in the intro docs; if people care about this I could run the examples on another OS?
- The `date now` example isn't quite accurate because of [this bug](https://github.com/nushell/nushell/issues/4698). I left a note there to fix up the docs when the bug is fixed.
- The syntax coloring's a bit off because the entire snippet is interpreted as a bash script. That should probably be changed eventually but it's no worse than before:

image

image


fdncred:
Looks good! Thanks!

PR #265, opened 2022-03-08
Add remaining missing block parameters

rgwood:
Hopefully this is the last of the `each` examples that need to be updated with a block parameter.


jntrnr:
looks good!

PR #270, opened 2022-03-09
elaborate on escaping details

ammkrn:
From an exchange on nu's discord; I wasn't sure what "escaping" meant and was informed that it works by using `PATH`. Jt used this internal/external language to differentiate the two.


fdncred:
we probably need to go a little further at some point and disambiguate between string escaping vs external command escaping. i think we call them both `escaping` which may be confusing to some people.

fdncred:
Thanks!

PR #275, opened 2022-03-10
Update lists page for latest Nu

rgwood:
Fixing some examples that are currently broken on `main` (due to breaking changes in `reduce -n`, the deprecation of `insert`, and mandatory block parameters).


fdncred:
Thanks!

Issue #85, opened 2021-01-23
RSS/Atom feed?


lambdagolem:
I just wanted to ask if there is support/if it would be possible to add support for RSS or atom on the blog? The old blog had support for that I think. I understand that this is probably rather niche and not a priority for the project if it is possible.

Mrfiregem:
As a workaround, I used the Github releases feed since they link to the blog, but having the full article would be much better.

rgwood:
Yeah, this would be great. The website currently uses [VuePress](https://vuepress.vuejs.org/) and it looks like there are a few RSS plugins out there for VuePress. PRs would be very welcome.

rgwood:
Hi, this is done! We've got RSS/Atom/JSON feeds for the blog now:

https://www.nushell.sh/rss.xml
https://www.nushell.sh/feed.atom
https://www.nushell.sh/feed.json

PR #276, opened 2022-03-11
Add RSS feed for blog

rgwood:
As requested in #85, adding RSS/Atom/JSON feeds for the Nu blog.

I looked around and it seems like [vuepress-plugin-feed](https://github.com/webmasterish/vuepress-plugin-feed) is the most popular RSS plugin for VuePress. The documentation wasn't great but it was fairly easy to install.

After adding this, the website build output includes 3 new files: `feed.atom`, `feed.json`, and `rss.xml`. Example: [feeds.zip](https://github.com/nushell/nushell.github.io/files/8229031/feeds.zip). They can be accessed at https://www.nushell.sh/feed.atom etc.

The feeds are _not_ yet linked to from the website; will add links later once I'm 100% confident the feeds are working as expected in production.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -390,6 +390,16 @@ module.exports = {
   plugins: [
     "@vuepress/plugin-back-to-top",
     "@vuepress/plugin-medium-zoom",
+    ["feed", {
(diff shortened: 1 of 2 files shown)


jntrnr:
nice

PR #280, opened 2022-03-12
Nu Book: Reorganize installation page

rgwood:
Making some changes to the installation page, which I previously found a bit confusing.

- Split the install instructions into 3 top-level sections: "Pre-built binaries", "Package managers", "Build from source"
- Add the repology.org badge showing all package managers
- Clarify that crates.io is building from source (IMO this is not obvious to Rust newcomers)
- Move the Windows Terminal instructions to their own section down below (avoids cluttering the main install instructions)

### DIFF
--- a/book/installation.md
+++ b/book/installation.md
@@ -1,38 +1,27 @@
 # Installing Nu
 
-The best way currently to get Nu up and running is to install from [crates.io](https://crates.io), download pre-built binaries from our [release page](https://github.com/nushell/nushell/releases), or build from source.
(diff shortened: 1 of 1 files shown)


rgwood:
whoops, looking at the failing build now...

rgwood:
Build should be fixed now. Need to leave the snippet files around because they're still used by other languages.

fdncred:
at some point we should also document the use of a windows terminal config fragment, but I think this looks good for now. thanks!

PR #282, opened 2022-03-13
Remove link to demo until it's updated

rgwood:
[The WASM demo](http://nushell.sh/demo) is a little out of date (v0.38), and it's looking like it won't get updated before the v0.60 release. Let's remove the prominent link to the demo for now, it'll help avoid some user confusion.


fdncred:
thanks

PR #283, opened 2022-03-13
Update env var docs for latest Nu

rgwood:
I noticed that the Environment page was a little out of date (it referred to an `env` section in config that no longer exists). Fixed that up and also did some other cleanup while I was in the area (added links, added an example to the Configuration page, etc.).


jntrnr:
Thanks!

Issue #278, opened 2022-03-11
break up command_reference


fdncred:
Our new search engine, Algolia, doesn't like massive pages. I don't really blame it. It would be nice to break up the command reference page into alphabetical pages so that these results could be found by searching.

Like:
a.md/html -> alias, all? ansi, any? append
b.md/html -> benchmark, build-string
etc

![image](https://user-images.githubusercontent.com/343840/157901272-e6df5097-3148-4c0a-98f2-580ef61f7802.png)

apparently this requires some type of scripting because this is what the command_reference.md looks like
```
# Command Reference

<div v-for="command in $site.pages.filter(p => p.path.indexOf('/book/commands/') >= 0)">
   <h1><code>{{ command.title }}</code></h1>
   <Content :page-key="command.key"/>
</div>
```

rgwood:
@fdncred As an alternate solution, what if we just change the command reference page to have less content?

Right now it seems to merge _everything_ including examples from every individual command page into 1 monster page. I think the reference page would be more readable if it skipped examples and had a table with links to each individual command (and maybe a short description of each). That would solve the Algolia problems too.

fdncred:
I think I'd like to keep examples on the command page, since I think examples are more important than text most of the time. As far as having a reference page with links to commands, that's kind of what I'm suggesting above. My suggestion was just to order that reference page alphabetically with links to the `a` commands and `b` commands etc. I'm open to other organization strategies though. I just want to make the page smaller so it can be ingested by the crawler.

rgwood:
Hmm. I should have been more clear. Right now we have:

1. Individual command pages (ex: [`ls`](https://www.nushell.sh/book/commands/ls.html), [`where`](https://www.nushell.sh/book/commands/where.html))
2. [The huge command reference page](https://www.nushell.sh/book/command_reference.html) which just concatenates the content of every individual command page

I'm thinking we change the command reference page to be much more succinct, with links to the individual command pages. The individual command pages would still contain examples.

> My suggestion was just to order that reference page alphabetically with links to the a commands and b commands etc.

I might have misunderstood the initial proposal - it sounded like you also wanted to split the existing reference page into multiple pages. Which is doable but I think it would be simpler to keep it as 1 page.

fdncred:
I'm fine with combining our suggestions somehow but I'd rather not have 300+ individual pages for the commands, which is why i was trying to group them by first letter of the command.

rgwood:
[...]

PR #285, opened 2022-03-14
Add command links to book

rgwood:
When I was reading through the book, I felt like it would benefit from more links to individual command pages. I've added a ton of command links via careful find+replace; this is not exhaustive but I think it covers most commands on the most important book pages.

Note: this is a _lot_ of commits. Please squash merge; I'm not trying to farm commit statistics, this was just the easiest way to keep track of individual changes (and maybe easier for review too)


fdncred:
nice add! the only thing I'd say about is that we really need to break up the command reference into smaller groups, like referenced here https://github.com/nushell/nushell.github.io/issues/278, so that our search crawler can actually find these commands. If that happens, all these links may need to be updated. but let's land this. thanks!

PR #289, opened 2022-03-14
move make_docs.nu from nushell to the book

stormasm:
The script is now working on the book side of the world....

Meaning in a couple of days if it appears to be stable we can delete

https://github.com/nushell/nushell/tree/main/docs/commands


fdncred:
nice - was just talking about these commands.

PR #291, opened 2022-03-14
Make command reference more compact

### LINKED ISSUE #278: break up command_reference

rgwood:
Fixes #278. [The command reference page](https://www.nushell.sh/book/command_reference.html) is a beast that brings Chrome and Algolia to their knees, because it's generated by concatenating every individual command page (including examples). This PR changes the command reference to a much more compact table with links:

image

It does so by sticking command usage info in [front matter](https://v0.vuepress.vuejs.org/guide/markdown.html#front-matter) where it can be used in a VuePress template.


rgwood:
@fdncred I know you're not the biggest fan of keeping 1 Markdown file per command, but I think this is a fairly straightforward way to address our Algolia problems. Happy to reconsider or scrap this solution if you disagree.

fdncred:
let's run with it and see how it goes/indexes/searches. Thanks for all the hard work!

PR #292, opened 2022-03-15
Add links to RSS+Atom feeds

### LINKED ISSUE #85: RSS/Atom feed?

rgwood:
Closes #85. The RSS+Atom feeds added in https://github.com/nushell/nushell.github.io/pull/276 seem good in my RSS reader, adding links at the top of the blog page:

image

https://www.nushell.sh/rss.xml
https://www.nushell.sh/feed.atom

### DIFF
--- a/blog/README.md
+++ b/blog/README.md
@@ -1,3 +1,5 @@
 # Nu Blog
 
+[RSS](/rss.xml) / [Atom](/feed.atom)
+
 <BlogPosts />


jntrnr:
nice

PR #300, opened 2022-03-18
Book: mention `help commands`

rgwood:
It occurred to me that we don't really mention `help commands` in a place where a new user reading the docs would stumble across it. Fixing that.


fdncred:
another thing I think we should advertise is `help -f <topic>`. maybe we should have a `Discoverability` section (or something like that) that shows people how to find things out themselves.

PR #301, opened 2022-03-18
temporarily remove the contributor book until we can get it updated

fdncred:
I just commented out all the contributor book references in the config.js. I did not remove references to it in the md files.


rgwood:
Thanks!

I think this will leave the website without much info on plugins. If the Contributor Book’s plugins section is mostly up to date, I could copy the info into the user Book.

fdncred:
It's all out of date except the Introduction and Philosophy pages.

Issue #295, opened 2022-03-16
Organize book chapters better


kubouch:
Right now, the book chapters are ordered quite randomly and there is too many of them in one column on the left. It would be great if we could group them into fewer groups so it's easier to navigate.

rgwood:
Good call, I would love to have fewer pages. Off the top of my head:

- Could probably group the "Nu map from..." pages together. Maybe "Coming from Bash" too
- "Coloring and Theming in Nushell" could arguably go in Configuration
- Pages that arguably don't need their own page "Metadata", "Creating your own errors"
- Could maybe combine Operators and Math

fdncred:
@rgwood Would you be willing to take a stab at this based on whatever you think is a first step at it?

One thing I keep thinking is that when someone goes to the page, they don't necessarily notice the search bar, so they look down the list on the left, which isn't organized alphabetically, so it's kind of hard to find what you're looking for. I'm not sure we can make it alphabetically (maybe we can) but some type of organization, like you're suggesting, would be a great help to the reader I think.

rgwood:
Sure thing, I can try to clean this up a bit.

I believe the pages are ordered manually in config.js, so we can do whatever makes the most sense.

Probably don't want to do full alphabetical (good to keep Introduction+Installing at the top, maybe others too) but I bet we can improve on the ordering somehow.

fdncred:
> 

ya, agreed.

kubouch:
@rgwood I mean instead of merging pages into each other, I was thinking about adding one more layer of indirection: Grouping the existing pages by topic, like the Rust book for example: https://doc.rust-lang.org/stable/book/

I wouldn't go for alphabetical ordering but rather simple first, advanced later. Maybe we could even introduce chapter / section numberings, just like the Rust book.

I'm just drawing inspiration from an existing book that does this pretty well, in my opinion.

rgwood:
@kubouch Gotcha. I'm not sure when/if I'll have time to try that but [the official VuePress docs](https://vuepress.vuejs.org/guide/#how-it-works) have an "Advanced" group so it should be possible:

<img width="192" alt="image" src="https://user-images.githubusercontent.com/26268125/158901541-026932d9-2d2b-467a-aa9e-26d1c7b8d9a1.png">

kubouch:
Yeah, something like the screenshot would look great. For now, we can start reordering the pages and merging when it makes sense.

rgwood:
[...]

PR #302, opened 2022-03-18
Organize book pages by section

rgwood:
An initial pass at organizing the book pages better, as suggested in #295. Also tweaked some page names for consistency.

image


rgwood:
Merging since people seemed positive on this in #website-and-docs

PR #309, opened 2022-03-18
Clean up commands

rgwood:
- `update_commands.py` has been replaced by `make_docs.nu`, deleting it
- ~`inc`, `gstat`, and `query*` commands are no longer present in latest Nu; deleting their pages to remove them from the command reference~
- rerunning `make_docs.nu` after registering the `inc`, `gstat`, and `query*` commands, so their docs get updated correctly


fdncred:
the plugins are in the latest nu. I'm not sure what you're saying about `inc`, `gstat`, and `query`.

rgwood:
@fdncred My bad - I didn't realize those are in plugins. I'm building Nu with `cargo build --features extra` and I don't have those commands, how can I get them?

image

fdncred:
you have to register plugins to use them. `register -e capnp /path/to/nu_plugin_gstat` and the rest of them. you should see them in the `version` command.

fdncred:
I think you're right that the *.py needs to go, but not the rest of the files.

rgwood:
Makes sense. 

The command pages probably need to be updated though. When I have time (tomorrow?) I'll build the plugins, register them, and rerun `make_docs.nu`

fdncred:
Sounds good!

rgwood:
@fdncred Should be good now.

jntrnr:
thanks!

PR #323, opened 2022-03-20
Add externs and custom completions

jntrnr:


rgwood:
This is super helpful to have documented!

I might combine the "Externs" and "Escaping to the system" pages if you don't object. Conceptually I think they can both fit under an "External commands" topic.

PR #328, opened 2022-03-22
Update Plugins page in book

rgwood:
The Plugins page was very bare-bones. Adding some info to explain how plugins work in broad strokes.

This could use a review from someone who's worked with plugins before.


fdncred:
I think that's pretty close to how plugins work. One thing that is missing is how the signatures are cached. That's a big improvement over old plugins. When you call `register` the signature is returned and put in a file called `$nu.plugin-path`. So, now when starting up nushell, nushell doesn't have to interrogate the plugin binaries again. It just opens this file and registers the commands.

@elferherrera Can you read through these preliminary docs? I'd like your take on it before landing this.

rgwood:
Thanks Darren, added some info about `$nu.plugin-path`.

rgwood:
Thanks for the feedback!

PR #329, opened 2022-03-22
Fix minor typos and clarify `Parsing and evaluation...` section

EjPlatzer:
I found two minor typos in the beginning of the book. 

Also, I wanted to try to clarify the `Parsing and evaluation...` section of `Thinking in Nu` because I think this is by far the most confusing section for people coming to Nu. I know there have been discussions of how to clarify it on Discord, but I figured I'd submit some minor changes as a first effort.

---

Personally, I think that the given example is more confusing than is necessary. This is especially true because if you run each line as a separate command in the shell, it will work perfectly - and this is how most people will think to try the script. 

Instead, I would highlight the second example and use that to explain the overall concept. It's much more clear (to me, at least) that sourcing a dynamic path doesn't work because the source command tries to parse the contents of the path before the dynamic expression has been evaluated. We could, for example, give this self-contained example:

```bash
> source $nu.config-path
```

This has the added benefit that it would fail even if run from the shell, unlike the current primary example. Also, I think this is a familiar command to people coming from other shells, because sourcing the [`.bashrc/.zshrc/...`] is the normal way to reload the configuration values. I think that isn't best practice in Nushell, but this section is just about explaining the concepts, not instructing in the proper use of Nu.

### DIFF
--- a/book/README.md
+++ b/book/README.md
@@ -10,7 +10,7 @@ Hello, and welcome to the Nushell project. The goal of this project is to take t
 Nu takes cues from a lot of familiar territory: traditional shells like bash, object based shells like PowerShell, gradually typed languages like TypeScript, functional programming, systems programming, and more. But rather than trying to be a jack of all trades, Nu focuses its energy on doing a few things well:
 
 * Being a flexible cross-platform shell with a modern feel
-* Solving problems as a modern programming language that works the structure of your data
+* Solving problems as a modern programming language that works with the structure of your data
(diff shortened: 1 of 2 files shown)


tikue:
> This is especially true because if you run each line as a separate command in the shell, it will work perfectly - and this is how most people will think to try the script.

Thank you! I was banging my head against this for a few minutes trying to understand what I was missing.

fdncred:
Thanks!

PR #331, opened 2022-03-23
Update plugin register examples and regenerate docs for all commands

hustcer:
Update plugin register examples and regenerate docs for all commands


jntrnr:
thanks :)

PR #335, opened 2022-03-24
Starship: add session key instructions & correct typos

davidkna:
This PR adds instruction for setting the session key that is used for deduplicating log messages and correct some typos. 

By the way, what is `hide PROMPT_COMMAND_RIGHT` supposed to do exactly? It doesn't seem to do anything.

### DIFF
--- a/book/3rdpartyprompts.md
+++ b/book/3rdpartyprompts.md
@@ -49,11 +49,12 @@ let-env PROMPT_INDICATOR = $"(ansi y)$> (ansi reset)"
 
 1. Follow the links above and install starship.
 2. Install nerdfonts depending on your preferences.
-3. Set the starship shell environment variable to `nu` by running this commend `let-env STARSHIP_SHELL = "nu"`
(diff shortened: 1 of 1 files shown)


fdncred:
`hide` hides the `PROMPT_COMMAND_RIGHT` env var as if it doesn't exist. I think this would allow the default `PROMPT_COMMAND_RIGHT` to be used.


```
Hide symbols in the current scope

Symbols are hidden by priority: First aliases, then custom commands, then environment variables.

Usage:
  > hide <pattern>

Flags:
  -h, --help
      Display this help message

Parameters:
  pattern: import pattern

Examples:
  Hide the alias just defined
  > alias lll = ls -l; hide lll

  Hide a custom command
  > def say-hi [] { echo 'Hi!' }; hide say-hi

  Hide an environment variable
  > let-env HZ_ENV_ABC = 1; hide HZ_ENV_ABC; 'HZ_ENV_ABC' in (env).name
```

fdncred:
Thanks

PR #336, opened 2022-03-24
Fix broken table rendering for epub book, close #199

### LINKED ISSUE #199: Generate a downloadable ebook so I can read it on my kindle
Online Nubook is helpful, and it would be better if we can generate a downloadable ebook so I can read it on my kindle even without network. Thanks again.

hustcer:
Fix broken table rendering for epub book, Read https://github.com/nushell/nushell.github.io/issues/199 for more detail


fdncred:
thanks!

PR #337, opened 2022-03-24
Add larger starship example

jntrnr:


book/configuration.md
```diff
@@ -113,4 +128,3 @@ you may want to configure `STARSHIP_SHELL` environment variable when setting the
 config set prompt "STARSHIP_SHELL=nushell starship prompt"
```
fdncred, line 128: config set prompt?

jntrnr: I can delete that too. It's in the current book

PR #338, opened 2022-03-25
Fix starship config to handle negative exit codes

TyPR124:
Currently, if you use starship and a command exits with a negative value, you get an error from starship prompt because it treats, ex, `-1` as a flag. I found passing the status flag as `$'--status=($env.LAST_EXIT_CODE)'` works.

If you want to test this behavior, including seeing how starship parses the status, add this to your starship config:

```
[status]
disabled = false
```

And then you can use something like `nu -c 'exit -10'` to see what happens.


TyPR124:
Closing in favor of fixing this in #337

fdncred:
thanks

PR #341, opened 2022-03-26
Document optional parameters and rest parameters

AaronC81:
I've been trying out Nushell as my daily shell again after the 0.60 release!

This PR adds a couple of short sections to the book explaining functionality which I had to look around a bit to find, specifically:

- Optional positional parameters `def something [x?] { ... }`
- Rest parameters: `def something [...x] { ... }`

This is my first PR to these docs, so more than happy to respond to any review comments about style/consistency/etc!

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -106,6 +106,48 @@ The currently accepted types are (as of version 0.59.0):
  - `string`
  - `variable`
 
+## Optional positional parameters
+
+By default, positional parameters are required. If a positional parameter is not passed, we will encounter an error:
+
+```
+  × Missing required positional argument.
+   ╭─[entry #23:1:1]
+ 1 │ greet
+   ·      ▲
+   ·      ╰── missing name
+   ╰────
+  help: Usage: greet <name>
+```
+
+We can instead mark a positional parameter as optional by putting a question mark (`?`) after its name. For example:
+
+```nushell
+def greet [name?: string] {
+  echo "hello" $name
+}
+
+greet
+```
+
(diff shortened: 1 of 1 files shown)


book/custom_commands.md
```diff
@@ -162,6 +204,38 @@ Now, we can call this updated definition using the shorthand flag:
...
+def greet [vip: string, ...name: string] {
+  echo "hello to our VIP" $vip
+  echo "and hello to everybody else:"
+  for $n in $name {
+    echo $n
+  }
+}
+
+#     $vip          $name
+#     .--. .----------------------.
```
rgwood, line 236: Are these lines supposed to be here?

rgwood:
This is a great first PR, thanks! Left 1 question.

book/custom_commands.md
AaronC81: Ah, yep - those are supposed to highlight which arguments get passed into which parameters. (The first line covers only the first argument and is marked `$vip`, while the second line covers the rest of them which are passed into `$name`.) I can see how it might look a tad strange! Is there a clearer way this could be shown?

rgwood: Ah OK, I get it now - wasn't quite sure what was going on at first glance. I think some of that might be that I didn't realize the `.` and `-` are just parts of the same line; this looks a tiny bit better to me but let me know which way you prefer and I can merge the PR either way.

```suggestion
#     $vip          $name
#     ---- ------------------------
```

AaronC81: I prefer that too, thanks!

PR #343, opened 2022-03-26
docs: table border options

kai687:
This PR improves the list of table border options:

- remove duplicate listing of the `rounded` option
- uses a bulleted list instead of a numbered list (since there isn't an order to the options, it's better to use a bulleted list).


fdncred:
Thanks

PR #344, opened 2022-03-26
docs(german): fix grammar in description

kai687:
Fixed the wrong gender of the adjective "neue" in the description.

"Art" (species, kind) is female, so it must be "eine **neue** Art"


rgwood:
Thanks!

PR #345, opened 2022-03-26
Describe interpolated string escape workarounds

AaronC81:
When first dealing with interpolated strings, I found that I needed to escape some parentheses, and couldn't find any documentation about how to do this. A search in Discord said that there isn't any proper way to do this yet (though please correct me if I'm wrong!)

This PR documents the two ways of working around this that I'm aware of.

### DIFF
--- a/book/working_with_strings.md
+++ b/book/working_with_strings.md
@@ -63,7 +63,7 @@ String interpolation has both a single-quoted, `$' '`, and a double-quoted, `$" 
 
 Interpolated strings do not currently support any way of escaping parentheses. If you would like to include parentheses in the resulting string, one workaround is to use sub-expressions which evaluate to the `(` or `)` characters, such as [`char lp` and `char rp`](commands/char.md). For example:
 
-```nushell
+```
 > $"2 + 2 is (2 + 2) (char lp)you guessed it!(char rp)"
 2 + 2 is 4 (you guessed it!)
 ```
--- a/book/working_with_strings.md
+++ b/book/working_with_strings.md
@@ -59,19 +59,12 @@ greetings, Alice
 
 By wrapping expressions in `()`, we can run them to completion and use the results to help build the string.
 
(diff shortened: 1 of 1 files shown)


fdncred:
I thought we added escaping parens recently. Now I'm going to have to go back and check. LOL.

fdncred:
Found it. This is what characters are escape-able. 
https://github.com/nushell/nushell/blob/3484e0defd76ae35ee569e6c021028401184b601/crates/nu-parser/src/parser.rs#L2175-L2333

AaronC81:
Ah, interesting! I can see that the parentheses are in that list, but how do you use these in an interpolated string?

This doesn't seem to work for me:

```
~$ $"2 + 2 is \(2 + 2)"
Error: nu::parser::parse_mismatch (https://docs.rs/nu-parser/0.60.0/nu-parser/enum.ParseError.html#variant.Expected)

  × Parse mismatch during operation.
   ╭─[entry #17:1:1]
 1 │ $"2 + 2 is \(2 + 2)"
   ·             ▲
   ·             ╰── expected supported escape character
   ╰────

```

fdncred:
ya, you're right. This works `"one\(two\)"` but this doesn't `$"one\(two\)"`. @jntrnr where do we need to update this so that it works in interpolated strings too?

jntrnr:
@fdncred in the parser there's a command that parses interpolated strings (on phone, can't remember exact name). The way it splits needs to respect escapes

jntrnr:
found it

fdncred:
@AaronC81 with change nushell/nushell#4982 escapes should work as intended.

jntrnr:
The fix is in as https://github.com/nushell/nushell/pull/4982. For this PR, we could land some note about 0.60 having an issue, or we could wait and have the fix show up in 0.61

fdncred:
@AaronC81 if you want to clarify your PR we can land it. I think we should just show how it works now and not talk about 0.60 or 0.61 since we're getting closer to the 0.61 release.

AaronC81:
Apologies for the delay! Is this the kind of wording you were thinking of? Happy to tweak if not :)

fdncred:
thanks

fdncred:
thanks

PR #346, opened 2022-03-26
Add `str collect` to "Nu map from ..." tables, and fix entries for `trim`

AaronC81:
This PR:

- Adds `str collect` to the three "Nu map from..." tables. This subcommand is named differently from most other languages I'm familiar with (I'm used to it being called something like `join`), so I missed this completely when looking for it. Hopefully this would make it more discoverable.
- Renames the `trim` rows to `str trim`. The `trim` command was [removed a while ago](https://github.com/nushell/nushell/pull/2560), so this brings the tables up-to-date with this change.

I think most of `str`'s other subcommands have names very similar to other languages, so I'm not sure they all need to be documented here - but if you think otherwise then I'm happy to add some more too :)


rgwood:
Thanks!

PR #347, opened 2022-03-27
Add husky and prettier to format code automatically while committing code

hustcer:
This is a huge change: add husky and prettier to format code automatically while code committing, do `npm i` after pulling the latest code.  This PR could be rejected if we don't want code formatting


rgwood:
Cool! How long does it take for the code formatter to run?

hustcer:
> Cool! How long does it take for the code formatter to run?

Don't worry, only the modified docs will be formatted while committing, that will take less than 1 sec. For a full format please run `npm run pretty` and it will take about 10 seconds. We need the full formatting only once.

jntrnr:
Oh wow! That's pretty big.

I think with the auto-generated command docs -- if we haven't already, we could update the script to output something that's more style-conforming.

hustcer:
> Oh wow! That's pretty big.
> 
> 
> 
> I think with the auto-generated command docs -- if we haven't already, we could update the script to output something that's more style-conforming.

agree，i will update it later

PR #357, opened 2022-03-30
add reference to reduce with example

fellnerse:


rgwood:
Hi, thanks for the PR!

I'm not quite sure how to word this. Your example is converting a list of records (a table) to a single record; I understand [your motivation for doing this](https://github.com/nushell/nushell/discussions/5032) but the description doesn't make that clear. I'm not sure this would make sense to someone reading the docs for the first time. Hmmm.

rgwood:
Gave this a bit more thought.

I think the end of [the Tables section](https://www.nushell.sh/book/types_of_data.html#tables) would be a better place for it (since by then, the reader would have more understanding of tables and records). Maybe with a wording like "To convert 2 columns of a table to a single record:"

rgwood:
Hmm. I think that particular example might be a little confusing, because if a/b/c/d are known ahead of time `{a: b, c: d}` is a simpler way to create that record. Documentation is hard 😅

fdncred:
are we still working on this? if not, let's close it.

PR #358, opened 2022-03-31
book/coming-form-bash: fix type == dir

herlon214:
At version `0.60.0` if you run `ls | where type == Dir` it doesn't return anything. Using the lowercase `dir` works fine.


fdncred:
nice catch. thanks!

PR #360, opened 2022-03-31
Update instructions on aliasing open

filaretov:
The previous instructions did not work, because the first alias would also refer to the system's `open`.

This is not an ideal solution, because we lose the nice help info from Nushell's `open`, but it works.


rgwood:
Nice, thanks for putting this together!

PR #366, opened 2022-04-03
Update environment.md

chtenb:
Add an example of how to extend the path variable, which I think is a very common use case that people would search for.


fdncred:
good add. can you also add an example with `append` so people know that both `prepend` and `append` exist?

chtenb:
@fdncred I've mentioned append as well

fdncred:
thanks

rgwood:
Thanks! Questions about this come up a lot, good to have it documented

PR #367, opened 2022-04-03
Add migration instruction for git bash users

chtenb:


fdncred:
tweaked the wording a bit. thanks!

PR #369, opened 2022-04-03
Use tip blocks (Fixes #365)

### LINKED ISSUE #365: Render quotation blocks with more emphasis
If you look at any page with quotation block, like for example https://www.nushell.sh/book/line_editor.html#configuration, you will see that the blocks are rendered in a grayed out fashion. It happened to me several times that I unconsciously missed some of those blocks even though they contained critical information to me. I know that this is mildly subjective, but if it hurts the documentation readability for me then I'm probably not the only one, as a rule of thumb.

Can we improve the rendering of these blocks? Perhaps they should not be quotation blocks even? Since they are not quotations, but rather important remarks.

chtenb:


chtenb:
@fdncred Is there a way to view the output of this PR somewhere? Or can we only build it locally for inspection?

fdncred:
> @fdncred Is there a way to view the output of this PR somewhere? Or can we only build it locally for inspection?

I think the only way to see it is to build it locally or submit a PR and run with it and see what it looks like. I usually do the latter.

fdncred:
let's just run with this - if we don't like it we can always revert the pr

rgwood:
@chtenb Thanks! For future reference, `npm run dev` is the easiest way to build and autorefresh the website locally: https://github.com/nushell/nushell.github.io/blob/main/package.json#L7

(you will need Node installed, and you'll have to run `npm install` beforehand)

PR #372, opened 2022-04-05
Indicate that `which` works like Bash's `type`

BatmanAoD:
In Bash, `which` only works on executables; to get information about an alias, function, or built-in, you need to use `type`. Nushell's `which` is therefore more like Bash's `type`.

I'm not sure if the length of the new line justifies reformatting the table so that the end-of-line pipes align again.


BatmanAoD:
...is there a simple way to see this rendered somewhere, since GitHub doesn't render the table as a table?

rgwood:
@BatmanAoD The website uses VuePress. `npm run dev` is the easiest way to build and autorefresh the website locally: https://github.com/nushell/nushell.github.io/blob/main/package.json#L7

(you will need Node installed, and you'll have to run `npm install` beforehand)

BatmanAoD:
Is it possible to view what GitHub Actions generates, since it looks like it generates the static files?

rgwood:
I've never tried, I just build the website locally. Either way this seems like a small+safe change, will merge it. Thanks for the contribution!

PR #373, opened 2022-04-06
Translate table of contents and README from commit: c13a71d11

hustcer:
Translate table of contents and README to zh-CN from commit: c13a71d11


fdncred:
thanks

PR #374, opened 2022-04-07
Translate installation to zh-cn from commit: c13a71d11

hustcer:
Translate installation to zh-cn from commit: c13a71d11
And do some docs formatting


fdncred:
thanks Justin

PR #375, opened 2022-04-07
Translate thinking_in_nushell and moving_around to zh-cn 

hustcer:
Translate thinking_in_nushell and moving_around to zh-cn


fdncred:
thanks

PR #380, opened 2022-04-08
Translate working with lists and strings to zh-CN from commit: 13129315c

hustcer:
1. Translate working with lists and working with strings to zh-CN from commit: 13129315c
2. Format book/line_editor.md


fdncred:
thanks

PR #383, opened 2022-04-09
Translate working with tables and pipeline to zh-CN from commit: 1312…

hustcer:
Translate working with tables and pipeline to zh-CN


fdncred:
thanks

PR #384, opened 2022-04-09
Translate custom commands and aliases to zh-CN from commit: 218f600a9

hustcer:
Translate custom commands and aliases to zh-CN from commit: 218f600a9


fdncred:
thanks

Issue #261, opened 2022-03-07
Documents translation status overview


hustcer:
This table holds the overview of the nushell docs' writing and translation status. We need your help to translate the docs. And please notify me after you finish some translation. Thanks

---

| name                            | en          | de | zh-cn | ja | es | pt-BR |
| ------------------------------- | ----------- | -- | ----- | -- | -- | ----- |
| 3rdpartyprompts.md              | In progress | -  | -     | -  | -  | -     |
| aliases.md                      | In progress | -  | -     | -  | -  | -     |
| coloring_and_theming.md         | In progress | -  | -     | -  | -  | -     |
| coming_from_bash.md             | In progress | -  | -     | -  | -  | -     |
| command_reference.md            | In progress | -  | -     | -  | -  | -     |
| configuration.md                | In progress | -  | -     | -  | -  | -     |
| creating_errors.md              | In progress | -  | -     | -  | -  | -     |
| custom_commands.md              | In progress | -  | -     | -  | -  | -     |
| dataframes.md                   | In progress | -  | -     | -  | -  | -     |
| environment.md                  | In progress | -  | -     | -  | -  | -     |
| escaping.md                     | In progress | -  | -     | -  | -  | -     |
| installation.md                 | In progress | -  | -     | -  | -  | -     |
| loading_data.md                 | In progress | -  | -     | -  | -  | -     |
| math.md                         | In progress | -  | -     | -  | -  | -     |
| metadata.md                     | In progress | -  | -     | -  | -  | -     |
| modules.md                      | In progress | -  | -     | -  | -  | -     |
| moving_around.md                | In progress | -  | -     | -  | -  | -     |
| nushell_map.md                  | In progress | -  | -     | -  | -  | -     |
| nushell_map_functional.md       | In progress | -  | -     | -  | -  | -     |
| nushell_map_imperative.md       | In progress | -  | -     | -  | -  | -     |
| nushell_operator_map.md         | In progress | -  | -     | -  | -  | -     |
| operators.md                    | In progress | -  | -     | -  | -  | -     |
| parallelism.md                  | In progress | -  | -     | -  | -  | -     |
| pipeline.md                     | In progress | -  | -     | -  | -  | -     |
| plugins.md                      | In progress | -  | -     | -  | -  | -     |
| regular_expressions.md          | In progress | -  | -     | -  | -  | -     |
| scripts.md                      | In progress | -  | -     | -  | -  | -     |
[...]

PR #385, opened 2022-04-09
update main page [german]

sebastian-xyz:
Match mainpage for german translation with english version and update `i18n-meta.json` according to #261.


fdncred:
thanks

PR #386, opened 2022-04-10
update i18n related script and meta data

hustcer:
Update i18n related script and meta data
For more detail see: https://github.com/nushell/nushell.github.io/issues/261


fdncred:
so cool to see nushell scripts doing real work

PR #390, opened 2022-04-10
nushell#4924 Update data types with binary literal syntax

merkrafter:
Can be published after the nushell version including
Issue https://github.com/nushell/nushell/issues/4924
PR https://github.com/nushell/nushell/pull/5149
was released.


jntrnr:
thanks!

PR #391, opened 2022-04-11
Translate operators and variables_and_subexpressions to zh-CN

hustcer:
Translate operators and variables_and_subexpressions to zh-CN
and ignore formatting of some docs


jntrnr:
thanks!!

PR #393, opened 2022-04-12
Add env.nu to Configuration

kubouch:
Also reworded a bit to reflect that we have two config files now.


rgwood:
Nice update, thanks!

PR #395, opened 2022-04-12
remove old_book

fdncred:
It's probably time to remove the 0.44 book.


jntrnr:
Agreed. Should we delete it?

fdncred:
let's wait to delete. I'm thinking we might get some pushback on this PR in a few days.

PR #397, opened 2022-04-13
Update coming_from_bash.md

fennewald:
Reading through the docs, noticed this, looks out of date. Confirmed functionality with a quick test.

### DIFF
--- a/book/coming_from_bash.md
+++ b/book/coming_from_bash.md
@@ -25,7 +25,7 @@ Note: this table assumes Nu 0.14.1 or later.
 | `mkdir -p <path>`                    | `mkdir <path>`                                   | Creates the given path, creating parents as necessary             |
 | `touch test.txt`                     | `touch test.txt`                                 | Create a file                                                     |
 | `> <path>`                           | `| save --raw <path>`                            | Save string into a file                                           |
-| `>> <path>`                          | `<not yet possible>`                             | Append string to a file                                           |
+| `>> <path>`                          | `| save --raw --append <path>`                   | Append string to a file                                           |
 | `cat <path>`                         | `open --raw <path>`                              | Display the contents of the given file                            |
 |                                      | `open <path>`                                    | Read a file as structured data                                    |
 | `mv <source> <dest>`                 | `mv <source> <dest>`                             | Move file to new location                                         |


rgwood:
Thanks! I don't think the `--raw` flag is necessary; am I wrong about that?

fennewald:
I'm unsure about that myself. I was following the apparent precedent from the example above. In some limited testing so far, the only interesting case is something like:
```
> echo { foo: 1 } | save test.json
> cat test.json
{
    "foo": 1
}
```
While both of these examples fail
```
> echo { foo: 1 } | save test.txt
Error: nu::shell::unsupported_input (https://docs.rs/nu-protocol/0.61.0/nu_protocol/enum.ShellError.html#variant.UnsupportedInput)

  × Unsupported input
   ╭─[entry #53:1:1]
 1 │ echo { foo: 1} | save test.txt
   ·                  ──┬─
   ·                    ╰── Record([("foo", Int)]) not supported
   ╰────
```
```
> echo { foo: 1 } | save --raw test.json
Error: nu::shell::unsupported_input (https://docs.rs/nu-protocol/0.61.0/nu_protocol/enum.ShellError.html#variant.UnsupportedInput)

  × Unsupported input
   ╭─[entry #55:1:1]
 1 │ echo { foo: 1} | save --raw test.json
   ·                  ──┬─
   ·                    ╰── Record([("foo", Int)]) not supported
   ╰────
```

These examples aren't really relevant to `--append` though. My first guess is the documentation suggests `--raw` as it protects against hidden data formatting.

rgwood:
Gotcha. Following the existing precedent works for me, we can clean that up later if we decide that `--raw` is unnecessary.

Thanks for looking into that and improving the docs!

fdncred:
One of the points of `--raw` is when you don't want nushell deciding what to do with the text. Consider these examples.

1. open Cargo.toml
2. open Cargo.toml --raw

Try those and you'll understand better I think.

fennewald:
@fdncred that makes a lot of sense, and aligns with my intial guess about the purpose being related to no hidden data formatting.

`--raw` is definitely necessary imo, consider the following example:
```
> echo '{ foo: 1 }' | save test.json
> cat test.json
"{ foo: 1 }"
```
as opposed to
```
> echo '{ foo: 1 }' | save --raw test.json
> cat test.json
{ foo: 1 }
```

PR #399, opened 2022-04-13
Fix command-less `cd` example

Cyborus04:
The book still shows `cd`-ing without the command as `> dir_name` where it should be `> ./dir_name`


fdncred:
thanks!

PR #400, opened 2022-04-14
Fix some typo and Translate configuration & environment to zh-CN from commit: f5987a82d

hustcer:
Fix some typo and Translate configuration & environment to zh-CN from commit: f5987a82d


fdncred:
thanks

PR #402, opened 2022-04-15
Fix "Concatenate tables" section from "Table docs"

Sygmei:
New method of concatenating table is now using the `append` command.

This PR aims to document this change accordingly


rgwood:
Thanks!

PR #416, opened 2022-04-24
Update loading_data.md

morzel85:
Fetch example changed to `fetch https://blog.rust-lang.org/feed.xml` (https://www.jonathanturner.org/feed.xml is not working).


rgwood:
Thanks for the fix!

Issue #405, opened 2022-04-16
Two pages with the same title


adzubla:
In "Nu Book" there are two pages with the same title, which is a little confusing:

- Nu Fundamentals > Loading Data > Working with strings
- Nu Fundamentals > Working with strings

I sugest changing the page in Loading Data to "handling strings" or "parsing strings".

adzubla:
![Captura de tela 2022-04-16 141440](https://user-images.githubusercontent.com/9607562/163684843-585eda5c-3aa9-4bad-9a6c-005eb2b0455a.png)

fdncred:
Feel free to change the names to differentiate.

PR #424, opened 2022-04-29
Changed the confusing title of "Working With Strings"

merelymyself:
As per issue #405, there are two pages of the title "Working With Strings". I have changed the first to "Handling Strings" in accordance with the suggestions in the issue and changed the relevant links too.


fdncred:
thanks

PR #426, opened 2022-05-03
Fix broken nu_scripts repo link

reillysiemens:
I was reading the [documentation on prompt configuration](https://www.nushell.sh/book/coloring_and_theming.html#prompt-configuration-and-coloring) and noticed this link didn't _quite_ go to the right place. I think this change will fix that.


rgwood:
Thanks for the fix, Reilly!

-Reilly

PR #433, opened 2022-05-04
Explain how to run aliased paths

Dan-Gamin:
Just a simple explanation about using the `run-external` command to run aliased paths. 

_(Is there a better name for this?)_

### DIFF
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -36,6 +36,20 @@ Here is an alias with more than one pipe
 alias lt = (ls | sort-by modified -r | sort-by type)
 ```
 
+## Path Aliases
+
+You may want to add paths to executables as aliases. In order to do this you can use the `run-external` command to run the program.
+
+Here is an alias that will run Firefox (on Windows):
+
+```
+alias firefox = run-external 'C:\Program Files\Mozilla Firefox\firefox.exe'
+```
+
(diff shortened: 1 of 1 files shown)


fdncred:
we shouldn't land this because this is a reserved command that will change in the future. it's not really for use.

Dan-Gamin:
> we shouldn't land this because this is a reserved command that will change in the future. it's not really for use.

Oh.

Dan-Gamin:
I guess this will be put on hold until a better command is finalized.

kubouch:
You can also do `^'some/long/path'` instead of `run-external`

Dan-Gamin:
> You can also do `^'some/long/path'` instead of `run-external`

That's not explained explained in the docs right? (Doing it with commands in the PATH  is explained, doing it with actual path strings is not.

kubouch:
Yeah, it might not be covered well enough. You can do for example this as well: `let cmd = 'some/path/to/command`; ^$cmd` which would also be good to document.

Rather than in the "Aliases" section, I would add it somewhere more appropriate, since it's not alias-specific. For example here: https://www.nushell.sh/book/pipeline.html#working-with-external-commands

fdncred:
thanks for understanding.

PR #435, opened 2022-05-07
Document $in

flying-sheep:
There’s no mention of `$in` in the book so far, let’s fix that.


rgwood:
Thanks for documenting this!

PR #446, opened 2022-05-24
Remove outdated reference to `open` pager

sholderbach:
Previously `open` included syntax highlighting and paging. This has since changed and causes confusion


schuelermine:
gn, you beat me :D

PR #458, opened 2022-05-26
book/operators.md: add `not` operator

schuelermine:


kubouch:
Nice, thanks! Would you mind adding `and` and `or` operators as well? (They are similar to `&&` and `||`.)

schuelermine:
I can’t seem to get them to work, how do they work?

schuelermine:
Oh, OK, I’m not on the latest version, that makes sense

schuelermine:
Does `and` and `or` work just like `&&` and `||` or is there some difference?

fdncred:
> Does `and` and `or` work just like `&&` and `||` or is there some difference?

They're literally the same thing in the code.

https://github.com/nushell/nushell/blob/219c719e986db46b75b8623332e2a23a057b949b/crates/nu-parser/src/parser.rs#L4069-L4070

fdncred:
thanks for the contribution!

PR #465, opened 2022-05-30
Upgrade vuepress to v2 and add dark mode support

hustcer:
### Upgrade vuepress to v2 and add dark mode support

This might be a long journey, so I will create a draft first.

### Check List

- [x] Make sure all node dependencies upgrade to the latest version, especially `vuepress` related modules;
- [x] Make sure dark mode works;
- [x] Make sure all the styles work correctly as before;
- [x] Make sure all the plugins work well;
- [x] Make sure every page / table rendered successfully;

Anything else need to be checked?


rgwood:
Thank you so much for this! Amazing work!

PR #472, opened 2022-06-02
Revert "Upgrade vuepress to v2 and add dark mode support"

hustcer:
Reverts nushell/nushell.github.io#465
The command detail page is blank: https://www.nushell.sh/book/commands/append.html  
We can revert it, Or I can fix it later.


hustcer:
REF: 
1. https://github.com/vuepress/vuepress-next/issues/6
1. https://v2.vuepress.vuejs.org/reference/frontmatter.html#frontmatter

hustcer:
I will fix it, I know how to do that already

PR #477, opened 2022-06-03
Fix search box style, make it wider

hustcer:
Fix search box style, make it wider

### DIFF
--- a/.vuepress/styles/index.scss
+++ b/.vuepress/styles/index.scss
@@ -1,6 +1,14 @@
 .navbar .navbar-items-wrapper {
   flex-direction: row-reverse;
 }
+
 button.toggle-dark-button {
   margin-right: 30px;
+}
+
+#docsearch-container {
+  width: 239px;
+  .DocSearch-Button-Container {
+    width: 150px;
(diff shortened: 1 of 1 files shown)


fdncred:
It's not very wide on safari. Not sure why.
Screen Shot 2022-06-03 at 7 42 38 AM

hustcer:
It seems that it's in mobile mode not PC mode

fdncred:
> It seems that it's in mobile mode not PC mode

that's weird. I'm on my Mac.

hustcer:
I think you didn't take a snapshot for the full nav bar. The search box is wider than before, however, it needn't be tooooo wide, there is no need to to that

fdncred:
ya, it wasn't the full bar - just enough to help you see how short the search box was. here's the full bar. 
Screen Shot 2022-06-03 at 8 14 07 AM

it just looks small to me. it works fine though. if this is the way it's supposed to be then that's fine.

hustcer:
I think It's okay, and will work fine on your mobile phone, we should consider for various kinds of devices

PR #481, opened 2022-06-04
Add more documentation about strings

### LINKED ISSUE #204: Nu Book should have a section about strings
I was learning nu shell using Nu Book and noticed that documentation specific to strings is missing.

Nu Book should have a section about strings: "Working with strings" or "Working with text"
There are some examples of it but they are all scattered in topics about other subjects.

The idea is to show how nu could be used instead of sed/awk to do text processing.

Suggested topics:

- trim
- upper/lower case
- substrings
- spliting
- string compare
- regex matches and groups
- parsing
- type conversion to/from strings
- ansi colors

Dan-Gamin:
Resolves #204.


fdncred:
thanks

PR #489, opened 2022-06-07
Replace `str find-replace` with `str replace` in `coming_from_bash.md`

CAD97:
`str find-replace` is deprecated and no longer functional


fdncred:
Can you just update it to `str replace`? That's what we changed it to.

CAD97:
Yes

PR #494, opened 2022-06-12
Some new German translation with new snippet links

petrisch:
Some files with translation and new snippet links.


petrisch:
@hustcer hopefully I got it right this time :smile:

hustcer:
Thanks

PR #497, opened 2022-06-13
Fixed example causing type_mismatch in working_with_lists.md

### LINKED ISSUE #496: Example in working_with_lists.md causes type_mismatch Error
# Problem
There seems to be a mistake in the "Working with lists" section, at the "Iterating over lists" subsection.
There it states:

> The [reduce](https://github.com/nushell/nushell.github.io/blob/main/book/commands/reduce.md) command computes a single value from a list. It uses a block which takes 2 parameters: the current item (conventionally named it) and an accumulator (conventionally named acc). To specify an initial value for the accumulator, use the --fold (-f) flag. To change it to have index and item values, add the --numbered (-n) flag. For example:

```sh
let scores = [3 8 4]
echo "total =" ($scores | reduce { |it, acc| $acc + $it }) # 15

echo "total =" ($scores | math sum) # easier approach, same result

echo "product =" ($scores | reduce --fold 1 { |it, acc| $acc * $it }) # 96

echo $scores | reduce -n { |it, acc| $acc + $it.index * $it.item } # 3 + 1*8 + 2*4 = 19
```

However when running the last example an error occurs:

```sh
C:\Users\user\Desktop\shelltest〉let scores = [3 8 4]
C:\Users\user\Desktop\shelltest〉echo $scores | reduce -n { |it, acc| $acc + $it.index * $it.item }
Error: nu::shell::type_mismatch (link)
× Type mismatch during operation.
   ╭─[entry #80:1:1]
 1 │ echo $scores | reduce -n { |it, acc| $acc + $it.index * $it.item }
   ·                                      ──┬─ ┬ ──────────┬─────────
   ·                                        │  │           ╰── int
   ·                                        │  ╰── type mismatch for operator
   ·                                        ╰── record<index: int, item: int>
   ╰────
```

# Solution

The error message is a pretty clear indication what went wrong.
So when run with the item value on the accumulator taken into account...

```sh
C:\Users\user\Desktop\shelltest〉let scores = [3 8 4]
C:\Users\user\Desktop\shelltest〉echo $scores | reduce -n { |it, acc| $acc.item + $it.index * $it.item }
19
```
...it works as intended.

f2hafner:
Fixed example causing type_mismatch in "Iterating over lists" subsection.
resolves #496

### DIFF
--- a/book/working_with_lists.md
+++ b/book/working_with_lists.md
@@ -78,7 +78,7 @@ echo "total =" ($scores | math sum) # easier approach, same result
 
 echo "product =" ($scores | reduce --fold 1 { |it, acc| $acc * $it }) # 96
 
-echo $scores | reduce -n { |it, acc| $acc + $it.index * $it.item } # 3 + 1*8 + 2*4 = 19
+echo $scores | reduce -n { |it, acc| $acc.item + $it.index * $it.item } # 3 + 1*8 + 2*4 = 19
 ```
 
 ## Accessing the list


fdncred:
thanks

PR #499, opened 2022-06-14
docs(book): Update the "Coming from Bash" page

aslilac:
- `chmod` is totally a thing
- Line continuations have been supported for a while

### DIFF
--- a/book/coming_from_bash.md
+++ b/book/coming_from_bash.md
@@ -34,7 +34,7 @@ Note: this table assumes Nu 0.14.1 or later.
 | `rm <path>`                          | `rm <path>`                                      | Remove the given file                                             |
 |                                      | `rm -t <path>`                                   | Move the given file to the system trash                           |
 | `rm -rf <path>`                      | `rm -r <path>`                                   | Recursively removes the given path                                |
-| `chmod`                              | `chmod`                                          | Changes the file attributes                                       |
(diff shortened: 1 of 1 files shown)


kubouch:
Thanks! I wouldn't call it just "line continuation", though. I'd phrase it "Line continuation is possible within a subcommand" or something along these lines.

kubouch:
Actually, also `chmod` is not a Nushell command. It just calls the Linux program but won't work on Windows, for example (or is there chmod on Windows?). You could say that it calls the Linux executable but it's not a part of Nushell.

aslilac:
> `chmod` is not a Nushell command

You could say the same for mkdir, cp, mv, rm, etc. 🤷‍♀️

> I wouldn't call it just "line continuation", though

Yeah, now that you point it out, I feel like "line continuation" might even be a bit too technical of a way to describe it. Lmk what you think of my new proposal!

aslilac:
Actually I just realized Nushell actually *is* normalizing all the commands I listed. 😅 Should we just remove `chmod` then?

fdncred:
If you want to leave it in there, I'd prefix it with a caret, which is the formal way of calling externals in nushell. Like `^chmod`.

kubouch:
I would say that you can run `chmod` as any other external command. The current formulation still suggests that `chmod` is a part of Nushell which isn't true.

PR #508, opened 2022-06-17
Remove old_book related stuffs

hustcer:
Remove old_book related stuffs


rgwood:
Nice, thanks for taking care of this.

PR #512, opened 2022-06-18
DE translation for loading data and working with strings

petrisch:
@hustcer Here the new PR.
Thanks again for the review.


hustcer:
Thanks, It's much better now, one more thing: 3a998f96f8 doesn't seem to be a commit from the main branch. You can do the following to get this commit hash:

1. Run `git log` on main branch, find the commit that your translation based on, let's say you get the following output:

```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────
commit fa908953079d2fc8fb15c8ebddf82ad95b90b869
Author: Justin Ma <hustcer@outlook.com>
Date:   Fri Jun 17 09:59:15 2022 +0800

    Remove old_book related stuffs (#508)

──────────────────────────────────────────────────────────────────────────────────────────────────────────────
commit 0a40c03bef96687e7cad6b39e16bb2ed978b113e
Author: Reilly Wood <reilly.wood@icloud.com>
Date:   Thu Jun 16 09:36:53 2022 -0700

    Remove MS CRT info now that we're statically linking it'

──────────────────────────────────────────────────────────────────────────────────────────────────────────────
commit 0c788acfec4199f2cc0870cc85f939c786687434
Author: Justin Ma <hustcer@outlook.com>
Date:   Thu Jun 16 14:42:45 2022 +0800

    Fix command overlapping for make_docs.nu and Refresh commands for v0.64 (#505)

```
And supposing your translation based on commit `0c788acfec4199f2cc0870cc85f939c786687434`, then

2. Run `git rev-parse --short 0c788acfec4199f2cc0870cc85f939c786687434` to get a short commit id, and that's it.

petrisch:
Ah you give the one that its based on, so basically the last commit before my changes will be applied to right?
That makes sense! thx. It will take me a while because I don't have access to it today, so if you  want, you can correct it again.

hustcer:
> so basically the last commit before my changes will be applied to right?

In most cases this is correct.
Ok, I will update the PR and merge it, thanks again

PR #522, opened 2022-06-27
Links `types_of_data` page to `describe` command

merelymyself:
I feel this makes it make more sense / more cohesive? / helps new users?


fdncred:
thanks

PR #523, opened 2022-06-29
Syntax highlighting as `console`

toastal:
The code blocks are shell sessions and not shell scripts which is why the highlighted oddly (`install` a different color)

---

Out of scope project suggestions:

* Support a mirror on an open source Git forge
* Support an open source communication alternatives like IRC or Matrix

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,13 @@ Nushell is available as [downloadable binaries](https://github.com/nushell/nushe
 
 #### macOS / Linux:
 
-```sh
+```console
 $ brew install nushell
 ```
 
 #### Windows:
 
-```powershell
+```console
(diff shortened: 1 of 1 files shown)


fdncred:
Same as the nushell issue, I'm not sure how I feel about tables in blue. I'm not really sure what `console` buys us. I'd rather just have our own custom theme somehow.

fdncred:
I think we'll keep it how it is now. Thanks for your time in this submission and trying to make us look better.

toastal:
This one doesn't have a table. These shell session examples are both *technically* the wrong syntaxes.

fdncred:
ok

PR #524, opened 2022-06-29
Fix outdated `$false` value

relby:
`false` and `true` are builtin values since 0.60.0 version. Therefore a dollar sign is redundant.


fdncred:
thanks

PR #525, opened 2022-07-02
Changes misleading example

merelymyself:
It seemed to suggest that $it was necessary in {|$somevar| $somevar.size > 10kb }.


fdncred:
Right. You can name the variable whatever you want.

PR #526, opened 2022-07-03
DE translation for pipelines and scripts

petrisch:
and changed title of working_with_scripts.
It just looked weird.
At some point i will have to make the structure correct,
but I think finishing the files itself benefits more.
Slowly approaching...


hustcer:
Thanks, BTW, there are some command links start with `commands/` that need to be changed to `/book/commands/` to avoid `404`

petrisch:
oh damn, I'm making the same mistakes over and over. I really need to ramp up local testing.

hustcer:
@petrisch No worries at all. Thanks for your contribution 👍

PR #531, opened 2022-07-08
Add helpful links

merelymyself:
Small convenience change that was just bugging me. All the `ls` were hyperlinked, so the other commands should be too.


fdncred:
thanks

PR #532, opened 2022-07-13
add light theme relative doc

WindSoilder:
Add light theme description to handle https://github.com/nushell/nushell/issues/5961


fdncred:
Nice add! We have questions about this from time to time.

PR #533, opened 2022-07-14
Remove entry for `count`

merelymyself:
it would show up first when people search for `wc` on the page, but it's no longer a command. `size` is the more useful alternative.


kubouch:
`wc` is a Linux utility to count words (and `wc -l`). I would still keep it there since `length` is pretty much its alternative.

merelymyself:
I don't mind - I just feel that size is much more true to the functionality of `wc`, because it shows the "newline, word, and byte counts for each file"

kubouch:
Actually, I think we can put both `size` and `length` as `wc` substitutes. I didn't know we had a `size` command :laughing: . 

I got a bit confused since the PR diff didn't show any `size`.

fdncred:
Agreed, I'd add `size` to `length` here even though `size` is the near identical clone of `wc`.

merelymyself:
`size` already exists as a row in the table - i think line 74.

PR #534, opened 2022-07-14
More generic `describe` usage

flying-sheep:


fdncred:
I'm not sure it's a good idea to change it here since these files are programmatically generated. There are apparently 3 describe commands that will be used depending on context. You can see them with `help describe`. There's one for "normal" usage, one for DB, and one for Dataframes. What we need to be able to do is infer the correct one at the correct time.

flying-sheep:
well, if “help describe” shows all three, the “describe” online docs should too. The two I merged here have a very comparable purpose, the 3rd one does column statistics on a dataframe.

Instead we have 2 somehow.

fdncred:
>  if “help describe” shows all three, the “describe” online docs should too.

agree.

@hustcer what do you think? do we need to update our doc generation script?

hustcer:
@fdncred Sorry for the late reply, I just saw this. It's a good idea to update the doc generation script and vue-press page rendering related stuff as the commands' doc layout will be changed a lot. It's not an easy task, PR is welcomed.

fdncred:
closing this because it needs to be done via the script.

PR #540, opened 2022-07-23
Update working_with_lists.md (removing unnecessary use of echo)

morzel85:
Some of the code examples contained use of `echo` which is not necessary in current Nu. Let's encourage the simplest use possible.


fdncred:
thanks

PR #542, opened 2022-07-26
Add overviews of chapters; Move Introduction to top level

kubouch:
Previously, the book was mostly a collection of pages that were not tied together. This PR aims to connect the content better by the following:
* Bring Introduction into its own chapter, make it a landing page and add overview of each chapter to it.
* Added an overview of the Nushell ecosystem to Introduction to connect the different pieces we have: Nushell, Nana, nu_scripts, etc.
* For each chapter, make a page for it with a quick overview of its contents. These pages can also point out some of the most important points of the chapter.

Hopefully, it will make the book feel more connected and approachable.

getting_started

Also added a few points, like a proper mention of NUON, while I was at it...


fdncred:
Nice touch. It makes the book easier to consume, I think. Definitely like more organization.

rgwood:
Thank you! Agreed that this makes the book more approachable.

jntrnr (approved):

PR #548, opened 2022-08-01
DE Translation for Coloring and Theming 

petrisch:
Plus some update on links and meta data.
I wasn't able to contribute as much as I wanted the latest time...


hustcer:
Looks good, thanks

PR #552, opened 2022-08-04
FIX: update the links to the default config files

amtoine:
In this PR, I have replaced all occurences of some old and broken links to the default config files.

These files are now [default_config.nu](https://github.com/nushell/nushell/blob/main/crates/nu-utils/src/sample_config/default_config.nu) and [default_env.nu](https://github.com/nushell/nushell/blob/main/crates/nu-utils/src/sample_config/default_env.nu) and the book did not link to these new locations on some pages :confused: 

I've run global pattern searches to make sure:
  - I have caught all the bad URLs mentioning one of the two config files in [nushell/nushell.github.io](https://github.com/nushell/nushell.github.io)
  - the two files above are correct and the only ones in [nushell/nushell](https://github.com/nushell/nushell)

Hope you'll find that usefull :relieved:
cheers :tada:

### DIFF
--- a/book/environment.md
+++ b/book/environment.md
@@ -145,7 +145,7 @@ BAR
 ## Environment variable conversions
 
 You can set the `ENV_CONVERSIONS` environment variable to convert other environment variables between a string and a value.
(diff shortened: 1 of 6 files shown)


fdncred:
thanks

amtoine:
your're welcome :yum:

PR #558, opened 2022-08-16
fix: all clickable items should have the same color

azzamsa:
Hi.

In the previous state. It is hard to know that the command is clickable.

image

To tell reader (without thinking) that those commands is clickable, they need to have the same color as other clickable item.

image


fdncred:
thanks

PR #559, opened 2022-08-16
docs: contributing workflow is outdated

azzamsa:
It is not in sync with https://github.com/nushell/nushell.github.io/blob/b2fc46f3d72c11f9fa29aa7dcb55c33ac41768e0/.github/workflows/ci.yml#L15


fdncred:
thanks

PR #561, opened 2022-08-16
Update direnv.md

johnae:
To actually have direnv reevaluate any new changes without stepping out/in to the directory again, it needs to run in pre_prompt instead of env_change.


johnae:
Also see https://github.com/nushell/nu_scripts/pull/276

fdncred:
thanks

PR #567, opened 2022-08-19
Create coming_from_cmd.md

CAD97:
Like coming_from_bash.md, this maps cmd builtins to nu concepts. The main goal is to soften [#6359: CMD.exe internal commands no more recognized and throwing errors/breaking scripts](https://github.com/nushell/nushell/issues/6359), which was caused by no longer falling back to calling cmd.exe (https://github.com/nushell/nushell/pull/6253) when a command is not recognized.

### DIFF
--- a/book/coming_from_cmd.md
+++ b/book/coming_from_cmd.md
@@ -1,0 +1,57 @@
+# Coming from CMD.EXE
+
+This table was last updated for Nu 0.66.0.
+
+| CMD.EXE                              | Nu                                               | Task                                                              |
+| ------------------------------------ | ------------------------------------------------ | ----------------------------------------------------------------- |
+| `ASSOC`                              |                                                  | Displays or modifies file extension associations                  |
+| `BREAK`                              |                                                  | Trigger debugger breakpoint                                       |
+| `CALL <filename.bat>`                | `<filename.bat>`                                 | Run a batch program                                               |
+|                                      | `nu <filename>`                                  | Run a nu script in a fresh context                                |
+|                                      | `source <filename>`                              | Run a nu script in this context                                   |
+|                                      | `use <filename>`                                 | Run a nu script as a module                                       |
+| `CD` or `CHDIR`                      | `$env.PWD`                                       | Get the present working directory                                 |
+| `CD <directory>`                     | `cd <directory>`                                 | Change the current directory                                      |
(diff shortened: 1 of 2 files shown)


hustcer:
Thanks for your contribution, can you update the `.vuepress/config.js` and add a menu for this doc?
```diff
                 '/book/coming_from_bash.md',
+                '/book/coming_from_cmd.md',
```

CAD97:
Did so

hustcer:
Great, let's land it

book/coming_from_cmd.md
```diff
@@ -0,0 +1,57 @@
+# Coming from CMD.EXE
+
+This table was last updated for Nu 0.66.0.
+
+| CMD.EXE                              | Nu                                               | Task                                                              |
+| ------------------------------------ | ------------------------------------------------ | ----------------------------------------------------------------- |
+| `ASSOC`                              |                                                  | Displays or modifies file extension associations                  |
+| `BREAK`                              |                                                  | Trigger debugger breakpoint                                       |
```
fdncred, line 8: Some of these should probably be cleaned up a bit. break doesn't trigger a break point in cmd.exe. https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/break

fdncred:
I this is probably helpful to people in Windows. Thanks for adding it. I'd like to see these definitions tweaked please. Some are a bit off. I'd use Microsoft's definitions https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/windows-commands

CAD97:
I was looking at `/?` when I wrote these, but it's worth noting that these are more generally "task I'm trying to accomplish" than "what exactly does the cmd version do", and is biased towards the nushell side of the migration.

book/coming_from_cmd.md
CAD97: According to `/?`, it does with extensions. Without extensions, it doesn't do anything under Windows.

CAD97: From that page,

No effect:

> Because the break command has no effect

Trigger breakpoint:

> If command extensions are enabled and running on the Windows platform, inserting the break command into a batch file enters a hard-coded breakpoint if being debugged by a debugger.

fdncred: oh, I see. you're right. I stopped reading after this part.

> Sets or clears extended CTRL+C checking on MS-DOS systems. If used without parameters, break displays the existing setting value.

PR #570, opened 2022-08-20
Update coming_from_cmd.md for nu 0.77

CAD97:
(Just downloaded the new version from winget)

The notable fix is changing `cp <destination> <source>` to `cp <source> <destination>`.

I also went ahead and added an interesting use of cmd `COPY` which I've run into.


fdncred:
thanks

PR #572, opened 2022-08-21
Note Windows caveat in Escaping to the System

CAD97:
`^external` no longer runs commands through the CMD interpreter, so does not pick up builtins (except for the explicitly allowed ones). Make this more readily apparent.


rgwood:
Thanks for documenting these!

PR #577, opened 2022-08-26
Remove ambiguity about command quotes

### LINKED ISSUE #566: command name string clarification
The custom commands section states

> In Nushell, a command name is a string of characters or a quoted string.

However this description may be misinterpreted as saying that commands can be called with a (possibly quoted) string as the command name, e.g. `ls | "my subcommand"`. Maybe it would be good to clarify here, or later in the custom commands section, that while commands can be defined with quoted-string names, they cannot be called that way.

merelymyself:
Also removes reference to `123` being a valid command name; although defining it throws no error, it's impossible to call it.

Given the next subsection shows how to define a custom command with spaces in it, this should be fine, just removes ambiguity.

Fixes #566.

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -65,6 +65,14 @@ Now we can call our custom command as if it were a built-in subcommand of `str`:
 > str mycommand
 ```
 
+Of course, commands with spaces in them are defined in the same way:
+
+```nushell
+def "custom command" [] {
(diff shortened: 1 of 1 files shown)


hustcer:
ci should work now, please pull and rebase the main branch to update your PR, thanks

fdncred:
thanks

PR #579, opened 2022-08-26
Add wrap-around merge

Yethal:


kubouch:
Thanks for writing this up. I'm wondering if it should go into the Cookbook instead. The Book is mostly about teaching the fundamentals but the wrapping seems like a more advanced and specific task that it seems better suited for the Cookbook, in my opinion.

I'd also prefer having the examples written with spaces around pipes, like `| flatten` instead of `|flatten` as the latter makes it harder to read (again, my opinion).

hustcer:
ci should work now, please pull and rebase the main branch to update your PR, thanks

Yethal:
@kubouch I added spacing around pipes for better readability. Regarding the cookbook, do you want a new article in the cookbook for tables specifically?

kubouch:
Yeah, you can add a new article about more advanced table workflows and put it there. The cookbook is not very stuctured as it does not have much content so it does not matter as much where exactly it is put.

kubouch:
I think you might need to add the new .md file to the config, otherwise it wouldn't be visible on the website: https://github.com/nushell/nushell.github.io/blob/main/.vuepress/config.js#L196.

Yethal:
Yeah, I figured that might be necessary

kubouch:
Looks good, thanks!

PR #581, opened 2022-08-27
Try to fix broken CI

kubouch:
Let's try to fix the dynamic import of "path" causing errors.


hustcer:
I will try to fix it today

PR #589, opened 2022-09-05
Update types_of_data.md with missing values

chrjen:
'Duration' and 'File sizes' sections were incomplete with values missing.

For the duration `1month` I wrote "30 days" for clarification. A month can be anywhere from 28–31 days depending on the month while the program always seems to treat a month as 30 days.

### DIFF
--- a/book/types_of_data.md
+++ b/book/types_of_data.md
@@ -139,16 +139,19 @@ Eg) `1wk` is the duration of one week.
 
 This chart shows all durations currently supported:
 
-| Duration | Length          |
-| -------- | --------------- |
-| `1ns`    | one nanosecond  |
-| `1us`    | one microsecond |
-| `1ms`    | one millisecond |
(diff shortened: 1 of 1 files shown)


fdncred:
good catch. I'd also add that leap years are not calculated at all. no leap days, leap hours, leap seconds, no leap anything. So, a year, I believe is calculated at 365 days always.

chrjen:
> good catch. I'd also add that leap years are not calculated at all. no leap days, leap hours, leap seconds, no leap anything. So, a year, I believe is calculated at 365 days always. feel free to update year as 365 days if you want.

That is a little more tricky since year and decade are both abbreviated. I have updated the PR with some changes. I decided to simply write the name and instead put the actual length in days in parenthesis. Hopefully this is good.

fdncred:
yup, lgtm. thanks

PR #591, opened 2022-09-05
Update i18n script to `is-empty` name change

sholderbach:
Account for nushell/nushell#6464


sholderbach:
@hustcer I separated that change from the pure docs updates. I wasn't sure when or with which version you or CI is running that.

hustcer:
Great, thanks

Issue #593, opened 2022-09-05
Improve documentation for def/custom command definition


generic-github-user:
Some of the examples from the [page on custom commands](https://www.nushell.sh/book/custom_commands.html) may be worth adapting to help provide more comprehensive usage information in the documentation on the [`def`](https://www.nushell.sh/book/commands/def.html) statement; particularly those covering subcommand definitions, parameter types, default parameter values, etc. The command description itself and parameter information could probably also be made more thorough. I'll put together a PR to this effect and you can decide whether it would be beneficial to integrate into the documentation.

NotTheDr01ds:
Looks like this can be closed - It was for the `def` command and according to the PR was moved over to the main repo.  Regardless, I think the `def` doc is much improved in this area.

PR #594, opened 2022-09-05
Improve command documentation for def statement

generic-github-user:
These changes correspond to the ones I proposed in #593.


fdncred:
Changes to command/blah.md will get overwritten because they're auto generated. I think in order to make changes to command documentation is to change the rust code in the nushell side.

generic-github-user:
Oh, thanks for clarifying. I thought that since running `npm run dev` successfully built and served the documentation site this would be sufficient; I'll copy the updates over there and open a new PR (unless there is a better way). This might be worth mentioning in `CONTRIBUTING.md` as well.

PR #600, opened 2022-09-08
update plugin for newest changes

WindSoilder:
Reflect to changes in https://github.com/nushell/nushell/pull/6486


hustcer:
Thanks, and I think we should land it after the next release

hustcer:
0.68.1 released, let's land it

PR #603, opened 2022-09-10
Update modules.md

Yethal:
Remove information that importing a module from within a module is impossible as it's outdated.


kubouch:
Thanks!

PR #604, opened 2022-09-10
Cookbook : SSH passphase management

apraga:
Following the suggestion of #2812.


fdncred:
It might be worth saying that you can also achieve the same output with `transpose` instead of `reduce` like this.
```
ssh-agent -c | lines | first 2 | parse "setenv {name} {value};" | transpose -i -r -d
```

fdncred:
thanks

PR #606, opened 2022-09-12
Add versioning support for books

hustcer:
Add versioning support for books
The sad news: build time increased dramatically from 8min to about 22min


hustcer:
Close it for now as it takes too long to build and deploy

PR #610, opened 2022-09-24
Fix some typos

tensorush:
Decided to correct some typos while reading the docs.


dandavison:
This looks great @tensorush. I was also planning to do some proof-reading edits of the docs. The only thing I'd question here is that I do think the non-hyphenated word "builtin" is standard in this context.

https://www.gnu.org/software/bash/manual/html_node/Shell-Builtin-Commands.html
https://en.wikipedia.org/wiki/Shell_builtin

tensorush:
Hi, @dandavison! I agree that "builtin" is also legitimate, but I was making this change based on the observation that there are much more "built-in" than "builtin" occurrences throughout the docs.

fdncred:
I'm good with either but I like making things consistent and love the grammar and spelling corrections. Thanks so much!

dandavison:
Yes the main thing is fixing the typos; this builtin question is second-order.

IMO there are two distinct cases:

**Non-hyphenated:**

> which meant that the above builtins could be run

**Hyphenated:**

> will default to a built-in `LS_COLORS` setting

tensorush:
@dandavison I see. So, the preference may be to hyphenate when its an adjective and not hyphenate when its a noun. Still a matter of preference I guess.

fdncred:
Ya, I see what you're saying. I'd be ok with another PR for this.

PR #612, opened 2022-09-25
Update Modules and Overlays chapters

kubouch:


book/modules.md
```diff
@@ -147,73 +145,45 @@ hi there!
...
-export def hello [name: string] {
-    $"hello ($name)"
+export def hello [] {
+    $"hello ($env.MYNAME)"
 }
 ```
 
-`use` works the same way as with custom commands:
-
+`use` will run the code inside the `export-env` block and merge its environment into the current scope:
```
dandavison, line 162: Users will wonder why they can't simply use `let-env` at the top-level in their module, since the feature seems to behave exactly as one would expect with a top-level `let-env` (i.e. the variables are available in `def` bodies inside the module, and the variables get merged into the current scope when doing `use`). Do you think it's worth adding a sentence explaining why it's not done using a top-level `let-env` in a module?

kubouch: Well, it's not just about `let-env` but also `load-env` and any other calls that might bring in environments (e.g., defined with `def-env`) and in fact basically any imaginable code, including variables, that you might need for setting up the environment. Inside `export-env` it's nicely scoped and won't pollute the namespace of other commands inside the module. I'll add a sentence about this.

book/modules.md
```diff
@@ -147,73 +145,63 @@ hi there!
...
 
-> $env.ROLL
-6
+    let len = (tmp | str length)
 
-> $env.ROLL
-6
+    load-env {
+        TMP_LEN: $len
+    }
```
dandavison, line 185: This helps! However, as a user reading this I'm now wondering why I have to use `load-env`? Why don't I just write

```
let-env TMP_LEN = $len
```

here, seeing as above we wrote

```
export-env {
    let-env MYNAME = "Arthur, King of the Britons"
}
```

?

kubouch: It's just an example that illustrates you can bring in the environment in any way, not just `let-env`.

dandavison: OK I see. I think this section would be even clearer if it was something like:
```
export-env {
    def name [] { "Arthur" }
    def title [] { "King of the Britons" }

    let-env MYNAME = $'(name), (title)'
}
```

Otherwise it risks seeming like `load-env` must be used when the `export-env` block contains computed values. But it's a fairly minor thing, just a suggestion.

sholderbach:
@kubouch anything stopping us from landing this PR?

kubouch:
Oh, I forgot to merge it!

PR #613, opened 2022-09-26
Explain … for a repeating positional parameter

drbrain:


fdncred:
What are "repeating" positional parameters? I think you're talking about `...arg`. If so, I believe we call those "rest" parameters because it collects all the rest of the arguments aka all the remaining arguments.

drbrain:
I'm happy to change this PR to label what I call "repeating" parameters "rest parameters" or whichever other term is correct, but I'm unsure if I've been directed to make that change.

It seems the `...arg` doesn't need to be in tail position for completions to work correctly:

```
  export extern "git mv" [
    ...source: glob
    destination: path
    --force(-f)   # Force renaming or moving of a file even if the target exists
    -k            # Skip move or rename actions which would lead to an error condition
    --dry-run(-n) # Do nothing; only show what would happen
    --verbose(-v) # Report the names of files as they are moved
  ]
```

When I use this I get no error:

```
❯ git mv Cargo.lock Cargo.toml scripts
❯ git status
On branch REDACTED
Your branch is up to date with 'origin/REDACTED'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    Cargo.lock -> scripts/Cargo.lock
	renamed:    Cargo.toml -> scripts/Cargo.toml
```

Or after undoing all that and using globs:

```
❯ git mv *.lock *.toml scripts
❯ git status
On branch REDACTED
Your branch is up to date with 'origin/REDACTED'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    Cargo.lock -> scripts/Cargo.lock
	renamed:    Cargo.toml -> scripts/Cargo.toml
	renamed:    rustfmt.toml -> scripts/rustfmt.toml
```

fdncred:
I think either "the rest of the parameters" or "the remaining parameters" is probably good. thanks!

drbrain:
Updated in 95f85a0c18181c3a7c13b99b22f95639e1c687f3

fdncred:
Thanks.

PR #619, opened 2022-10-01
Use `str collect` instead of `str join`

yang-ling:
`str join` does not exist.


hustcer:
Which version of nushell are you using? `str join` and `is-empty` are available in the latest nu

PR #625, opened 2022-10-04
Merge docs with the same command name

hustcer:
Merge command docs with the same command name, this is a fix to https://github.com/nushell/nushell.github.io/pull/534


hustcer:
Merge docs with the same command name

hustcer:
It works fine for me, Please check: http://localhost:8081/book/command_reference.html and http://localhost:8081/book/commands/describe.html when you run it locally

rgwood:
Thanks!

PR #631, opened 2022-10-07
update the modules chapter with let-env command

maxim-uvarov:
Current help is obsolete. So I rewrote parts with `export env`. 

I found that the last part isn't actual. So I deleted it. 

And finally, when the name is the module name (assuming the previous `greetings` module):

```
> use greetings.nu

> $env.MYNAME
Arthur, King of the Britons

> greetings hello "world"
hello world!

> hide greetings

> $env."greetings MYNAME"  # error! environment variable not found!

> greetings hello "world" # error! command not found!
```

### DIFF
--- a/book/modules.md
+++ b/book/modules.md
@@ -153,7 +153,7 @@ The syntax is slightly different than what you might be used to from commands li
 ```
 # greetings.nu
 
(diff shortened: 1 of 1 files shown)


maxim-uvarov:
oh, it was already rewritten in #612 
ok:)

PR #632, opened 2022-10-12
Fix a sample code on the "moving around" chapter

ohno418:
Remove hidden files from the result of `ls` command. It doesn't seem to output hidden files.

### DIFF
--- a/snippets/moving_around/ls_deep_glob_example.sh
+++ b/snippets/moving_around/ls_deep_glob_example.sh
@@ -1,15 +1,13 @@
  ls **/*.md
-────┬───────────────────────────────────────────┬──────┬─────────┬────────────
- #  │ name                                      │ type │ size    │ modified 
-────┼───────────────────────────────────────────┼──────┼─────────┼────────────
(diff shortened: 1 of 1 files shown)


sholderbach:
Thanks for catching that discrepancy. Interestingly adding `-a` doesn't seem to change that. So this might be a regression we need to address in general.

fdncred:
It looks like it's platform dependent maybe? This is from Windows.
image

fdncred:
On Linux with 0.69.2 I get this:
image

and `ls -a`
image

sholderbach:
On linux `ls` vs `ls -a` behaves as expected. The problem is the inconsitent behavior with a glob e.g. `ls **/*.md` vs `ls -a **/*.md`

sholderbach:
Apart from the bug there let's move forward with your PR.

sholderbach:
OK, gave this a test drive on linux again: `ls -a <GLOB>` seems to behave as expected as well. (was mistakenly running in the docs repo instead of the nushell source repo)

PR #633, opened 2022-10-12
Explain run-external redirection and add example

drbrain:


sholderbach:
Closed by nushell/nushell#6712

PR #634, opened 2022-10-13
add reference to semicolons, remove redundant `group` from types_of_data

merelymyself:
Fixes https://github.com/nushell/nushell/issues/6668.

The `groups` section is deprecated.


fdncred:
thanks

yhm-amber:
Did that means the `;` is just equal to a `new line` now ?

----

*I do think the `group` feature is an interest idea, but it just should not appear by making `new line` diff with `;` .*

PR #635, opened 2022-10-13
Fix example variables using `-` in their name

HoLLy-HaCKeR:
This PR fixes examples using the `-` character in their name, which does not seem to be valid in nu 0.69.1.

I found these occurences with the regexes `let \w+-` and `\$\w+-`, so it's possible I missed something.

### DIFF
--- a/book/commands/input.md
+++ b/book/commands/input.md
@@ -25,5 +25,5 @@ usage: |
 
 Get input from the user, and assign to a variable
 ```shell
-> let user-input = (input)
+> let user_input = (input)
 ```
--- a/book/commands/input.md
+++ b/book/commands/input.md
@@ -25,5 +25,5 @@ usage: |
 
 Get input from the user, and assign to a variable
 ```shell
-> let user_input = (input)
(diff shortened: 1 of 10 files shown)


book/commands/input.md
```diff
@@ -25,5 +25,5 @@ usage: |
 
 Get input from the user, and assign to a variable
 ```shell
-> let user-input = (input)
+> let user_input = (input)
```
fdncred, line 28: I believe the command md files are generated systematically, so changing this here will just get overwritten next time they're generated. So, this should probably be changed in the crates/nu-command/src/platform/input.rs file.

fdncred:
It all looks good except the one comment I made. If you remove that one, and perhaps fix it in the nushell code, we can land this PR.

book/commands/input.md
HoLLy-HaCKeR: I've undone the change to this line and opened pr nushell/nushell#6716

fdncred: thanks

PR #636, opened 2022-10-15
Update working_with_strings.md

zephaniahong:
Hi! 
As the example for trimming strings used the short-form flags, I thought it'd be good to just mention it prior to avoid confusion for newer users.


rgwood:
Good idea. Thanks!

PR #637, opened 2022-10-15
Update working_with_lists.md

zephaniahong:
The output comment seems to be in the wrong code block.


sholderbach:
Good catch! Thanks

PR #638, opened 2022-10-15
Update working_with_lists.md

zephaniahong:
The previous echo result displayed a list [total =, 15] which I think might not be what the user expects. I changed it to use string interpolation which results in `total = 15`.


fdncred:
thanks

PR #641, opened 2022-10-18
Add note about subexpression with external commands

WindSoilder:
I've faced the relative issue about using subexpression with external commands, and also find that someone else is facing the same problem too: 

Relative: https://github.com/nushell/nushell/issues/5898
Discord message: https://discord.com/channels/601130461678272522/601130461678272524/1022757912130633738
https://discord.com/channels/601130461678272522/614593951969574961/959930011756924959

So I think it's worth to note about it


fdncred:
agreed. i wonder if there is a way to know if we should auto-trim?

WindSoilder:
I think there is a way, when we're buiding Expression during parsing, we can know if the expression is sub-expression.  Then we can make auto-trim logic

PR #642, opened 2022-10-18
Add documentation for completions with descriptions

jmoore34:
Document the ability to return a list of `{value: "", description: ""}` records from a completion function.

context: https://discordapp.com/channels/601130461678272522/958067223187062834/1002353516515442759


sholderbach:
Awesome thanks!

PR #643, opened 2022-10-19
update term size command docs

chdalski:
Update the term size command documentation to the help output of the command in version 0.70.0

Signed-off-by: Christoph Dalski <chdalski.coding@gmail.com>

### DIFF
--- a/book/commands/term_size.md
+++ b/book/commands/term_size.md
@@ -1,6 +1,6 @@
 ---
 title: term size
-version: 0.69.1
+version: 0.70.0
 platform: |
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks but changes to command help messages need to take place in the nushell repo for the command itself because these help files are systematically generated and therefore will get overwritten the next time we generate them.

chdalski:
Ah I see - than this one can be closed :+1: thank you.

fdncred:
Please feel free to change the help message in the command though so that this is taken care of.

PR #645, opened 2022-10-22
Node script that automatically generates `book/commands` MD pages from Nushell's own `help` commands

webbedspace:
This Node script `update_book_commands.cjs` does the following when run:
* Run `nu -c "help commands | where not is_custom | select name category | group-by name | to json"` to obtain a list of all Nushell commands
* For each command, run `nu -c "help ${name} | ansi strip"` to obtain the full help file for that command (plus overloads).
* Generate and save a .MD file for that command in `book/commands`.

This has the following advantages:
* `book/commands` pages can be easily synched up with Nushell (provided the maintainer is running the current version)
* Removed commands are no longer clogging up the `book/commands` folder.
* Labels for each section of the `book/commands` files (such as "Signature" vs "Usage") are now fully coordinated.

Incidentally, while writing this, I added a column to Commands Reference showing the categories of each command, as helpful additional information.


rgwood:
Thanks for the PR! I think most of this is already covered by the make_docs.nu script: https://github.com/nushell/nushell.github.io/blob/main/book/make_docs.nu

webbedspace:
Well, that script isn't mentioned anywhere in the repository's README, isn't in the repository's bottom level, nor in any of the four tasks in `package.json`…………… 🦟

It also seems like it hasn't been run by anyone in a long time, considering that dead commands like https://www.nushell.sh/book/commands/gstat.html are still online as we speak…

rgwood:
IIRC the current script doesn't delete commands. But yeah, it probably should, and it would be good to make its existence more obvious.

It's a little unfortunate that the README.md doubles as the website front page, makes documenting things like that harder.

webbedspace:
> IIRC the current script doesn't delete commands.

https://github.com/nushell/nushell.github.io/blob/f2a8b963f13f41621fa712360f7bbb9fb1410cea/book/make_docs.nu#L12

fdncred:
gstat is not dead. It's a plugin. The script to regenerate the docs is ram with each release. You may see that in the closed prs.

webbedspace:
Thanks for the explanation.

PR #646, opened 2022-10-22
Add categories column to command_reference

webbedspace:
LET'S TRY THIS AGAIN

`book/command_reference` now lists the categories of each command alongside their names and descriptions, so as to better convey which class of problems they are meant to be used with.

Also reran `make_docs.nu`.


fdncred:
You should build and register all plugins prior to running the script so that we have a reference for all the commands.

rgwood:
> Also, the make docs script was just ram two days ago.

Main point of this improvement is adding the categories - that's new.

.prettierignore
```diff
@@ -1,12 +1,12 @@
...
-
-operators.md
-dataframes.md
-nushell_map.md
-*_bash*.md
-book/commands/*.md
-.vuepress/.temp
-.vuepress/.cache
-.vuepress/dist/
+
```
rgwood, line 1: Did you mean to update this file? The GitHub web UI doesn't make it obvious what changed (probably line endings?)

rgwood:
Thanks for this! 

Is the package.json change intentional?

.prettierignore
webbedspace: Just line endings… I think. I didn't touch it myself.

rgwood:
> Is it OK for the docs to list commands which aren't ready to use in the stock Nushell distribution??

Hmm, good question. I'm not sure we've discussed this, but I guess I would lean toward leaving plugin commands out of the online docs.

webbedspace:
> Is the package-lock.json change intentional?

I don't fully understand how package-lock works in NPM, but I was under the impression that always committing it as-is was the standard. I could revert it if you want.

rgwood:
> I was under the impression that always committing it as-is was the standard.

I think my rule of thumb is to only commit a lockfile change if the change was intentional.

> I could revert it if you want.

Yes please; please remove the unintentionally modified files from the PR.

Otherwise it looks good to go, once we sort out the question about plugins.

fdncred:
> Main point of this improvement is adding the categories - that's new.

@rgwood It's also deleting commands, which I'm not in favor of. 

deleted
------
exec
gstat
inc
query
query json
query web
query xml

added
-----
register query

fdncred:
> ? Is it OK for the docs to list commands which aren't ready to use in the stock Nushell distribution??

Yes. What gives you the impression that plugins aren't ready to use?

rgwood:
I think @webbedspace is saying that plugins aren't ready to use out-of-the-box with Nushell; some setup is required.

IMO it's reasonable to exclude plugin commands from the official command list, but I don't feel too strongly about it.

rgwood:
To move this forward, perhaps we just leave the plugin command docs as-is for now?

@webbedspace if you can revert the file deletions I think this will be in a good place to merge

rgwood:
Thanks for the PR!

PR #647, opened 2022-10-23
Release notes for nushell 0.71

sholderbach:
Please add your important new features and breaking changes to the release notes by commiting to/opening a PR against the `release-notes-0.71` branch! Thanks for helping out!


sholderbach:
@fdncred @merelymyself @PerBothner @webbedspace @rgwood 
I saw that you made some user facing changes that might be worth talking about in the release notes. Feel free to fill in the sections with a few short sentences or a good code example. If you don't have time the team will probably still be able to fill in the blanks (with maybe a less perfect example)

@dandavison do we need to document things about the change to the completion logic (for users that take the new default config or those that run the old default config)?

PerBothner:
If you're OK with a plug for DomTerm, how about the following? If you want I can also add a screenshot.

## How nushell outputs values can now be customized via `display_output` hook (PerBothner)

This new (somewhat experimental)  hook is useful for modifying how values are printed or diverting the output to a different window. This could be an IDE or "notebook" interface that can handle "rich output" such as HTML.
See [this section of the book](https://www.nushell.sh/book/hooks.html#basic-hooks#filtering-or-diverting-command-output).

A motivating example was to use inline HTML in [the DomTerm terminal emulator](https://domterm.org).
```let-env config = ($env.config | upsert hooks {
    if "DOMTERM" in (env).name {{
        display_output: { to html --partial --no-color | domterm hcat  }
    }} else {{ }}
```

PerBothner:
_If you feel a screenshot would be good, please add the following after my previous comment:_

dt-nu-table

The screenshot shows the output from the `du` and `ls Make*` commands.  (The first table is mostly scrolled out, but the the header is shifted so as to keep it visible.)

dandavison:
I'm not sure I'm going to have time to write a section on completion unfortunately :/

blog/2022-11-08-nushell-0.71.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+## General consistency accross commands and types has improved:
+### More iterating filter commands now support the `$in` variable consistently (webbedspace)
+https://github.com/nushell/nushell/pull/6951
+### `List`s finally support `upsert` (fdncred)
+https://github.com/nushell/nushell/pull/6932
+### `merge` can now also operate on records (webbedspace)
+https://github.com/nushell/nushell/pull/6919
```
webbedspace, line 38: Add this:
"
Formerly, `merge`'s type signature only accepted, as an argument, blocks returning records instead of just plain records. This restriction, which conflicted with `append`'s signature, has now been lifted. It is also likely that, due to the fact that little expressive power is gained from having `merge` accept blocks returning records, that that part of its type signature will be removed in a future version.
"

blog/2022-11-08-nushell-0.71.md
```diff
@@ -0,0 +1,82 @@
...
+- The column name for the command name in `$nu.scope.commands` has been changed from `command` to `name`. Thus, `$nu.scope.commands | where name =~ zip` would give you specific command info. ([#7007](https://github.com/nushell/nushell/pull/7007))
+- The `str distance` command now directly returns an `Int` instead of a record `{"distance": <Int>}` ([#6963](https://github.com/nushell/nushell/pull/6963))
+- The argument order of `format filesize` was made consistent with other commands that accept a `CellPath` to specify fields that should be affected by it. Now the string specifying the format is the first positional argument and the cell path is specified in the following positional argument(s). ([#6879]https://github.com/nushell/nushell/pull/6879)
+- The `--perf` flag for Nu was removed as part of an effort to simplify diagnostics logging [#6834](https://github.com/nushell/nushell/pull/6834)
+
+- @dandavison completion defaults
+
+# Next Steps
+
+We're continuing to work on language updates and hope to have more information for you soon. If you'd like to become part of the effort, please [join the discord](https://discord.gg/NtAbbGn).
```
webbedspace, line 80: Hasn't the "hope to have more information for you soon" line been in the past several release notes? What does it refer to? The 0.80 draft roadmap?

webbedspace (changes requested):

PR #648, opened 2022-10-25
docs: 📝 align with #6904

melMass:
document the changes in #6904
defaults to not overwriting | added a force flag


hustcer:
Thanks for your contribution, the command docs will be update by `book/make_docs.nu` script at the next nushell release, so we need not modify it manually.

PR #649, opened 2022-10-26
fix description of build-string's second example

kambala-decapitator:


fdncred:
thanks

webbedspace:
@fdncred Why was this merged if it only changes an autogenerated file? (i.e. the `/book/commands/*` files) The correct thing to do is to change the Nushell `help` command for `build-string`, right?

fdncred:
@webbedspace because the rust code change already happened in the build-string source code.

PR #650, opened 2022-10-26
add right prompt release notes

nibon7:


fdncred:
@nibon7 can you say what false does and perhaps have a picture of false too?

fdncred:
thank you!

PR #652, opened 2022-10-26
Fixed various issues in make_docs.nu

webbedspace:
* Certain hard-coded commands' `extra_usage` is wrapped in code block markup to prevent their code from being interpreted as markdown, instead of every command's `extra_usage` being wrapped.
* No longer deletes files due to erasing platform-specific commands.
* Notes now appear before parameters, in accordance with the actual nu `help`.
* Added task to `package.json` to run it.
 
~~Also reran make_docs.nu.~~

### DIFF
--- a/book/make_docs.nu
+++ b/book/make_docs.nu
@@ -94,21 +94,10 @@ def get-doc [command] {
   let ex = $command.extra_usage
   # Certain commands' extra_usage is wrapped in code block markup to prevent their code from
   # being interpreted as markdown. This is strictly hard-coded for now.
-  let extra_usage = if $ex == "" { "" } else if $command.command in ['def-env' 'export def-env' 'as-date' 'as-datetime' ansi] {
-    $"## Notes(char nl)```text(char nl)($command.extra_usage)(char nl)```(char nl)"
+  let extra_usage = if $ex == "" { "" } else if $command.name in ['def-env' 'export def-env' 'as-date' 'as-datetime' ansi] {
+    $"## Notes(char nl)```text(char nl)($ex)(char nl)```(char nl)"
   } else {
-    ""
-  }
-
-  let parameters = if $no_param { "" } else { $"## Parameters(char nl)(char nl)($params)(char nl)(char nl)" }
-
-  let ex = $command.extra_usage
-  # Certain commands' extra_usage is wrapped in code block markup to prevent their code from
-  # being interpreted as markdown. This is strictly hard-coded for now.
-  let extra_usage = if $ex == "" { "" } else if $command.command in ['def-env' 'export def-env' 'as-date' 'as-datetime' ansi] {
-    $"## Notes(char nl)```text(char nl)($command.extra_usage)(char nl)```(char nl)"
-  } else {
-    $"## Notes(char nl)( $ex )(char nl)(char nl)"
+    $"## Notes(char nl)($ex)(char nl)"
   }
 
   let examples = if ($command.examples | length) > 0 {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "repository": "github:nushell/nushell.github.io",
   "type": "module",
   "scripts": {
+    "update": "nu book/make_docs.nu",
```
sholderbach, line 8: Interesting that this didn't conflict with the changes in the commit. 
Mhh wouldn't the command reference then be independent of what is in the folder on the repo?

webbedspace: I'm not sure what you're asking. The idea with this command is that it assumes whoever runs `npm run update` always has the most recent publicly-released version installed.

sholderbach: Ah sorry, I have to admit that I am not really familiar with the build process for github pages. I assumed this would be run on deploy. Having this as a shorthand certainly makes it easier.

fdncred:
@webbedspace @sholderbach Are we ready and do we need to land this before it runs again after the release?

webbedspace:
I decided that having all those re-compiled `commands` pages inside the PR was making it unwieldy, so I've removed them, leaving just the `make_docs.nu` and `package.json` changes. Those can be applied whenever it becomes convenient to re-run `make_docs.nu`, instead of just whenever this PR is accepted.

book/make_docs.nu
```diff
@@ -89,16 +91,24 @@ def get-doc [command] {
...
-    # TODO: find a better way to display plain text with minimal formatting
-    $"## Notes
-```text
-($command.extra_usage)
-```
-"
+  let ex = $command.extra_usage
+  # Certain commands' extra_usage is wrapped in code block markup to prevent their code from
+  # being interpreted as markdown. This is strictly hard-coded for now.
+  let extra_usage = if $ex == "" { "" } else if $command.command in ['def-env' 'export def-env' 'as-date' 'as-datetime' ansi] {
```
hustcer, line 97: Why `ex` and `extra_usage` defined twice in line 105 and 108?

hustcer, line 97: Can we define `['def-env' 'export def-env' 'as-date' 'as-datetime' ansi]` as a const variable?

book/make_docs.nu
```diff
@@ -89,16 +91,24 @@ def get-doc [command] {
...
+    $"## Notes(char nl)```text(char nl)($command.extra_usage)(char nl)```(char nl)"
```
hustcer, line 98: Can `$command.extra_usage` be replaced by `$ex`?

book/make_docs.nu
webbedspace: That was a mistake when rebasing, thanks.

book/make_docs.nu
```diff
@@ -8,8 +8,10 @@ if $book_exists == false {
   exit --now
 }
 
-# Clean all commands and regenerate the docs
-do -i { rm book/commands/*.md }
+# Old commands are currently not deleted because some of them
+# are platform-specific, and a single run of this script will not regenerate
```
hustcer, line 12: Looks good to me now, just one last question: about 'some of them are platform-specific', can you list which ones？

webbedspace: It came up awhile ago, but `registry query` is Windows-only, and a few others (can't remember which) are POSIX-only.

hustcer: I will register all the necessary plugins and run `source book/make_docs.nu` to generate the docs, that won't miss the docs of plugins

sholderbach (approved):

hustcer:
Thanks

PR #654, opened 2022-10-31
Mention nushell also run on BSD systems

rapenne-s:
It's not much but I imported nushell into OpenBSD packages, so the website can now include BSD in the operating system list where it's known to run.

I think it's worth mentioning because it implies nushell is really portable, but as it's too niche I prefered not adding any extra info with regard to installation, and it's already covered by package manager / source install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ actionText: Get Started →
 actionLink: /book/
 features:
   - title: Pipelines to control any OS
-    details: Nu works on Linux, macOS, and Windows. Learn it once, then use it anywhere.
+    details: Nu works on Linux, macOS, BSD and Windows. Learn it once, then use it anywhere.
   - title: Everything is data
     details: Nu pipelines use structured data so you can safely select, filter, and sort the same way every time. Stop parsing strings and start solving problems.
   - title: Powerful plugins
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ actionText: Get Started →
 actionLink: /book/
 features:
   - title: Pipelines to control any OS
-    details: Nu works on Linux, macOS, BSD and Windows. Learn it once, then use it anywhere.
+    details: Nu works on Linux, macOS, BSD, and Windows. Learn it once, then use it anywhere.
   - title: Everything is data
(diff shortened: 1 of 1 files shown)


sholderbach:
Awesome! Feel free to add recommendations to the installation instructions if you have a particular package source proven out or there are any special notes. 
As we do run our tests only on the three big platforms we might not always catch isssues (on our side or with our dependencies) that might arise from slight differences in system calls / fs between linux and the BSDs. I am not familiar with the BSD world, are there significant differences between OpenBSD, FreeBSD etc. that it makes sense to specify OpenBSD as the one you vetted?

sholderbach:
OK apparently building on FreeBSD also succeeds recently: https://github.com/nushell/nushell/pull/6456

There seems to be a slight limitation for the `ps` command (https://github.com/nushell/nushell/pull/6456/files#diff-bd1f8d1c03c0b2cbfc9322d28fe71ff96b94dea93b435a0117a5305fb950157d) and locale management (which we are still behind the curve on) might also have some extra complications: https://github.com/nushell/nushell/pull/6085

rapenne-s:
`pkg_add nushell` will be available soon for the development version users, but only in 6 months for the release users, so I prefer not to mention it anyway.

I didn't build it with extra features, I got error but I don't really know the rust ecosystem so I need to figure what's missing or what must be fixed, and I realize I don't have the built-in `ps` command :thinking: if I understand the FreeBSD pr correctly, `ps` was explicitly disabled for the systems not part of android/linux/windows/mac, it would require a rewrite of the function I suppose. But not having ps isn't not a deal breaker IMO.

rapenne-s:
> I am not familiar with the BSD world, are there significant differences between OpenBSD, FreeBSD etc. that it makes sense to specify OpenBSD as the one you vetted?

BSD systems have a ~30 years old common ancestor IIRC, but they are definitely different systems. If it's fine for you, I can modify my PR to list OpenBSD and FreeBSD instead of just `BSD`, it's better to be accurate.

sholderbach:
Thank you very much for your amazing blogpost! It created quite a buzz and certainly introduced nushell to a bunch of new people.

Sorry, forgot to move ahead with the PR. I think it is fine like that and we can add platform specific notes or guides at any later date.

PR #656, opened 2022-11-04
Create a section for the 0.80 philosophy

jntrnr:
This is very much a work-in-progress and welcomes comments.

### DIFF
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -29,7 +29,7 @@ In cases where the user would reasonably have an expectation of the functionalit
 The intent of Nushell's design is to be a scripting language. Scripting can take many forms, from simple pipelines to large scripts. Nushell should handle these with ease by meeting reasonable expectations in terms of:
 - Modularity
 - Readability
-- Programming language features and convenience\
+- Programming language features and convenience
 
 ### Structured data
 
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -10,7 +10,7 @@ Designs should work towards serving this goal. Those that don't should be remove
 
 ### Shell-first
 
-Nushell should work to serve its roll as a shell and a language with a focus on shells. Shell types of activities include, but aren't limited to:
+Nushell should work to serve its role as a shell and a language with a focus on shells. Shell types of activities include, but aren't limited to:
 
 * Running commands
 * Redirecting stdout/stdin/stderr
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -42,7 +42,22 @@ Values in Nushell can be records, lists, tables, binary data, and more. Being ab
 Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell enables composition via the use of pipes (`|`), just like Unix pipelines.
 
 Composing commands, both built-in and user-defined, is a core piece of Nushell. The design of Nushell and its standard library must support both easily composing commands as well as allowing the user to easily create compose-able commands.
+### Command philosophy
 
+Specifically, Nushell's philosophy about commands is represented by the following positions (these apply especially to the library of builtin commands distributed with nushell, but are also good guidance for your own commands and plugins):
+
+1. There should be one -- and preferably only one -- obvious way to do it.
+  You will probably recognize this from the "[Zen of Python](https://peps.python.org/pep-0020/)"!  
+
+2. The aim of our library of builtin commands is to provide a collection of simple and composable primitive commands, that covers essential shell programming needs and allows virtually anything to be built via composition and creation of custom user commands and plugins.
+  That's as opposed to maintaining an extensive catalog of more specialized commands in the core nushell codebase: more specialized functionality should exist as community-contributed libraries (e.g. plugins/nushell libraries).
+
+3. If something can be done conveniently by composition of simpler commands then we do not add flags and options to do the same thing.
+
+4. The primary input to a command should generally be supplied as "input", not as a positional argument.
+  This is what enables pipeline composition.
+
+5. A given command, on a given input, should generally always produce the same type of output data structure; the presence/absence/value of arguments and flags shouldn't change the output type.
 ## Command signatures and their parts
 
 ### Signature
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -18,9 +18,8 @@ Nushell should work to serve its role as a shell and a language with a focus on 
     * Ctrl-C
     * Ctrl-D
     * and others
-* Unix-only: Handling background tasks
-    * Ctrl-Z
-    * Background/foregrounding of tasks
+* Support for background tasks
+   * In Unix, this may mean Ctrl-Z and traditional background/foregrounding of tasks or a more modern approach
 
 In cases where the user would reasonably have an expectation of the functionality of the shell, we should make every effort to include. If a feature would be a reasonable expectation (say ctrl-z on Unix), then we should have OS-specific functionality for that platform that meets that expectation.
 
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -119,8 +119,7 @@ The core language and standard library needs to cover the following categories t
 * Querying, filtering, and manipulating structured data
 * Network connectivity
     * note: Network support is fundamental because with it users can easily acquire and install Nushell extensions
-* Basic formats:
-    * CSV and JSON support
+* Basic formats (exact list to be determined by common use cases)
 * Basic date support
 
 The following categories should be moved to plugins:
@@ -128,8 +127,7 @@ The following categories should be moved to plugins:
 * Database connectivity
 * Dataframe support
 * Hash functionality
-* Format support
-    * Any non-CSV or JSON file format
+* Uncommon format support
 * Experimental commands
 * Binary data functionality
 * Random number support
@@ -158,6 +156,8 @@ Bash logic operators:
 cat foo.txt && cat bar.txt
 cat foo.txt || cat bar.txt
 ```
+
+We don't have plans to support the full bash language, only the above capabilities that would be common muscle memory for most bash users.
 
 ### Limited mutation
 
@@ -190,9 +190,8 @@ Closure example:
 ls | each { |x| $x.name + "foo" }
 ```
 
-### Math without spaces
+## Extensions
 
-It will be possible to not require spaces in math. For example, `1+1` instead of requiring `1 + 1`. While a small change, this is part of a set of changes to make the language feel a little less surprising to people coming from other languages.
+We will need to design and implement an easy-to-use extension mechanism that allows people to extend Nushell with additional functionality. This will be important both for helping shrink the core Nushell down but also to make it easy for contributors to create and share their own extensions.
 
-*JT: Is this more important than having nice looking dates? This doesn't seem to be as important now if we lean into a shell-first focus.*
-
+The extension mechanism will need a standard form by 1.0. For 0.80, we should have our best guess at a complete solution that meets the design goals.
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -39,7 +39,7 @@ Values in Nushell can be records, lists, tables, binary data, and more. Being ab
 
 ### Pipelines
 
-Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell leverages this composition via the use of pipes (`|`), just like Unix pipelines.
+Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell enables composition via the use of pipes (`|`), just like Unix pipelines.
 
 Composing commands, both built-in and user-defined, is a core piece of Nushell. The design of Nushell and its standard library must support both easily composing commands as well as allowing the user to easily create compose-able commands.
 
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -128,9 +128,9 @@ The following categories should be moved to plugins:
 * Hash functionality
 * Uncommon format support
 * Experimental commands
-* Binary data functionality
-* Random number support
-* Math support
+* Supplemental random number support (eg `random dice`)
+* Supplemental math support from `math`
+* Supplemental binary data functionality from `bytes`
 * Advanced date support
 
 We should select commands to be in the core categories which meet the common use cases for Nushell. Commands that are in the core categories that are uncommon use cases should move to optional extensions.
--- a/contributor-book/philosophy_0_80.md
+++ b/contributor-book/philosophy_0_80.md
@@ -1,0 +1,182 @@
+# Nushell design philosophy
+
+## Core philosophy
+
+Nushell is "A shell-first scripting language for working with structured data flowing through pipelines".
+
+Designs should work towards serving this goal. Those that don't should be removed or moved to optional add-ons.
+
+## Core design
+
+### Shell-first
+
+Nushell should work to serve its roll as a shell and a language with a focus on shells. Shell types of activities which include, but aren't limited to:
+
+* Running commands
+* Redirecting stdout/stdin/stderr
+* Unix-only: Properly handling signals
+    * Ctrl-C
+    * Ctrl-D
+    * and others
+* Unix-only: Handling background tasks
+    * Ctrl-Z
+    * Background/foregrounding of tasks
+
+In cases where the user would reasonably have an expectation of the functionality of the shell, we should make every effort to include. If a feature would be a reasonable expectation (say ctrl-z on Unix), then we should have OS-specific functionality for that platform that meets that expectation.
+
+### Scripting language
+
+The intent of Nushell's design is to be a scripting language. Scripting can take many forms, from simple pipelines to large scripts. Nushell should handle these with ease by meeting reasonable expectations in terms of:
+- Modularity
+- Readability
+- Programming language features and convenience\
+
+### Structured data
+
+In Nushell, all data is "structured". For Nushell, this means that values can take shapes beyond just a simple string.
+
+Values in Nushell can be records, lists, tables, binary data, and more. Being able to convert into structured data and work with structured data is fundamental to Nushell.
+
+### Pipelines
+
+Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell leverages this composition via the use of pipes (`|`), just like Unix pipelines.
+
+Composing commands, both built-in and user-defined, is a core piece of Nushell. The design of Nushell and its standard library must support both easily composing commands as well as allowing the user to easily create compose-able commands.
+
+## Command signatures and their parts
+
+### Signature
+
+The signature of the command describes the following:
+
+- The name of the command
+- The usage information (documentation) for the command
+- The name and type of:
+    - Named parameters
+    - Positional parameters
+- The type of:
+    - Input
+    - Output
+
+### When to use input
+
+Input is intended for one or both of:
+- Pipeline composition
+- Streams of data
+
+### When to use output
+
+All return values are output.
+
+### When to use positional parameters
+
+Use a positional parameter when:
+- A parameter is required
+
+### When to use rest parameters
+
+Use a rest parameter when:
+- A command takes some number of additional optional arguments of the same type
+
+### When to use named parameters
+
+Use a named parameter (a flag parameter) when:
+- The value you need is optional
+
+### When to use a switch flag
+
+Use a switch flag when:
+- You need to allow the user to change the default action of the command
+
+### Comments
+
+Commands should be commented to describe their function and usage. This documentation should also cover parameters and their use.
+
+## Core categories
+
+The core language and standard library needs to cover the following categories to support common use cases for Nushell:
+
+* Filesytem
+* Operating system
+* Manipulating the environment
+* Parsing string data into structure data
+* Querying, filtering, and manipulating structured data
+* Network connectivity
+    * note: Network support is fundamental because with it users can easily acquire and install Nushell extensions
+* Basic formats:
+    * CSV and JSON support
+* Basic date support
+
+The following categories should be moved to plugins:
+
+* Database connectivity
+* Dataframe support
+* Hash functionality
+* Format support
+    * Any non-CSV or JSON file format
+* Experimental commands
+* Binary data functionality
+* Random number support
+* Math support
+* Advanced date support
+
+We should select commands to be in the core categories which meet the common use cases for Nushell. Commands that are in the core categories that are uncommon use cases should move to optional extensions.
+
+## Language design
+
+Rather than describe the whole of the language here, this section describes the changes expected to come as part of 0.80 and how they differ from the 0.60 series.
+
+### More bash-ism
+
+Being a shell-focused language means incorporating more of the expected shell-features into the language. These include:
+
+Redirection (with common variations):
+```
+cat foo.txt > bar.txt
+cat foo.txt >> bar.txt
+cat foo.txt 2> bar.txt
+```
+
+Bash logic operators:
+```
+cat foo.txt && cat bar.txt
+cat foo.txt || cat bar.txt
+```
+
+### Limited mutation
+
+In 0.80, there will be a limited form of mutation that works with the local command.
+```
+mut x = 100
+$x = 200
+print $x
+```
+
+This will help with some patterns where people wanted to calculate something in a loop but didn't have an easy way to do so previously.
+
+### Splitting closures and blocks
+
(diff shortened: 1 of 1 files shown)


fdncred:
Just a note of process here, when we land this, we'll also want to make changes to the site so that the contributor book is findable again, it's hidden now. I think it's commented out in a few places.

I'm also thinking that it may be good to link to the contributor book somewhere from the book, maybe 'thinking in nu' section?

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+Commands should be commented to describe their function and usage. This documentation should also cover parameters and their use.
+
+## Core categories
+
+The core language and standard library needs to cover the following categories to support common use cases for Nushell:
+
+* Filesytem
+* Operating system
+* Manipulating the environment
+* Parsing string data into structure data
```
dandavison, line 102:
```suggestion
* Parsing string data into structure data
* Formatting structured data as standard string formats (CSV and JSON)
```

Suggestion is to replace the CSV/JSON entry below with this.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+* Querying, filtering, and manipulating structured data
+* Network connectivity
+    * note: Network support is fundamental because with it users can easily acquire and install Nushell extensions
+* Basic formats:
+    * CSV and JSON support
+* Basic date support
+
+The following categories should be moved to plugins:
```
dandavison, line 110: Yes, I think so! I know this is a big change, but I think it has the potential to move us into a new era of a v2 builtins stdlib that is really starting to be on-course for 1.0.

I'm not super familiar with the plugin technology. As an aside, eventually it will be appropriate for people to implement libraries of utilities for the public in nushell itself rather than Rust plugins.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+cat foo.txt > bar.txt
+cat foo.txt >> bar.txt
+cat foo.txt 2> bar.txt
+```
+
+Bash logic operators:
+```
+cat foo.txt && cat bar.txt
+cat foo.txt || cat bar.txt
+```
```
dandavison, line 144: I definitely agree with those two categories of bashisms entering the language.

I wouldn't want to see much more though without plenty of justifying design discussion.

dandavison (commented):
I think this is a great start on nailing down some design philosophy behind current and future nushell. I'll try to make some similar / complementary notes myself.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+### Scripting language
+
+The intent of Nushell's design is to be a scripting language. Scripting can take many forms, from simple pipelines to large scripts. Nushell should handle these with ease by meeting reasonable expectations in terms of:
+- Modularity
+- Readability
+- Programming language features and convenience\
+
+### Structured data
+
+In Nushell, all data is "structured". For Nushell, this means that values can take shapes beyond just a simple string.
```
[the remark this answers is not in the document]
rgwood: I think of the type information as providing structure.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+
+* Filesytem
+* Operating system
+* Manipulating the environment
+* Parsing string data into structure data
+* Querying, filtering, and manipulating structured data
+* Network connectivity
+    * note: Network support is fundamental because with it users can easily acquire and install Nushell extensions
+* Basic formats:
+    * CSV and JSON support
```
rgwood, line 107: I think a case could be made for keeping TOML/YAML/XML in the core as well. But I'd be open to just CSV+JSON.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+
+* Database connectivity
+* Dataframe support
+* Hash functionality
+* Format support
+    * Any non-CSV or JSON file format
+* Experimental commands
+* Binary data functionality
+* Random number support
+* Math support
```
rgwood, line 120: IMO these 3 feel like they should be in core. If Nu can't do math or work with binary data out of the box, that feels like a shortcoming.

rgwood:
Thank you for putting this together!

Given that this proposal would move a lot of functionality into plugins, I would like to think a bit about the UX for acquiring plugins.

Let's imagine a scenario where we've moved Excel support into a plugin, and a user has just installed Nu as a single executable. **What steps does the user need to do to acquire the plugin with `open foo.xlsx` / `from xlsx` support?** Some high-level ideas:

### 1. Fully or partially automatic plugin installation

When the user runs `open foo.xlsx`, Nu prompts the user: "A plugin exists for .xlsx files. Would you like to install it?" If the user chooses "yes", Nu downloads+installs the plugin.

### 2. Single-command installation

The user can run `install plugin excel` (or similar) to download+install the plugin. Nu automatically downloads the right version of the plugin compiled for the right OS+architecture, and saves it in a sensible location on disk.

### 3. Manual installation

Like today, the user needs to download a plugin file and then `register` it. The onus is on the user to make sure they get a file with the right OS+architecture.

## Why does this matter?

Currently there is a fair amount of friction involved in installing plugins. Going from a single executable to multiple files and a registration command is a big increase in deployment/installation complexity!

I worry that if we move lots of functionality into plugins without making it easier to install plugins, the functionality in plugins will be underused. I know that I've occasionally avoided Nu plugins because I can't be bothered.

## What should we do?

I think I lean toward option 2; it feels like a sweet spot between simplicity and avoiding magical behavior. 

Note that I haven't touched on implementation at all; I think running our own plugin registry for this will be a lot of work. Worth digging into after we have a rough consensus on the UX.

fdncred:
@rgwood nice writeup! I'd vote for number 1 having a message telling you to do number 2. I'd prefer to stay less magical initially.

I think I'd want to augment the number 2 option with something like `install plugin --all` or some such thing to get all the officially supported nushell plugins.

Then I'd also want to have some such thing to list all the plugins like `install plugin --list` that would return something that we can see as a nushell table.

I don't want to get too fancy right off the bat but I think installing all plugins and seeing what plugins are available is important to a good UX.

contributor-book/philosophy_0_80.md
FilipAndersson245: +1 for TOML

rgwood:
I wrote some more about the design+implementation of `install plugin`: https://hackmd.io/@nucore/Bklhq5VHs

There are a lot of open questions. I'm happy to add some more info about the subject to this document once I have a better understanding of the way forward.

contributor-book/philosophy_0_80.md
jntrnr: Agreed - I think we want to catch the most common bash-isms without trying for some kind of full bash-compatibility.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,183 @@
...
+In Nushell, all data is "structured". For Nushell, this means that values can take shapes beyond just a simple string.
+
+Values in Nushell can be records, lists, tables, binary data, and more. Being able to convert into structured data and work with structured data is fundamental to Nushell.
+
+### Pipelines
+
+Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell leverages this composition via the use of pipes (`|`), just like Unix pipelines.
+
+Composing commands, both built-in and user-defined, is a core piece of Nushell. The design of Nushell and its standard library must support both easily composing commands as well as allowing the user to easily create compose-able commands.
+
```
dandavison, line 45:
```suggestion
### Command philosophy

Specifically, Nushell's philosophy about commands is described by the following positions:

1. There should be one -- and preferably only one -- obvious way to do it.
  You will probably recognize this from the "[Zen of Python](https://peps.python.org/pep-0020/)"!  

2. If something can be done conveniently by composition of simpler commands then we do not add flags and options to do the same thing.

3. The primary input to a command should generally be supplied as "input", not as a positional argument.
  This is what enables pipeline composition.

4. A given command, on a given input, should generally always produce the same type of output data structure; the presence/absence/value of arguments and flags shouldn't change the output type.
```

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,183 @@
...
+### Structured data
+
+In Nushell, all data is "structured". For Nushell, this means that values can take shapes beyond just a simple string.
+
+Values in Nushell can be records, lists, tables, binary data, and more. Being able to convert into structured data and work with structured data is fundamental to Nushell.
+
+### Pipelines
+
+Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell leverages this composition via the use of pipes (`|`), just like Unix pipelines.
```
dandavison, line 42:
```suggestion
Nushell takes the Unix philosophy of pipelines to heart. Commands should be built with the intent of composition. Nushell enables composition via the use of pipes (`|`), just like Unix pipelines.
```

contributor-book/philosophy_0_80.md
dandavison: Sorry I dumped these here without commenting. They're just a quickly written-down attempt at some command philosophy.

jntrnr: I'll land them as is. I wrote the other text pretty loosely as well and I'm assuming we'll continue refining it before we land it.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,198 @@
...
+
+### When to use a switch flag
+
+Use a switch flag when:
+- You need to allow the user to change the default action of the command
+
+### Comments
+
+Commands should be commented to describe their function and usage. This documentation should also cover parameters and their use.
+
```
dandavison, line 109:
```suggestion
## Naming commands

Good naming is incredibly important! Here are some guidelines:

- Find out what names other well-known programming languages use for the thing you are creating (if something similar exists). Don't choose a name that differs from all other languages without a good reason.

- As always in programming, if you are uncomfortable with the name chosen for something, stop! It often means there's something wrong with your design. Maybe it's doing too much? Or maybe the concept it's representing is still a bit confused?

- Don't expect it to be easy -- it may take some time to settle on the right name for a command (and its options, if it needs any). Don't hesitate to come and talk to other Nushell contributors in Discord, or discuss it in GitHub.

- Find out what names Nushell uses for related things -- your name should be harmonious with those. E.g. if `head` already exists, maybe it makes sense for another command called `tail` to exist? (Or maybe the thing called `head` shouldn't have that name! Naming is hard so don't assume current names are perfect or immutable.)

- Remember that when you're choosing a name, you're making a decision that affects the code of every Nushell user in the world. (Of course, this means that fixing bad naming decisions will break the code of Nushell users and force them to update. We still won't always get it right the first time, but this is one reason that we put so much emphasis on it.)

- If the name you feel is the "right name" is already taken, don't just settle for an inferior name -- please start a discussion about the situation instead. It may mean that there is a broader conceptual or naming problem that needs to be addressed.

- Please explain how you chose the name in your PR -- writing this will help with the process of naming it!

The ultimate goal here is to create an ecosystem of commands and terminology that is harmonious, and "feels right". Yes, this is both highly subjective and hard work! But it is an extremely important factor determining how enjoyable and powerful it is to work in the programming environment that we are creating, and ultimately it is this that will determine the success of the project.
```

PerBothner:
As I discuss in issue https://github.com/nushell/nushell/issues/6787, I believe the way Nu auto-generates `config.nu` and `env.nu` is an "anti-pattern". Changing this would be a good thing to put on the roadmap.

contributor-book/philosophy_0_80.md
fdncred: I agree with the sentiment here. Indeed, naming can be difficult, but REnaming is frequently trivial. I'd rather see some awesome wow thing that changes the way we think about things, that is named badly, then have folks struggling for hours trying to come up with the "right" name and never really starting to code. Try to do all the things above but don't stop coding because you don't have the perfect name for something.

Here's a real-world example. zhiburt, who is the author of tabled and a frequent contributor to nushell recently introduced a PR to show off some cool tui things with horizontal and vertical scrolling. The initial name was tabless, kind of clever like table and less combined, I get it. But the point of the PR wasn't the name, it was the total awesomeness that can come with a tui. That thing could've been named bob initially and I would've been equally blown away with it. It truly was a beautiful demo and surprise. I tried to encourage him, and we go back and forth about features and he kept on chugging away. Then someone objected to the name and suggested scroll. zhiburt changed it in a few minutes and scroll is what it is currently named now. Is that the perfect name? I have no idea, but tbh, I'm not really concerned with the exact "right" name. I want to use the tui he introduced whatever its final name ends up being. Having said that, it's sometimes easier to spot the wrong name than to pick the right one. bob would've been the wrong name. lol. And yet again, sometimes other people can just come up with good names, but you can't because you can't see the forest for the trees.

fdncred: I agree with the sentiment here. Indeed, naming can be difficult, but REnaming is frequently trivial. I'd rather see some awesome wow thing that changes the way we think about things, that is named badly, then have folks struggling for hours trying to come up with the "right" name and never really starting to code. Try to do all the things above but don't stop coding because you don't have the perfect name for something.

Here's a real-world example. zhiburt, who is the author of tabled and a frequent contributor to nushell recently introduced a PR to show off some cool tui things with horizontal and vertical scrolling. The initial name was tabless, kind of clever like table and less combined, I get it. But the point of the PR wasn't the name, it was the total awesomeness that can come with a tui. That thing could've been named bob initially and I would've been equally blown away with it. It truly was a beautiful demo and surprise. I tried to encourage him, and we go back and forth about features and he kept on chugging away. Then someone objected to the name and suggested scroll. zhiburt changed it in a few minutes and scroll is what it is currently named now. Is that the perfect name? I have no idea, but tbh, I'm not really concerned with the exact "right" name. I want to use the tui he introduced whatever its final name ends up being. Having said that, it's sometimes easier to spot the wrong name than to pick the right one. bob would've been the wrong name. lol. And yet again, sometimes other people can just come up with good names, but you can't because you can't see the forest for the trees.

dandavison: Thanks @fdncred, I agree we should iterate on these guidelines together. It is true that renaming is trivial when something's not merged. However it's still true that if you are struggling with naming something then that often points to a problem with the design of what you're working on.

But also, what I was getting at (and which I didn't make clear) is that renaming is not at all trivial when something is merged and released. So the purpose of my bullet points was to emphasize that PRs should not be merged until multiple people are happy with the naming. Currently, there are many bad naming and API decisions in the library of Nushell builtins, but those builtins are out there in the wild being used by Nushell users and perhaps even playing a role in working software. So we'll need to make breaking changes to fix this, but we also need to change our own processes so that we get better at releasing builtins that don't need breaking changes to fix design decisions.

jntrnr: @dandavison - we'll definitely need to set up naming guidelines so we can get commands names in a predictable/readable shape for the 1.0 release. I think it might take a few revisions to get it how we want it (since I'm sure we even know just yet 😅 )

Can we split the naming piece into its own philosophy doc? That way we can land this one sooner and then add more components to it one after another as they get ready.

dandavison: > Can we split the naming piece into its own philosophy doc?

Absolutely! I just wanted to get some thoughts and draft text down on the subject for others to comment on.

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,182 @@
...
+* Running commands
+* Redirecting stdout/stdin/stderr
+* Unix-only: Properly handling signals
+    * Ctrl-C
+    * Ctrl-D
+    * and others
+* Unix-only: Handling background tasks
+    * Ctrl-Z
+    * Background/foregrounding of tasks
```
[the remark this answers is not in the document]
WindSoilder: I'm generally agree with @dandavison

Additionally, it seems the only way to manage task is spawning task by multi-processing, or we have no way to cancel task(we can't cancel a thread).  And it requires our whole engine-state to be serializable/deserializable.  Currently I found `pueue` and `terminal multiplexers` are working really fine to me.   For Ctrl-Z, sometimes I'm using this, but I don't think it's necessary either.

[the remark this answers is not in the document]
lukexor: This is probably the thing I miss most from bash, but I've found work arounds. Ctrl-z is great for backgrounding neovim and resuming without having to restart the LSP server, which for rust-analyzer has an annoyingly long startup time.

I use kitty these days for a quick new shell session, and there are neovim terminal plugins but the downside is they all take you out to a different window or force you into a smaller screen area

contributor-book/philosophy_0_80.md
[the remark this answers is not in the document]
lukexor: Agreed. In other shells, everything is just “string". Nushell at least provides a distinction while still flexible enough to fall back to String or raw binary when the structured type isn't known

contributor-book/philosophy_0_80.md
lukexor: I'm with @dandavison partially. The redirection operators from bash are notoriously difficult to use correctly, but some thought on how to do it in a nutshell way would be useful.

It's very difficult to redirect certain things in nushell that bash can do without resorting to just jumping down into bash with a do block. Specifically things like infinite streams of stderr, or splitting stderr and stdout to different files at the same time, teeing one of them, etc

contributor-book/philosophy_0_80.md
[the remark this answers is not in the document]
lukexor: > I'm generally agree with @dandavison
> 
> Additionally, it seems the only way to manage task is spawning task by multi-processing, or we have no way to cancel task(we can't cancel a thread). And it requires our whole engine-state to be serializable/deserializable. Currently I found `pueue` and `terminal multiplexers` are working really fine to me. For Ctrl-Z, sometimes I'm using this, but I don't think it's necessary either.

My only gripe with pueue so far is interactivity. Having to send commands through pueue instead of just reattaching a shell to the process is unergonomic

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,198 @@
...
+
+Nushell is "A shell-first scripting language for working with structured data flowing through pipelines".
+
+Designs should work towards serving this goal. Those that don't should be removed or moved to optional add-ons.
+
+## Core design
+
+### Shell-first
+
+Nushell should work to serve its roll as a shell and a language with a focus on shells. Shell types of activities include, but aren't limited to:
```
EjPlatzer, line 13:
```suggestion
Nushell should work to serve its role as a shell and a language with a focus on shells. Shell types of activities include, but aren't limited to:
```

contributor-book/philosophy_0_80.md
```diff
@@ -0,0 +1,198 @@
...
+    * note: Network support is fundamental because with it users can easily acquire and install Nushell extensions
+* Basic formats:
+    * CSV and JSON support
+* Basic date support
+
+The following categories should be moved to plugins:
+
+* Database connectivity
+* Dataframe support
+* Hash functionality
```
WindSoilder, line 130: I think this should keep in core too.  In bash I usually using md5sum to calculate a downloaded file to verify it's not corrupted.

contributor-book/philosophy_0_80.md
WindSoilder: I think we can have more format supported, in additional to @rgwood mentioned, I'd like to have `ssv`, `ini`, 'xlsx` supports.   But these would introduce too much dependencies into nushell core itself, I'm not sure...

WindSoilder:
There is one more thing I'm concerned about is command overloading(currently we have some commands with dataframe which involves overloading) should we support something like this?  

I'd put one vote for no for the following reason:
1. the command itself should support only some types of input, and put the prefix name like `dfr xxx` is easiler to use and read, it let's us know we're handling dataframes, although it's a bit verbosing
2. Assume we have two plugin both introduce `to-datetime` command, and they both accept `Value::int` as input, and these can easily lead to ambiguous, and it's hard to detect which plugin introduce ambiguity.

Yethal:
One thing I'm concerned about is move away from the current model where vast majority of the functionality is contained within a single statically-compiled binary towards a multi-binary model (without a builtin way to download plugins). 

* It creates more burden for human users since now aside from having to install nushell itself they will also need to install multiple plugins (of course they have to do it now as well but it's much less of an issue because vast majority of nushell's functionality is contained within the main binary)
* If nushell functionality is split between multiple binaries which binaries should be released to e.g homebrew or winget? Only the core one? Core one with all plugins? Some of them? Should each plugin be a separate homebrew/winget/aur/whatever package. 
* Same question for linux repositories once distributions start to ship nushell in their repos (or even prepackage it in installation media)
* For non-human users (pipelines, automated provisioning etc) the fact that only a single statically compiled binary needs to be present on the remote machine (that might as well be transferred over scp at the beginning of a job) makes nushell a very feasible foundation upon which future such solutions could be built. I realize that most of the existing nushell userbase is comprised of people using it in interactive mode on their own machines however all of the things that make nushell amazing for usage by humans also make it amazing (perhaps even more so) for usage by machines.

dandavison:
@Yethal the counterpoint to what you said is that, while it would be nice to create, maintain, and distribute a large amount of high-quality functionality in many different areas of the universe of command-line tools, it's impossible to do that with a small open source project.

fdncred:
@Yethal I tend to go back and forth on this, trying out different technologies like plugins to see what works. One way to maybe solve some of the points you bring up is to make commands bound to features. so, then we could have:
1. nushell + core
2. nushell + core + db
3. nushell + core + db + dataframe
4. etc, etc.

and people could download the build that they want. i'm not sure if this is a good route though because it could be more confusing perhaps.

i'm going to be using the one with `nushell + all`.

jntrnr:
@Yethal - agreed that we can't just move to the plugin style we have now, it has to be made significantly better. It should be trivial to install and use plugins, so that things like code examples largely just work. If it's any harder than a single line or even Nushell just automatically grabbing one of the "extra" style plugins when you ask for it, then it's too hard and will be a barrier for new users.

I'm hoping we can design a plugin system that's extremely easy to use. I'm chatting with folks that have worked in packaging systems and in other language projects to see what other folks think is the best practice here, but I think we should be able to make it nicer than what we have today.

I agree with @dandavison's point that we need to make the core more focused. We'll need to do that to ship a strong 1.0. As part of that, though, we'll need a very easy way to extend that core with the functionality the user needs.

rgwood:
>  I'm concerned about is move away from the current model where vast majority of the functionality is contained within a single statically-compiled binary

> it's impossible to [distribute a large amount of high-quality functionality] with a small open source project.

I agree with both of these points. This is how I'm currently thinking about the problem:

| Approach | Description | Pros | Cons  | 
|---|---|---|---|
|  Feature Minimalist | nu.exe includes very few commands. Most users need to install plugins because Nu is not very useful without them | nu.exe is tiny and quick to compile. Will incentivize us to deliver great plugin UX+tooling. | Single-binary deployment is no longer practical for most users. Hard for beginners to know which plugins to install. Will force us to spend a lot of time+energy on plugin UX+tooling. |
|  Just Right | nu.exe includes a tight, curated set of commands that is useful for, say, 80%[^1] of use cases. The other 20% need to install plugins | Great for the 80%.  | Potentially not great for the 20%. | 
|  Feature Maximalist  | nu.exe includes every command under the sun. Plugins are rarely used | Easy deployment story |  nu.exe is very big and slow to compile. The sheer number of built-in commands means it's hard for beginners to know which ones to use. Plugin system will probably languish because it's underused | 

IMO we are currently somewhere between Just Right and Feature Maximalist. I suspect that by trimming the existing command set we _might_ be able to get to Just Right. This would not preclude a better plugin system, but it might delay the need for one.

[^1]: 80/20 is just the first ratio that came to mind. Maybe the right ratio is closer to 50/50...

dandavison:
That's a great analysis @rgwood. The one thing I'd add is in Cons for Feature Maximalist I'd add something like

> The sheer number of built-in commands means that they won't all be well-named or well- designed, they won't all fit together coherently, and it will be hard for the team to handle the volume of bug reports, API design suggestions, and feature requests that they generate.

contributor-book/philosophy_0_80.md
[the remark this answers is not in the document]
jntrnr: I'll turn this into a more general "support for background tasks" as we'll probably want some solution that works reasonably well for them and hopefully one that could work crossplatform.

contributor-book/philosophy_0_80.md
jntrnr: Added some wording to limit this to just the bashisms described.

contributor-book/philosophy_0_80.md
jntrnr: I changed it so the exact formats are "to be decided" rather than listing them now. We should look to see which ones are used a lot in practice in current Nushell.

jntrnr:
Thanks for the feedback! Much appreciated.

I've folded a lot of it into revisions of this document. Looks like we have material for additional documents also: one on naming conventions and another on plugins.

Though a "batteries-included" version of Nushell would be difficult to ship at 1.0 level of quality, we should definitely consider nice ways for users to opt-in to having that experience if they need one (think meta-package of blessed extensions they can install using a one-liner).

PR #657, opened 2022-11-04
Simplified examples by removing unneeded `echo` calls

webbedspace:
Basically none of the given examples actually need `echo`, and they are improved by shortening them or replacing them with bare list `[ ]` syntax.

Also improved readability/accuracy of language comparison tables:

 * "All commands" text changed to "relevant commands".
 * Removed commands were removed.
 * Commands with no data in the other columns were removed due to having no relevant information and making the table needlessly long.
 * Some other rows were amended based on my limited experience.

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -308,7 +308,7 @@ greet earth mars jupiter venus
 ```
 
 ::: tip
-Each line of a command has its resulting value printed out when run, as long as it isn't `null`. Hence, `"hello all:"` above will be printed out despite not being the return value.
+Each line of a command has its resulting value printed out when run, as long as it isn't `null`. Hence, `"hello all:"` above will be printed out despite not being the return value. To prevent this, you can place `null` (or the `ignore` command) at the end of the pipeline, like so: `"hello all:" | null`. Also note that most file system commands, such as `save` or `cd`, always output `null`.
 
 We could call the above definition of the `greet` command with any number of arguments, including none at all. All of the arguments are collected into `$name` as a list.
 
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -8,9 +8,12 @@ An example definition of a custom command:
 
 ```nushell
 def greet [name] {
-  echo "hello" $name
+  ['hello' $name]
 }
 ```
+
+::: tip
+The value produced by the last line of a command becomes the command's returned value. In this case, a list containing the string `'hello'` and `$name` is returned.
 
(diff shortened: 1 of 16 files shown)


book/types_of_data.md
```diff
@@ -42,7 +42,7 @@ A string of characters that represents text. There are a few ways these can be c
   - `$"6 x 7 = (6 * 7)"`
   - `ls | each { |it| $"($it.name) is ($it.size)" }`
 - Bare strings
-  - `echo hello`
+  - `hello`
```
sholderbach, line 45: I think this would be evaluated as a command first. Bare strings are relevant in the arguments or as part of a data structure.

book/working_with_strings.md
```diff
@@ -46,15 +46,15 @@ Nushell currently supports the following escape characters:
 
 Like other shell languages (but unlike most other programming languages) strings can also be written without any quotes:
 ```sh
-> echo hello
+> hello
```
sholderbach, line 49: Same caveat as above

book/custom_commands.md
```diff
@@ -295,25 +298,28 @@ There may be cases when you want to define a command which takes any number of p
...
+    $n
   }
 }
 
 greet earth mars jupiter venus
 ```
 
+::: tip
+Each line of a command has its resulting value printed out when run, as long as it isn't `null`. Hence, `"hello all:"` above will be printed out despite not being the return value.
+
```
sholderbach, line 312: As a reader I would be interested how I can suppress that.

sholderbach:
Thanks for getting this more up to date!

book/types_of_data.md
webbedspace: Thanks!

book/types_of_data.md
```diff
@@ -42,7 +42,7 @@ A string of characters that represents text. There are a few ways these can be c
   - `$"6 x 7 = (6 * 7)"`
   - `ls | each { |it| $"($it.name) is ($it.size)" }`
 - Bare strings
-  - `hello`
+  - `"hello"`
```
sholderbach, line 45: I think under the title bare strings we should give an example for something like before `print hello` or `git branch hello`

book/working_with_strings.md
```diff
@@ -44,20 +44,47 @@ Nushell currently supports the following escape characters:
 
 ## Bare strings
 
-Like other shell languages (but unlike most other programming languages) strings can also be written without any quotes:
+Like other shell languages (but unlike most other programming languages) strings consisting of a single 'word' can also be written without any quotes:
```
sholderbach, line 47: Awesome that section reads much more informative now!

sholderbach (approved):

book/types_of_data.md
webbedspace: Ah, quite right.

rgwood:
Good change, thanks.

PR #659, opened 2022-11-05
docs: document changes from #7007

dmatos2012:
Document change from PR #7007 with column name rename.


sholderbach:
Awesome, thanks!

PR #660, opened 2022-11-05
Updated working_with_strings.md (closes #640)

### LINKED ISSUE #640: Documentation: backtick strings should be mentioned in the string section
### Describe the bug

I was going to suggest adding backtick strings (from JS), but apparently they already exist, and moreover seem to function the same as single-quotes except with the advantage that you can include single-quote characters inside them (which I believe is impossible for single-quote strings without interpolation). So, they deserve a mention.

Also, this is just an opinion of mine, but I think they should be considered the "idiomatic" non-escaped string over the single-quote kind, which means they are used in examples from now on in place of single-quotes. Single-quote strings can remain as a slightly-less-convenient alternative for people who expect that syntax to "just work".

### How to reproduce

n/a

### Expected behavior

n/a

### Screenshots

_No response_

### Configuration

n/a

### Additional context

_No response_

webbedspace:
Closes #640.

Also:

 * Added section for string operators
 * Added 'at a glance' table
 * Changed some examples because I feel like self-aggrandising text isn't, on the whole, actually that appealing to the newest of users

### DIFF
--- a/book/working_with_strings.md
+++ b/book/working_with_strings.md
@@ -274,15 +274,33 @@ With the [`parse`](commands/parse.md) command you can parse a string into column
 ╰───┴──────────┴────────────╯
 ```
 
-If a string is known to contain comma-separated or space-separated data, you can use `from ssv` or `from csv`:
+If a string is known to contain comma-separated, tab-separated or multi-space-separated data, you can use `from csv`, `from tsv` or `from ssv`:
 
 ```sh
-
+> "acronym,long\nAPL,A Programming Language" | from csv
+╭───┬─────────┬────────────────────────╮
+│ # │ acronym │          long          │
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for trying to make things clearer. Could you please change the term "Quirks" to "Notes"? Quirks has a negative connotation in my mind.

I've started to document strings [here](https://hackmd.io/yBEUx4czTMuHcTICt-xZAA). Feel free to add and tweak any of this content to your PR. It describes some things a little different than the way.

This [one](https://hackmd.io/@fdncred/BJqi8MrJj) isn't really meant for documentation but was meant to document current behavior with quoting. It's also incomplete but I thought it may be useful to you.

fdncred:
yup

PR #663, opened 2022-11-06
Fix let-env Path example to be accurate for Windows

webbedspace:
Closes https://github.com/nushell/nushell/issues/6733

I understand I could have reworded this to make it clear it works for POSIX only, but I decided to go with this instead because Windows capitalisation is a bit counterintuitive and could do with clarification more.

### DIFF
--- a/book/environment.md
+++ b/book/environment.md
@@ -38,7 +38,7 @@ Using the `let-env` command is the most straightforward method
 So, if you want to extend the Windows `Path` variable, for example, you could do that as follows.
 
 ```
-let-env Path = ($env.Path | prepend '\path\you\want\to\add')
+let-env Path = ($env.Path | prepend 'C:\path\you\want\to\add')
 ```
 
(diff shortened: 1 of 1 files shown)


book/environment.md
```diff
@@ -33,15 +33,15 @@ Using the `let-env` command is the most straightforward method
...
 
-'let-env' is similar to the **export** command in bash.
+'let-env' is similar to the **export** command in Bash.
 
-So if you want to extend the `PATH` variable for example, you could do that as follows.
+So, if you want to extend the Windows `Path` variable, for example, you could do that as follows.
 
 ```
-let-env PATH = ($env.PATH | prepend '/path/you/want/to/add')
+let-env Path = ($env.Path | prepend '\path\you\want\to\add')
```
fdncred, line 41: and you might want to have a drive letter in the path too like `c:\path\you\want\to\add`

fdncred:
thanks

PR #667, opened 2022-11-10
fix deprecated cmd

hbt:
`help commands | where is_plugin == true` doesn't work and returns 

```
Error: nu::shell::column_not_found (link)

  × Cannot find column
   ╭─[source:1:1]
 1 │ nu
   · ▲
   · ╰── value originates here
   ╰────
   ╭─[entry #4:1:1]
 1 │ help commands   | where is_plugin == true
   ·                         ────┬────
   ·                             ╰── cannot find column
   ╰────

```

Fixed to use `$nu.scope.commands` instead


rgwood:
Hi, I'm guessing you're running the latest development version of Nu on the `main` branch? This sounds related to this recent PR: https://github.com/nushell/nushell/pull/7052

It would probably be better to use the new `command_type` column.

fdncred:
ugh, that stinks. the problem is that dan's change didn't take into account plugins, it appears. so there's no way to look for plugins.

rgwood:
@fdncred Does `command_type == "plugin"` not work? Dan's PR has a match branch for plugins: https://github.com/nushell/nushell/pull/7052/files#diff-112ea240c2bb6485a9284b1e337273034f447b47de337dc34e8064cef2320b3bR94

fdncred:
@rgwood it doesn't look like it.
image

The `other` command_types here are plugins

rgwood:
OK. Probably worth an issue.

fdncred:
fixed here
https://github.com/nushell/nushell/pull/7088
Docs could be updated as per the OP
`help commands | where command_type == plugin`

fdncred:
Thanks

PR #671, opened 2022-11-15
add command signature section

WindSoilder:
Reflect to nushell pr: https://github.com/nushell/nushell/pull/7079

### DIFF
--- a/book/command_signature.md
+++ b/book/command_signature.md
@@ -7,7 +7,9 @@ Signatures(Cell paths are supported):
   <string> | str distance <string> -> <int>
 ```
 
-The first one before `|` is the type of input pipeline, then required argument type for the comamnd, finally the output type is `int`.  `(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply operation at the given cell paths, and replace with the result, like:  `ls | str distance 'nushell' 'name'`
+The first type name before `|` describes the type of input pipeline. The command name is followed by the required argument type(s) for the command. The output type is `int` and given after `->`.
+  
+`(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply an operation at the given cell path(s) in a nested structure or table, and replace the column or field with the result, like:  `ls | str distance 'nushell' 'name'`
 
 Here is another one example, `str join`:
 
--- a/book/command_signature.md
+++ b/book/command_signature.md
@@ -20,4 +20,5 @@ Signatures:
 
 It says that `str join` command expect input pipeline is a list of string, and take optional `string` type argument, finally the output type is `string`.
 
-Some commands don't required input pipepline data, the first argument type will be `<nothing>`, the same is happened to output type if one command returns empty.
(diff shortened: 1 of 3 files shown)


book/command_signature.md
```diff
@@ -4,18 +4,18 @@ nu commands contains a signature section, take `str distance` as example, the si
 
 ```
 Signatures(Cell paths are supported):
-  string, string => int
+  <string> | str distance <string> -> <int>
 ```
 
-The first one is always the type of input pipeline, then required argument type, finally the output type is `int`.  `(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply operation at the given cell paths, and replace with the result, like:  `ls | str distance 'nushell' 'name'`
+The first one before `|` is the type of input pipeline, then required argument type for the comamnd, finally the output type is `int`.  `(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply operation at the given cell paths, and replace with the result, like:  `ls | str distance 'nushell' 'name'`
```
sholderbach, line 10:
```suggestion
The first type name before `|` describes the type of input pipeline. The command name is followed by the required argument type(s) for the command. The output type is `int` and given after `->`.
  
`(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply an operation at the given cell path(s) in a nested structure or table, and replace the column or field with the result, like:  `ls | str distance 'nushell' 'name'`
```

book/command_signature.md
```diff
@@ -4,18 +4,18 @@ nu commands contains a signature section, take `str distance` as example, the si
...
 ```
 Signatures:
-  list<string>, string? => string
+  list<string> | str join <string?> -> <string>
 ```
 
 It says that `str join` command expect input pipeline is a list of string, and take optional `string` type argument, finally the output type is `string`.
 
-Some commands don't required input pipepline data, the first argument type will be `nothing`, the same is happened to output type if one command returns empty.
+Some commands don't required input pipepline data, the first argument type will be `<nothing>`, the same is happened to output type if one command returns empty.
```
sholderbach, line 21:
```suggestion
Some commands don't accept or require data through the input pipeline, thus the input type will be `<nothing>`.
The same is true for the output type if the command returns `null` (e.g. `rm`).
```

sholderbach (commented):
Awesome, thanks!

Some suggestions for wording below:

book/programming_in_nu.md
```diff
@@ -24,3 +24,5 @@ Modules let you define a public interface vs. private commands and you can impor
 [Overlays](overlays.md) build on top of modules.
 By defining an overlay, you bring in module's definitions into its own swappable "layer" that gets applied on top of other overlays.
 This enables features like activating virtual environments or overriding sets of default commands with custom variants.
+
+Some built-in command's help message shows [signature](command_signature.md).  You can take a look to get general rules about command signature.
```
sholderbach, line 28:
```suggestion
The help message of some built-in commands shows a [signature](command_signature.md).  You can take a look at it to get general rules how the command can be used.
```

PR #674, opened 2022-11-19
Fix variable path expression example

remlse:
Hello! I'm learning nushell with the book and I'm blown away so far.

I ran this example code and it seems to me `$my_value.name` returns that column as a list. `.0` is required to get the first element and the desired output.


fdncred:
thanks

PR #675, opened 2022-11-20
Add config.nu structure changes

webbedspace:
Note that the advice for "output your existing options" is liable to change based on whether Nushell issue nushell/nushell#7059 is implemented or not. That section is preliminary.

### DIFF
--- a/blog/2022-11-29-nushell-0.72.md
+++ b/blog/2022-11-29-nushell-0.72.md
@@ -99,7 +99,7 @@ The new grouped options look like this:
   }
 ```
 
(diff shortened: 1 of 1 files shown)


sholderbach:
Awesome! Thanks for the conversion snippet!

webbedspace:
> **EDIT:** Ah I missed that this might not work so well without the `$config` variable being properly bidirectional.

Actually, it's currently written such that it works with the current `$env.config` behaviour - since nushell/nushell#7059 seems like it won't be in this version, it's OK to include it as-is.

sholderbach:
Great!

PR #680, opened 2022-11-24
expand docs on using use with export-env

dmatos2012:
Hi, 
This PR tries to add an example usage to the `use` command when we try to export a module's environment. This comes because of discussion in nushell [issue](https://github.com/nushell/nushell/issues/6973). 

I have tried to play along the existing example to reflect what @kubouch taught me, so feel free to modify/add stuff. Thanks:)

### DIFF
--- a/book/modules.md
+++ b/book/modules.md
@@ -190,6 +190,49 @@ export-env {
 Only `$env.TMP_LEN` and `$env.OTHER_ENV` are preserved after evaluating the `export-env` module.
 :::
 
+
+If you also want to keep your variables in separate modules and export its environment, you could try to `export use` it:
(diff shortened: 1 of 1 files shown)


kubouch:
+1 for a Monty Python reference ;-).

I would formulate it with a counter-example: You might want to use `export use` in the module and expect its environment to propagate automatically, but that's not how it works. You need to define the `export-env` manually.

dmatos2012:
Agreed. Changed now to reflect that!

kubouch:
OK, looks good, thanks!

rgwood:
Nice, thanks for improving the docs!

PR #691, opened 2022-12-02
typo fix

raccmonteiro:
I was going to write about the new command `uniq-by` and saw @sholderbach saved me the trouble, thanks :) 
just noticed a typo


sholderbach:
Good catch! In L65 I missed an article for `a table` as well :D

PR #695, opened 2022-12-10
Fix dead links

thara:
Found a dead link in the book.
It seems that it had not been able to follow up changes in https://github.com/nushell/nushell.github.io/pull/634 .


fdncred:
Thanks

PR #696, opened 2022-12-11
add breaking change for split row

WindSoilder:


fdncred:
oops, i should've landed yours first. now there's conflicts. sorry.

WindSoilder:
Ah, that's ok, I resolve it

PR #704, opened 2022-12-15
Add a description file for `explore` command

zhiburt:
https://github.com/nushell/nushell/issues/7332


fdncred:
@zhiburt Thanks for this. It's a good start. However, we cannot put this in `commands/explore.md` because it will get overwritten by the docs auto generation script that we run regularly. Can you please make this a chapter in the book, maybe under the `(Not so) Advanced` section?

We can start with the documentation that you have here and add screenshots and such of each screen and command later. Thanks again!

zhiburt:
.

fdncred:
thanks!

PR #711, opened 2022-12-21
clarify columns and rows

### LINKED ISSUE #556: Defining rows and columns
### Summary:
While it is not always presented visually as horizonal rows and vertical columns, in nu, table columns have named headers, and table rows have numbered headers.
It might be good to add a definition of rows and columns like this to the book, and possibly a reminder in the section mentioned below, for anyone else who may be confused by the use of "column" in that specific example.

### Motivation:

Reading through the book, I was confused at the page on working with tables, as the following example seemed to use "column" to refer to a row of the table:
```
Adding a new column

We can use the insert command to add a new column to the table. Let's look at an example:

> open rustfmt.toml
─────────┬──────
 edition │ 2018
─────────┴──────

Let's add a column called "next_edition" with the value 2021:

> open rustfmt.toml | insert next_edition 2021
──────────────┬──────
 edition      │ 2018
 next_edition │ 2021
──────────────┴──────
```
Initially, I figured this was a mistake in the documentation, however I realised that while it is inconsistent with the way the data is visually presented as a table in this case, it is probably consistent with how nu treats tables:
Columns have named headers, rows have numbered headers, and as this is a single record, so the columns are able to be stacked vertically.

### Further thoughts
Based on my current understanding of nu tables, a table can be alternately thought of as either:
- a list with all elements being records, each record having the same field names in the same order
- a record with all fields being lists, each list having the same length

In this way, a column corresponds to a field when you interpret the table as a record of lists, and a row corresponds to an element (and corresponding list index) when you interpret the table as a list of records.

merelymyself:
Fixes #556.

### DIFF
--- a/book/working_with_tables.md
+++ b/book/working_with_tables.md
@@ -244,9 +244,8 @@ Let's add a column called "next_edition" with the value 2021:
 ──────────────┴──────
 ```
 
-This visual may be slightly confusing, because it looks like what we've just done is add a row. In this case, remember: rows have numbers, columns have names. As a table is just a list of records under the hood, in a sense we have added a row (to the list) and a column (to the table).
+This visual may be slightly confusing, because it looks like what we've just done is add a row. In this case, remember: rows have numbers, columns have names. If it still is confusing, note that appending one more row will make the table render as expected:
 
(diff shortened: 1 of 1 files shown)


fdncred:
thanks!

PR #713, opened 2022-12-22
Add section "Output result to external commands"

Tengs-Penkwe:


book/pipelines.md
```diff
@@ -83,3 +83,53 @@ And the pipeline:
 ```
 
 Are one and the same.
+
+And if you want to see the original form, you can use [`to text`](commands/to_text.md):
```
rgwood, line 87: Hmm. `to text` converts structured data into text in an opinionated way; it's not exactly "the original form"

Tengs-Penkwe: Thanks for correction

book/pipelines.md
```diff
@@ -83,3 +83,47 @@ And the pipeline:
 ```
 
 Are one and the same.
+
+## Output result to external commands 
+
+Sometimes you want to output your data to some external commands and continue processing them, but the data inside nushell are structured, which means if you output them directly, you will get result you don't want.
```
rgwood, line 89:
```suggestion
Sometimes you want to output Nushell structured data to an external command for further processing. However, Nushell's default formatting options for structured data may not be what you want.
```

book/pipelines.md
```diff
@@ -83,3 +83,47 @@ And the pipeline:
...
+```
+
+You decided to use `grep` and [pipe](https://www.nushell.sh/book/pipelines.html) the result to external `^ls`
+
+```
+> ls /usr/share/nvim/runtime/ | get name | ^grep tutor | ^ls -la $in 
+ls: cannot access ''$'\342\224\202'' 32 '$'\342\224\202'' /usr/share/nvim/runtime/tutor        '$'\342\224\202\n': No such file or directory
+```
+
+What's wrong? It turns out nu shell pipe the whole record(including all tabulators) to `^ls`, and `^ls` accept them all! so you need to convert the record to text first using [`to text`](commands/to_text.md) we mentioned above:
```
rgwood, line 116:
```suggestion
What's wrong? Nushell renders lists and tables (by adding a border with characters like `╭`,`─`,`┬`,`╮`) before piping them as text to external commands. If that's not the behavior you want, you must explicitly convert the data to a string before piping it to an external. For example, you can do so with [`to text`](commands/to_text.md):
```

rgwood (changes requested):
Thanks for the PR! Added some suggestions.

In the future, I think that piping a `list<string>` to an external command should just work - we shouldn't need `to text`. I'll raise an issue for that.

rgwood (approved):

PR #719, opened 2022-12-30
docs: 📝 document the catch block parameter

melMass:
As discussed on discord, this PR is about documenting the catch block optional parameter (containing the actual Error).
Not sure how to make it clear and not too lengthy...

### Proposed option

**Try to run a missing command, and parse the error**
```shell
> try { asdfasdf } catch {|err| if ("executable was not found" in $"($err)") { print "Missing Command" } else { print "Unexpected Error" }}
```

### Other considered options

1. **Try to run a missing command, do something and return the error.**: 
	```nu
	try { asdfasdf } catch {|err| echo 'missing' | $err }
	```
2. **Try to run a missing command, return the raw error as string**:
	```nu
	try { asdfasdf } catch {|err| print $"Got Error: ($err)"}
	```
	*Suggested by @kubouch , might be the most straightforward one actually*

### DIFF
--- a/book/commands/try.md
+++ b/book/commands/try.md
@@ -38,3 +38,8 @@ Try to run a missing command
 ```shell
 > try { asdfasdf } catch { echo 'missing' }
 ```
+
+Try to run a missing command, and parse the error
+```shell
+> try { asdfasdf } catch {|err| if ("executable was not found" in $"($err)") { print "Missing Command" } else { print "Unexpected Error" }}
+```


sholderbach:
Can you add the example to the `try` command directly so it is also accessible from inisde the shell as well?

kubouch:
Yeah, not here because these command .md files are autogenerated. https://github.com/nushell/nushell/blob/main/crates/nu-command/src/core_commands/try_.rs would be a better place to put the example.

I thought you meant to add it to the book but now I see it does not seem to have try/catch at all. We could add a chapter on error handling.

melMass:
Oops yes! Not the first time I make that mistake 😇

melMass:
Migrated here https://github.com/nushell/nushell/pull/7641

PR #724, opened 2023-01-01
Update plugins.md

TornaxO7:
If I'm understanding that correctly from [this PR](https://github.com/nushell/nushell/pull/7052) `command_type` should be `plugin` if it comes from a plugin.


rgwood:
Correct. Thanks for the fix!

PR #725, opened 2023-01-01
Update custom_completions.md

TornaxO7:
it's in my opinion more intuitive to have *not* the partial expand per default

### DIFF
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -90,6 +90,7 @@ let carapace_completer = {|spans|
 let-env config = {
     # ... your config
     completions: {
+        partial: true  # set this to false to prevent partial filling of the prompt
         external: {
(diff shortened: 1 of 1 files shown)


TornaxO7:
wait, it looks like that it's not disabling the auto-expansion for paths on my machine.

TornaxO7:
please take a look into the discord server. I posted a video in the help channel which shows what I mean.

fdncred:
If we want to make this change, it needs to start in the nushell repo with the default_config.nu and the rust code. We try to make all the rust defaults match the default_config.nu so that, even without a config, you still get the defaults that are in default_config.nu.

TornaxO7:
so should I close this PR then?

TornaxO7:
okeydokey

PR #726, opened 2023-01-02
Mention null/nothing in default entry

daniel-vainsencher:
Help people find solutions to nothing/nulls

### DIFF
--- a/book/commands/default.md
+++ b/book/commands/default.md
@@ -4,9 +4,9 @@ categories: |
   filters
 version: 0.73.1
 filters: |
-  Sets a default row's column if missing.
+  Sets a default row's column if missing/null/nothing.
 usage: |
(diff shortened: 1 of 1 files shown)


rgwood:
Hi, thanks for the PR. These docs are auto-generated from Nushell command signatures; are you able to make the change in the main Nushell repo? This is where you want to make the fix: https://github.com/nushell/nushell/blob/d1a78a58cd75000820a8380417840cfdd21bff55/crates/nu-command/src/filters/default.rs#L31

I think the current `default` description is a little confusing (not your fault); maybe it should be "Replace null/missing values with a new value"?

daniel-vainsencher:
Of course, thanks for the pointer. It had seemed a bit weird :)

PR #732, opened 2023-01-05
Try to rename from blog text to changeLog

WindSoilder:
Sorry I'm not really good at web page, so in this pr I just try to change text `Blog` to `Changelog` and see if it's working

### DIFF
--- a/.vuepress/configs/navbar/en.ts
+++ b/.vuepress/configs/navbar/en.ts
@@ -4,5 +4,5 @@ export const navbarEn: NavbarConfig = [
   { text: 'Book', link: '/book/' },
   // { text: "Contributor Book", link: "/contributor-book/" },
   { text: 'Cookbook', link: '/cookbook/' },
-  { text: 'Blog', link: '/blog/' },
+  { text: 'Changelog', link: '/blog/' },
 ];


sholderbach:
Yeah let's give it a try!

amtoine:
@WindSoilder 
i've been thinking about this and i think i find *changelog* more confusing than *blog* :thinking: 

in the current [*changelog* section of the book](https://www.nushell.sh/blog/) we have not only posts about new releases, but also regular blog posts like [*Nushell + Uutils = ❤️*](https://www.nushell.sh/blog/2023-09-05-why-uu.html) or [*Nushell turns 4 years old!*](https://www.nushell.sh/blog/2023-08-23-happy-birthday-nushell-4.html), which always sounds odd to me :confused:

WindSoilder:
The issue comes from we don't separate `ChangeLog` from `Blog`.
Zed editor provides a good show case: https://zed.dev/
They make both `ChangeLog` and `Blog` discorvable.

But currently I'm ok with revert the pr

fdncred:
We could have blog and changelog, or maybe just name it Blog/Changelog?

PR #748, opened 2023-01-21
Make reference link clickable.

chaoticenginerd:
Update the docs for the [`if` command](https://www.nushell.sh/book/commands/if.html)  to make the detail reference link clickable.


sholderbach:
Thanks for spotting this! The command docs are autogenerated from the help pages of the commands so your change would be overwritten when we update the docs with the next release. We might need to look into emitting hyperlinks properly in the `book/make_docs.nu` script.

PR #753, opened 2023-01-28
Add Github Actions workflow to check typos

Hofer-Julian:
- Add Github Actions workflow to check typos
- Fix existing typos

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -1,0 +1,20 @@
+name: Typos
+on: [pull_request]
+
+jobs:
+  run:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions Repository
+        uses: actions/checkout@v2
+
+      - name: Check spelling of book
+        uses: crate-ci/typos@master
+        with:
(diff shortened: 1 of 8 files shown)


fdncred:
I'm kind of wondering how we can use this effectively so far down stream. For anything that is wrong in the commands, we can't fix it in the docs here. We have to fix it in the rust code because all the command docs are auto generated with a nushell script.

The other changes to pages in the book, and things like that, we can make those changes here just fine.

Hofer-Julian:
> I'm kind of wondering how we can use this effectively so far down stream. For anything that is wrong in the commands, we can't fix it in the docs here. We have to fix it in the rust code because all the command docs are auto generated with a nushell script.

My impression is that `typos` targets code more than it does docs.
But because of that, we could also use it additionally for the nushell code base itself.

Hofer-Julian:
> My impression is that typos targets code more than it does docs.
But because of that, we could also use it additionally for the nushell code base itself.

If you want, I can create a follow up PR there.

fdncred:
I think this is fine for now but the split-by changes will get overwritten if the rust code in nushell isn't changed.

PR #754, opened 2023-01-28
Add separate page for setting Nu as default shell

Hofer-Julian:
- Adds instructions to add Nu to your terminal
- Advises against setting Nu as login shell

Heavily inspired by
https://elv.sh/get/#using-elvish-as-your-default-shell

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -13,6 +13,7 @@ export const sidebarEn: SidebarConfig = {
       collapsible: false,
       children: [
         '/book/installation.md',
+        '/book/default_shell.md',
         '/book/quick_tour.md',
         '/book/moving_around.md',
         '/book/thinking_in_nu.md',
--- a/book/default_shell.md
+++ b/book/default_shell.md
@@ -14,9 +14,9 @@
 ## Setting Nu as login shell (Linux, BSD & macOS)
 
 ::: warning
-We do not recommend setting your login shell to Nu.
-Some programs on your system might assume that your login shell is [POSIX](https://en.wikipedia.org/wiki/POSIX) compatible.
-Breaking that assumption often leads to unexpected issues.
+Nu is still in development and is not intended to be POSIX compliant.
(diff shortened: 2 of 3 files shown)


fdncred:
I like the changes except that the part that says "We do not recommend setting your login shell to Nu.". I use it as my login shell in WSL2 every weekday without issue. I can understand that we may need to say something but I don't want to deter people from using nushell, just the opposite.

I think it's ok to have a warning but I think we should use different words. Perhaps like "Warning, Nu is still in development and is not intended to be POSIX compliant. If you need strict POSIX compliance at login time, you should consider using a alternate login shell."

Maybe not those exact words, but something less like "Don't use nushell" 😆

Hofer-Julian:
> Maybe not those exact words, but something less like "Don't use nushell" 

I understand what you are saying, and I also want people to use nushell.
Even as their default shell, which is why I've added the terminal instructions.

> If you need strict POSIX compliance at login time

Unfortunately, that's not something you can choose.
There are just too many programs who just assume that your login shell is POSIX compliant, and the error messages tend to be very bad and/or hidden somewhere in the logs.

Some friends noticed that also with some vscode extensions, so that problem probably also occurs on WSL and macOS.

If you still want me to weaken the message or add alternative "default methods" (see https://wiki.archlinux.org/title/Fish), I am happy to.

fdncred:
Yes, it needs to be similar to what I suggested please. I don't want to discourage use.

Hofer-Julian:
Updated it to:
> Nu is still in development and is not intended to be POSIX compliant.
Be aware that some programs on your system might assume that your login shell is [POSIX](https://en.wikipedia.org/wiki/POSIX) compatible.
Breaking that assumption can lead to unexpected issues.

How do you like that one?

fdncred:
Much better. Thank you.

fdncred:
One last thing, Do you think the `installation` page should have a link to the `default_shell` page? I could go either way.

Hofer-Julian:
vuepress does that for us at the very end of the page.
I feel like that is sufficient :)

grafik

rgwood:
Thanks for improving the docs!

PR #757, opened 2023-01-29
Reflect the separation of block and closure

Hofer-Julian:
This change was introduced on 0.72

### DIFF
--- a/book/types_of_data.md
+++ b/book/types_of_data.md
@@ -401,7 +401,7 @@ There are numerous other commands for selecting and reducing the data in tables,
 Closures are anonymous functions that can close over variables outside their scope and can be passed as a value.
 For example, in the command `each { |it| print $it }` the closure is the portion contained in curly braces, `{ |it| print $it }`.
 Closure parameters are specified between a pair of pipe symbols (for example, `|it|`) if necessary.
-You can also use `$in` in most closures instead of providing a parameter: `each { print $in }`
+You can also use a pipeline input as `$in` in most closures instead of providing an explicit parameter: `each { print $in }`
 
 Closures are a useful way to represent code that can be executed on each row of data.
 It is idiomatic to use `$it` as a parameter name in [`each`](commands/each.md) blocks, but not required;
--- a/book/types_of_data.md
+++ b/book/types_of_data.md
@@ -399,10 +399,19 @@ There are numerous other commands for selecting and reducing the data in tables,
 ## Closures
 
 Closures are anonymous functions that can be passed a value through parameters and _close over_ (i.e. use) a variable outside their scope.
-Closures itself can be bound to a named and passed as a parameter.
+
 For example, in the command `each { |it| print $it }` the closure is the portion contained in curly braces, `{ |it| print $it }`.
 Closure parameters are specified between a pair of pipe symbols (for example, `|it|`) if necessary.
 You can also use a pipeline input as `$in` in most closures instead of providing an explicit parameter: `each { print $in }`
+
+Closures itself can be bound to a named variable and passed as a parameter.
+To call a closure directly in your code use the [`do`](commands/do.md) command.
+
+```nu
+# Assign a closure to a variable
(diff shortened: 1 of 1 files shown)


book/types_of_data.md
```diff
@@ -395,11 +396,30 @@ To remove rows from a table, you'll commonly use the `select` command with row n
 
 There are numerous other commands for selecting and reducing the data in tables, records and lists.
 
+## Closures
+
+Closures are anonymous functions that can close over variables outside their scope and can be passed as a value.
```
sholderbach, line 401: Maybe we could lighten the CS lingo for the first sentence

```suggestion
Closures are anonymous functions that can be passed a value through paramters and *close over* (i.e. use) a variable outside its scope.
```

book/types_of_data.md
```diff
@@ -395,11 +396,30 @@ To remove rows from a table, you'll commonly use the `select` command with row n
...
+For example, in the command `each { |it| print $it }` the closure is the portion contained in curly braces, `{ |it| print $it }`.
+Closure parameters are specified between a pair of pipe symbols (for example, `|it|`) if necessary.
+You can also use `$in` in most closures instead of providing a parameter: `each { print $in }`
```
sholderbach, line 404: Mhh `$in` semantics are always a bit confusing. Would be great if we could specify the rules when it is available in the future. 

I think in this case we should make clear that this is not traditional closing over the variable.

```suggestion
You can also use a pipeline input as `$in` in most closures instead of providing an explicit parameter: `each { print $in }`
```

book/types_of_data.md
```diff
@@ -395,11 +396,30 @@ To remove rows from a table, you'll commonly use the `select` command with row n
...
+
+Closures are a useful way to represent code that can be executed on each row of data.
+It is idiomatic to use `$it` as a parameter name in [`each`](commands/each.md) blocks, but not required;
```
sholderbach, line 407: Love the crossreferencing!

sholderbach (commented):
Thanks for getting the docs up to speed and clarifying this slightly confusing bit of nushell semantics

book/types_of_data.md
Hofer-Julian: There is a small typo: paramters -> parameters :)

book/types_of_data.md
Hofer-Julian: Agreed.

book/types_of_data.md
Hofer-Julian: Unfortunately, not my work.
I just copied the block paragraph and replaced block with closure :P

book/types_of_data.md
sholderbach: Also made a grammar mistake with `its` instead of `theirs`. Peak Verschlimmbessern

Hofer-Julian: I have to say, reading through the sentence, I still like the original a bit more.
That functions can "be passed a value through parameters" is not astounding, but that they "can be passed as a value" is worth mentioning.

Hofer-Julian: Done

book/types_of_data.md
sholderbach: Thanks for keeping it around then :)

book/types_of_data.md
```diff
@@ -399,11 +399,20 @@ There are numerous other commands for selecting and reducing the data in tables,
...
 Closure parameters are specified between a pair of pipe symbols (for example, `|it|`) if necessary.
 You can also use a pipeline input as `$in` in most closures instead of providing an explicit parameter: `each { print $in }`
 
+Closures itself can be bound to a named variable and passed as a parameter.
+To call a closure directly in your code use the [`do`](commands/do.md) command.
+
+```nu
+# Assign a closure to a variable
+let greet = { |name| print $"Hello ($name)"}
+do $greet "Julian"
```
Hofer-Julian, line 413: 😄

PR #761, opened 2023-02-01
Remove note about environment scoping in def_env page

Canop:
The current page contains an obsolete note warning users about blocks creating scopes constraining environment changes.

As block scoping rules changed, this note should be removed.

### DIFF
--- a/book/commands/def-env.md
+++ b/book/commands/def-env.md
@@ -23,37 +23,7 @@ usage: |
  -  `params`: parameters
  -  `block`: body of the definition
 
-## Notes
-```text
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for this, but in order for this to work, it needs to be changed in the nushell code base because anything under commands gets systematically generated when we make a new release. This is where it would need to be changed. We can probably land this after this code is changed. https://github.com/nushell/nushell/blob/5e957ecda69f9221875e304de3a566e74193c22d/crates/nu-command/src/core_commands/def_env.rs#L26-L55

PR #764, opened 2023-02-04
fix statement about `lines` output type ("table"→"list")

das-g:
Fix statement about `lines` output type, which is a `list`, not a `table`.

This fixes it in the original English and in the German translation. Other translations contain the same error, but will have to be fixed in separate changes, as I don't know them well enough.

PR #765, opened 2023-02-04
[es] fix statement about `lines` output type ("tabla"→"lista")

das-g:
Fix statement about `lines` output type, which is a `list`, not a `table`.

This should fix it in the Spanish translation. See PR #764 for a fix of the same problem in the English original and in the German translation.

I don't really know Spanish. I simply looked up the term to use in [Tipos de datos→Listas](https://www.nushell.sh/es/book/tipos_de_datos.html#listas) (and in [Tipos de datos→ Datos estructurados](https://www.nushell.sh/es/book/tipos_de_datos.html#datos-estructurados) where it occurs in singular, albeit not referring to the data type), so it's quite possible I'm introducing grammatical errors.

### DIFF
--- a/es/book/cargando_datos.md
+++ b/es/book/cargando_datos.md
@@ -75,7 +75,7 @@ Lo primero que queremos hacer al cargar el archivo es trabajarlo línea por lín
 ---+------------------------------
 ```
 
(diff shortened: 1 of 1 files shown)


hustcer:
Thanks

PR #769, opened 2023-02-05
replaces fetch with http in varius places

FilipAndersson245:
I also created a new image with the updated command to replace the fetch one.


fdncred:
Thanks. I think all the text changes look good. My only worry is the png is too small. Would it be possible to increase your font size to try and match the size of the original image closer?

FilipAndersson245:
> Thanks. I think all the text changes look good. My only worry is the png is too small. Would it be possible to increase your terminal font size and take another screen shot in order to try and match the size of the original image closer?

I updated with a higher res one,

fdncred:
Thanks.

PR #773, opened 2023-02-09
Fix changed history command to match default in config

jafriyie1:
This PR provides a fix from changing `Ctrl+x` to `Ctrl+r` on the [Reedline Page](https://www.nushell.sh/book/line_editor.html#completion-menu). This fix addresses this [issue](https://github.com/nushell/nushell.github.io/issues/737).


fdncred:
Thanks

Issue #774, opened 2023-02-09
Outdated -n flag within each command


jafriyie1:
Hi there! I'm currently going through the book since I'm actively learning about Nushell and I have noticed some issues within the documentation. For example, on this [page](https://www.nushell.sh/book/thinking_in_nu.html#variables-are-immutable) I was trying to run this code block: 
```
 ls | each -n { |it| $"Number ($it.index) is size ($it.item.size)" }
``` 
I was also trying to run this code block on this page:
```
let names = [Mark Tami Amanda Jeremy]
$names | each { |it| $"Hello, ($it)!" }
# Outputs "Hello, Mark!" and three more similar lines.

$names | each -n { |it| $"($it.index + 1) - ($it.item)" }
# Outputs "1 - Mark", "2 - Tami", etc.
```

When I ran these lines of code within Nushell I kept getting the following error: 
```
│ $names | each -n { |it| $"($it.index + 1) - ($it.item)" }
   ·             ┬
   ·             ╰── unknown flag
   ╰────
```
I was also getting errors related to the index and number as well. After running `help each` I was able to rectify the issue by piping `enumerate` before the each command. I think the documentation may need to be updated across the book so that others don't get bogged down by this. I would be happy to submit a PR and fix this across the documentation if it would be useful!

PR #776, opened 2023-02-10
Fix: updating pages that used each -n command

jafriyie1:
This PR addresses this [issue](https://github.com/nushell/nushell.github.io/issues/774). The docs were still using `each -n` so I have updated it to use `enumerate | each` as seen in the newest version of Nushell.


hustcer:
Thanks

PR #777, opened 2023-02-11
release notes for #7952

bobhy:
Doc for nushell/nushell#7952


fdncred:
I think you're html is malformed. I'm thinking there's either an extra `<tr>` or a missing `</tr>`.

fdncred:
thanks

PR #779, opened 2023-02-14
de/README.md: Alles ist im Singular

funkyfuture:
this just fixes a grammatical error in the german translation.

### DIFF
--- a/de/README.md
+++ b/de/README.md
@@ -8,7 +8,7 @@ actionLink: /de/book/
 features:
   - title: Pipelines um jedes Betriebssystem zu steuern
     details: Nu läuft auf Linux, macOS und Windows. Einmal lernen - überall verwenden.
-  - title: Alles sind Daten.
+  - title: Alles ist Daten.
     details: Pipelines in Nu nutzen strukturierte Daten die einfach ausgewählt, gefiltert und sortiert werden können. Nie wieder Strings parsen, sondern direkt anfangen Probleme zu lösen.
   - title: Leistungsfähige Plugins
     details: Nu kann dank des leistungsfähigen Plugin-Systems einfach erweitert werden.


funkyfuture:
the noun "Alles" surprisingy doesn't appear in the dwds, but there's an entry of an idiom where it is contained, all examples show that the numerus is singular: https://www.dwds.de/wb/Ein%20und%20Alles

funkyfuture:
i noticed that too. but i think the proper grammar just emphasizes that the expression is plain marketing gibberish.

and indeed, i luckily didn't have data for dinner.

even or especially in the domain of software it doesn't makes a distinguishable point. for any software everything is data.

hustcer:
@sholderbach Can we merge this PR?

de/README.md
```diff
@@ -8,7 +8,7 @@ actionLink: /de/book/
 features:
   - title: Pipelines um jedes Betriebssystem zu steuern
     details: Nu läuft auf Linux, macOS und Windows. Einmal lernen - überall verwenden.
-  - title: Alles sind Daten.
+  - title: Alles ist Daten.
```
ehasnain, line 11: I think a bit more comprehensible translation would be:

Alles besteht aus Daten.

das-g (approved):
> the noun "Alles" surprisingy doesn't appear in the dwds, but there's an entry of an idiom where it is contained, all examples show that the numerus is singular: https://www.dwds.de/wb/Ein%20und%20Alles

That's because there is (at least to my understanding) no noun "Alles", except as a substantiation of "alles" ([DWDS](https://www.dwds.de/wb/alles), [German Wiktionary](https://de.wiktionary.org/wiki/alles), [English Wiktionary](https://en.wiktionary.org/wiki/alles#German)).

> Mhh that translation sounds just weird.

Indeed, but with the change proposed here, it's at least grammatically correct, so that's an improvement.

> i noticed that too. but i think the proper grammar just emphasizes that the expression is plain marketing gibberish.
>
> […]
>
> even or especially in the domain of software it doesn't makes a distinguishable point. for any software everything is data.

Indeed. As the data traveler (Datenreisende) Wau Holland [famously uttered](https://de.wikiquote.org/w/index.php?title=Wau_Holland&oldid=517113):
> ### Alles ist eins – außer der Null

(They even made a [movie about him](https://allesisteins.film) titled thusly. But I digress.)

While even with this change, the title in question remains far from perfect, the change is an improvement and it's unclear which further improvement beyond it (if any) would be the right one. Thus I recommend this PR as-is, as PR reviews are probably not the right forum to come up with a yet better solution.

de/README.md
Kissaki: I'm a native German speaker. The originally suggested "Alles ist Daten" sounds awful and honestly doesn't make sense.

*(Alles is not singular. It's everything. So *sind* makes more sense.)*

I think the original or what @ehasnain suggested is much better. With the *details* context, what @ehasnain suggested is a good and fitting title.

*Honestly, I think the original EN text could be improved in the same way as here. The statement is too generic and text-shell-context-knowledge dependent. But we don't have to raise that here in this specific PR.*

cptpiepmatz:
I thought about this a bit and I think we may need to find something different and not a 1-to-1 translation, maybe:
- *"Daten statt Text"*
- *"Alles ist strukturiert"*

Kissaki:
Reading the desc, the keywords are structured data and pipelines (which I guess was kept untranslated). So how about "Strukturierte Daten" or "Strukturierte Daten und Pipelines" or "Strukturierte Daten in Pipelines".

In the context of a landing page and feature promo block I prefer "Strukturierte Daten und Pipelines".

Kissaki:
> At the same time we are loosing the pun on "everything is a file" when we try to be more general here.

I suspect that pun becomes less and less understandable by the target audience. Even now, I suspect fewer people will get it than those who do. And even those to could get the pun - personally, I know of the everything is a file concept, but the EN headline felt more clunky than anything.

Kissaki:
Please close this superseded/replaced MR.

The suggested change is replaced by #1838 as per discussion and MR description.

cptpiepmatz:
closed by #1838

PR #780, opened 2023-02-14
Fix typos in plugins.md

msmart:


sholderbach:
Thanks for helping out and making nushell better!

PR #782, opened 2023-02-15
Move command reference to navbar

Hofer-Julian:
- Move command reference to navbar
- Create one sidebar entry for each category
- Simplify javascript code


Hofer-Julian:
The command reference is really long at the moment.
I feel that change makes it easier to consume.

fdncred:
Thanks for this. It seems like a good change. I'm just not sure how to try it out. Is there an easy way to clone the repo and take it for a test drive?

Hofer-Julian:
> Thanks for this. It seems like a good change. I'm just not sure how to try it out. Is there an easy way to clone the repo and take it for a test drive?

You can checkout my branch and run `npm run dev`. That should serve the site locally.

If that doesn't work, then I can upload screenshots.

fdncred:
ok, i just had to do `npm install` then `npm dev run` and it worked great! Thanks for the tip.

1. This makes the command reference much more usable. Great idea!
2. It makes me want to make sure our categories are more accurate and categorize things better.
3. The only thing that's a little weird is when you're drilling down to a command, you click that command, and it takes you to a different place. Now, if you want to go back to looking at the categories again, where you were previously, it's several clicks to get back there. I don't think this complaint should hold up the PR, but it may be worth thinking about in the future.

Hofer-Julian:
>     3. The only thing that's a little weird is when you're drilling down to a command, you click that command, and it takes you to a different place. Now, if you want to go back to looking at the categories again, where you were previously, it's several clicks to get back there. I don't think this complaint should hold up the PR, but it may be worth thinking about in the future.

Agreed, I am happy to move the commands to `command-reference`, adapt `make_docs.nu` and fix the links in a follow-up PR

>     4. It makes me wonder if we can generate this automatically because it's pretty frequent that we add a category or move a command around. There's a bunch of those coming up in the next release and it makes me think this will be outdated in the next week.

That is a good point, and that's also what I've tried first.
The main annoyance is that this way  `make_docs.nu` would have to modify (generate?) `sidebar/en.ts`.
How would this then work with translations?
I am not sure.
Also, `make_docs` only works with re-runs on different platforms, which adds another level of complexity.

I've re-added the full command output to make sure that this PR at least isn't a regression.

hustcer:
> The command reference is really long at the moment. I feel that change makes it easier to consume.

Thanks, it's really an awesome change. I agree with that we should generate the `command-reference`s and nav items automatically:
1. For the `command-reference`s all the templates are quite similar to each other:
```console
$> diff command-reference/bits.md command-reference/bytes.md
1c1
< # Bits
---
> # Bytes
10c10
<           .filter(p => p.frontmatter.categories.includes('bits'))
---
>           .filter(p => p.frontmatter.categories.includes('bytes'))
```
I think at least we can create all the templates automatically. A much better approach might be create a `command reference` layout or template just like `.vuepress/components/BlogPosts.vue` (I'm not quite sure if it's feasible, need more investigation)

2. For the nav items how about create an command nav js file like:
```javascript
export const commandNavs = [
        'bits.md',
        'bytes.md',
        'conversions.md',
        'core.md',
        'database.md',
        'dataframe.md',
        'date.md',
        'default.md',
        'deprecated.md',
        'env.md',
        'experimental.md',
        'expression.md',
        'filesystem.md',
        'filters.md',
        'formats.md',
        'generators.md',
        'hash.md',
        'lazyframe.md',
        'math.md',
        'misc.md',
        'network.md',
        'platform.md',
        'random.md',
        'shells.md',
        'strings.md',
        'system.md',
        'viewers.md',
]
```
then we can `import` and use it in the sidebars

3. BTW: By running `$nu.scope.commands|select category|uniq` I got 30 categories, did we miss any of them? and how about merging category of `lazyframe`, `dataframe` and `dataframe or lazyframe`?

Hofer-Julian:
> I think at least we can create all the templates automatically. A much better approach might be create a `command reference` layout or template just like `.vuepress/components/BlogPosts.vue` (I'm not quite sure if it's feasible, need more investigation)

Creating the templates with a nu script sounds good to me.
For the Vue-specific approaches, I am afraid I don't know enough Vue to tackle that.
 
>     2. For the nav items how about create a command nav js file by a script with the content like:
> 
> 
> ```js
> export const commandNavs = [
>         'bits.md',
>         'bytes.md',
>         'conversions.md',
>         'core.md',
>         'database.md',
>         'dataframe.md',
>         'date.md',
>         'default.md',
>         'deprecated.md',
>         'env.md',
>         'experimental.md',
>         'expression.md',
>         'filesystem.md',
>         'filters.md',
>         'formats.md',
>         'generators.md',
>         'hash.md',
>         'lazyframe.md',
>         'math.md',
>         'misc.md',
>         'network.md',
>         'platform.md',
>         'random.md',
>         'shells.md',
>         'strings.md',
>         'system.md',
>         'viewers.md',
> ]
> ```
> 
> then we can `import` and use it in the sidebars

That is a great idea.
This should also be relatively easy to generate with a nu script

>     3. BTW: By running `$nu.scope.commands|select category|uniq` I got 30 categories, did we miss any of them? 

I've used `help commands | get category | group-by | columns`, and I _think_ I didn't miss any (excluding `dataframe or lazyframe` as discussed below).
Can you name the specific categories you are missing?

> And how about merging category of `lazyframe`, `dataframe` and `dataframe or lazyframe`?

I assume you mean on the Nushell side?
At least with `dataframe or lazyframe`, I also feel that category doesn't make too much sense.
The current implementation just ignores this category and appends its content to dataframe and lazyframe.

hustcer:
> Can you name the specific categories you are missing?

There are three categories: `prompt`(for gstat plugin), `debug`(maybe just added in v0.75.1? we can ignore it for now) and `dataframe or lazyframe`

> > And how about merging category of lazyframe, dataframe and dataframe or lazyframe?

> I assume you mean on the Nushell side?

I mean on the docs side, I'm not sure whether we should do it on Nushell side or not.

> At least with dataframe or lazyframe, I also feel that category doesn't make too much sense.
> The current implementation just ignores this category and appends its content to dataframe and lazyframe.

Agree, that's why I think we should merge them in doc nav

Hofer-Julian:
> There are three categories: `prompt`(for gstat plugin), `debug`(maybe just added in v0.75.1? we can ignore it for now) and `dataframe or lazyframe`

I've installed Nushell with `cargo install nu --features=dataframe`.
Can you tell me what I have to do to get the other plugins?

> I mean on the docs side, I'm not sure whether we should do it on Nushell side or not.

Not sure if generating categories *and* merging some of them is the best approach here.
In the current state, I could only easily merge them, because I've hard-coded the categories.

hustcer:
> I've installed Nushell with cargo install nu --features=dataframe.
Can you tell me what I have to do to get the other plugins?

You can clone the nushell repo and then run `install-all.sh` or `install-all.ps1`, then run `register ~/.cargo/bin/nu_plugin_gstat`

Hofer-Julian:
Regarding this PR, I think what needs to be done is:

1) Add `prompt` category

2) Adapting `make_docs.nu` so that it generates:
    -  `nav.ts` file with categories used by `sidebar/en.ts`
    - category markdown files in `command-reference/categories`
    - command markdown files in  `command-reference/commands`

3) Discussing on the Nushell side whether some categories should be refactored, especially `dataframe or lazyframe`

I have a preference for working on 2) in a follow-up PR, because I think this PR is already an improvement over the status quo. 
However, I would be willing to extend this one too.

hustcer:
> command markdown files in command-reference/commands

I think we should keep the command docs in `/book/commands/`, unless you can search `commands/` in `**/*.md` and modify all the affected links to avoid broken.

Hofer-Julian:
> > command markdown files in command-reference/commands
> 
> I think we should keep the command docs in `/book/commands/`, unless you can search `commands/` in `**/*.md` and modify all the affected links to avoid broken.

That's what I would have done, but I am also fine with leaving them where they are.

hustcer:
Any option is acceptable, maybe move to `command-reference/commands` is a better choice ?

Hofer-Julian:
> And do you think command-reference is a bit longer ? How about add commands/categories and make all commands placed in commands directly like commands/ls.md then https://www.nushell.sh/book/commands/ls.html will become https://www.nushell.sh/commands/ls.html, this will make the command url shorter. 😆

Fine by me.
We can in general rename "Command Reference" to "Commands".
I was considering both and just took one of them at random.
What do you think @fdncred?

fdncred:
@Hofer-Julian I like the suggestion. I also like `Commands` vs `Command Reference`.

Hofer-Julian:
I've now added the `prompt` category and renamed `command-reference` to `commands`


>     2. Adapting `make_docs.nu` so that it generates:
>        
>        * `nav.ts` file with categories used by `sidebar/en.ts`
>        * category markdown files in `command-reference/categories`
>        * command markdown files in  `command-reference/commands`
> 
> I have a preference for working on 2) in a follow-up PR, because I think this PR is already an improvement over the status quo. However, I would be willing to extend this one too.

Any thoughts on this ☝️?

hustcer:
Yeah, I think we can merge this one.

fdncred:
LGTM!

PR #785, opened 2023-02-16
Move `book/commands` to `commands/commands`

Hofer-Julian:
@hustcer unfortunately moving the commands directly to `commands` made too many problems, so `commands/commands` it is :)


hustcer:
Thanks, how about `commands/doc`? 
```console
https://www.nushell.sh/commands/commands/ls.html
# VS
https://www.nushell.sh/commands/doc/ls.html
```
I think the later one is more readable.

Hofer-Julian:
> Thanks, how about `commands/doc`?
> 
> ```
> https://www.nushell.sh/commands/commands/ls.html
> # VS
> https://www.nushell.sh/commands/doc/ls.html
> ```
> 
> I think the later one is more readable.

Maybe plural: docs?

hustcer:
`docs` is fine by me

Hofer-Julian:
> `docs` is fine by me

Done!

.prettierignore
```diff
@@ -6,7 +6,7 @@ operators.md
 dataframes.md
 nushell_map.md
 *_bash*.md
-book/commands/*.md
+commands/commands/*.md
```
hustcer, line 9: commands/docs/*.md

book/escaping.md
```diff
@@ -16,4 +16,4 @@ Escape to external command:
 
 ## Windows note
 
-When running an external command on Windows, nushell [used to](https://www.nushell.sh/blog/2022-08-16-nushell-0_67.html#windows-cmd-exe-changes-rgwood) use [Cmd.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd) to run the command, as a number of common commands on Windows are actually shell builtins and not available as separate executables. [Coming from CMD.EXE](coming_from_cmd.md) contains a list of these commands and how to map them to nushell native concepts.
+When running an external command on Windows, nushell [used to](https://www.nushell.sh/blog/2022-08-16-nushell-0_67.html#windows-cmd-exe-changes-rgwood) use [Cmd.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-/commands/docs/cmd) to run the command, as a number of common commands on Windows are actually shell builtins and not available as separate executables. [Coming from CMD.EXE](coming_from_cmd.md) contains a list of these commands and how to map them to nushell native concepts.
```
hustcer, line 19: This line shouldn't be changed

hustcer (commented):
docs in `/commands/docs` should not be formatted

.prettierignore
Hofer-Julian: Indeed, otherwise prettier formats the auto-generated files.

book/escaping.md
Hofer-Julian: Good catch, fixed!

commands/categories/bits.md
```diff
@@ -6,7 +6,7 @@
     computed: {
       commands() {
         return pages
-          .filter(p => p.path.includes('/book/commands/'))
+          .filter(p => p.path.includes('/commands/commands/'))
```
hustcer, line 9: There are 28 `/commands/commands/` need to be changed to `/commands/docs/`

cookbook/tables.md
```diff
@@ -21,7 +21,7 @@ Examples shown in [`Working with tables`](../book/working_with_tables.md) work f
 ───┴───┴───┴───┴───
 ```
 
-Second row in columns `c` and `d` is empty because our `second` table only contained a single row so nushell has nothing to fill the remaining rows with. But what if we wanted the smaller table to 'wrap around' and keep filling the rows? For that we can use the [`group`](../book/commands/group.md) command to split the larger table into subtables, merge each of them with the smaller table and then combine the merged tables together using [`flatten`](../book/commands/flatten.md) command like this:
+Second row in columns `c` and `d` is empty because our `second` table only contained a single row so nushell has nothing to fill the remaining rows with. But what if we wanted the smaller table to 'wrap around' and keep filling the rows? For that we can use the [`group`](../commands/docs/group.md) command to split the larger table into subtables, merge each of them with the smaller table and then combine the merged tables together using [`flatten`](../commands/docs/flatten.md) command like this:
```
hustcer, line 24: Why not use absolute path instead? `/commands/docs/flatten.md` instead of `../commands/docs/flatten.md`

commands/categories/bits.md
Hofer-Julian: Weird, seems like that change got lost somewhere, fixed now.

cookbook/tables.md
Hofer-Julian: Just used what was there before, fixed now.

hustcer:
Thanks, let's land it

fdncred:
Good work!

PR #789, opened 2023-02-18
Prepare for generation of categories

Hofer-Julian:
- Move command categories into separate file to make it easier to generate them later
- Move `make_docs.nu` to top-level
- Refactor `make_docs.nu`
  -  Mostly extract code blocks into functions
  -  Give variables more descriptive names
  -  It reproduces the generated commands, when used with nu 0.75 (two lines are different, so I assume those fixes came later)


hustcer:
https://github.com/nushell/nushell.github.io/blob/63259ce1d640f1e454595c4c1d70e41c0bcbf366/package.json#L8  BTW: The `update` command here need to be fixed too

Hofer-Julian:
> https://github.com/nushell/nushell.github.io/blob/63259ce1d640f1e454595c4c1d70e41c0bcbf366/package.json#L8
> BTW: The `update` command here need to be fixed too

Indeed, that is fixed now

hustcer:
Thanks

PR #792, opened 2023-02-22
Fix typo

Hofer-Julian:
I've tried to only fix this typo.
Not sure where the other changes come from, but they look reasonable to me.
Is it possible that `npm` injects some hook to call `prettier`?


hustcer:
Thanks, and `npm` already has a hook to call `prettier` now when you commit a change, however it works only when you had run `npm i` locally, some one may have not run `npm install` I guess.

PR #793, opened 2023-02-22
Generate categories

Hofer-Julian:


make_docs.nu
```diff
@@ -136,3 +144,51 @@ $"($example.description)
   )
   $doc
 }
+
+def generate-category-sidebar [unique_categories] {
+  let sidebar_path = (['.', '.vuepress', 'configs', "sidebar", "command_categories.ts"] | path join)
+  let list_content = (
+    $unique_categories |
+    each { |category| $category | str replace '\?' '' | str replace ' ' '_' } |
```
hustcer, line 152: `str replace -a ' ' '_'`

make_docs.nu
```diff
@@ -136,3 +144,51 @@ $"($example.description)
...
+  )
+  $"export const commandCategories = [
+($list_content)
+];
+" | save --raw --force $sidebar_path
+}
+
+
+def generate-category [category] {
+  let safe_name = ($category | str replace '\?' '' | str replace ' ' '_')
```
hustcer, line 164: `str replace -a ' ' '_'`

make_docs.nu
```diff
@@ -136,3 +144,51 @@ $"($example.description)
...
+  let doc_path = (['.', 'commands', 'categories', $'($safe_name).md'] | path join)
+  $"# ($category)
```
hustcer, line 166: `($category | str title-case)` will make docs look better

make_docs.nu
Hofer-Julian: Done, and extracted into `safe-path`

make_docs.nu
Hofer-Julian: Nice, I was looking for something like this, but missed it.

Hofer-Julian:
> Can we also fix the build so that we get the Windows-only commands like `registry query`. I think that's the only one but I'm not positive.

I am not sure how I would do this.

fdncred:
> I am not sure how I would do this.

probably not for this PR. @hustcer we need to include the windows tools as well.

The problem is, that if we build and generate only on mac or linux, we don't get the windows tool(s).

hustcer:
> probably not for this PR. @hustcer we need to include the windows tools as well.
> The problem is, that if we build and generate only on mac or linux, we don't get the windows tool(s).

I see, I will try to add that later

hustcer:
@Hofer-Julian `commands/categories/dataframe_or lazyframe.md` I think we can remove this file.

Hofer-Julian:
> @Hofer-Julian `commands/categories/dataframe_or lazyframe.md` I think we can remove this file.

Good catch, done.

hustcer:
Thanks, let's merge it.

PR #795, opened 2023-02-24
Add short flags to command docs, fix #509

### LINKED ISSUE #509: Documentation lacking short flags?
In commands such as `ls` and `print`, which are probably the commands new users will want to use the most, the short flags (eg `-n` for `print`) are not present in the website. Is this a deliberate choice, or should they be added as well?

hustcer:
Add short flags to command docs, fix #509

Before, `ls` for example:
```markdown
## Parameters

 -  `pattern`: the glob pattern to use
 -  `--all`: Show hidden files
 -  `--long`: Get all available columns for each entry (slower; columns are platform-dependent)
 -  `--short-names`: Only print the file names, and not the path
 -  `--full-paths`: display paths as absolute paths
 -  `--du`: Display the apparent directory size ("disk usage") in place of the directory metadata size
 -  `--directory`: List the specified directory itself instead of its contents
 -  `--mime-type`: Show mime-type in type column instead of 'file' (based on filenames only; files' contents are not examined)
```

After, `ls` for example:
```markdown
## Parameters

 -  `pattern`: the glob pattern to use
 -  `--all(-a)`: Show hidden files
 -  `--long(-l)`: Get all available columns for each entry (slower; columns are platform-dependent)
 -  `--short-names(-s)`: Only print the file names, and not the path
 -  `--full-paths(-f)`: display paths as absolute paths
 -  `--du(-d)`: Display the apparent directory size ("disk usage") in place of the directory metadata size
 -  `--directory(-D)`: List the specified directory itself instead of its contents
 -  `--mime-type(-m)`: Show mime-type in type column instead of 'file' (based on filenames only; files' contents are not examined)
```

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -97,7 +97,7 @@ def command-doc [command] {
       if $param.parameter_type == "positional" {
         $" -  `($param.parameter_name)`: ($param.description)"
       } else if $param.parameter_type == "switch" {
-        $" -  `--($param.parameter_name)`: ($param.description)"
+        $" -  `--($param.parameter_name)\(-($param.short_flag)\)`: ($param.description)"
       } else if $param.parameter_type == "named" {
         $" -  `--($param.parameter_name) {($param.syntax_shape)}`: ($param.description)"
       } else if $param.parameter_type == "rest" {
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -2,7 +2,7 @@ def main [] {
   # Old commands are currently not deleted because some of them
   # are platform-specific (currently `exec`, `registry query`), and a single run of this script will not regenerate
   # all of them.
-  #do -i { rm book/commands/docs/*.md }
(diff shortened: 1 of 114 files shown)


make_docs.nu
```diff
@@ -97,7 +97,7 @@ def command-doc [command] {
       if $param.parameter_type == "positional" {
         $" -  `($param.parameter_name)`: ($param.description)"
       } else if $param.parameter_type == "switch" {
-        $" -  `--($param.parameter_name)`: ($param.description)"
+        $" -  `--($param.parameter_name)\(-($param.short_flag)\)`: ($param.description)"
```
sholderbach, line 100: Maybe worth visually separating:
```suggestion
        $" -  `--($param.parameter_name)` `\(-($param.short_flag)\)`: ($param.description)"
```

sholderbach (approved):
Awesome!

PR #799, opened 2023-02-25
feat: Remove welcome message

gustavopmaia:
Add tutorial how to remove the welcome message.

This PR complements PR #8217 in nushell repository


book/configuration.md
```diff
@@ -67,6 +67,12 @@ Nushell follows underneath orders to locate the editor:
 
 You can learn more about setting up colors and theming in the [associated chapter](coloring_and_theming.md).
 
+## Remove Welcome Message
+
+To remove the welcome message, you need to edit your `config.nu`. Go to global configuration `let-env config` and set `show-banner` option to false, like this:
+
+@[code](@snippets/installation/remove_welcome_message.nu)
+
```
amtoine, line 75: i would add, for those who are not sure where to find the config file
- how to get the default config file, i.e. either by
  - going to https://github.com/nushell/nushell/blob/main/crates/nu-utils/src/sample_config/default_config.nu, grabbing the content of the file and putting it manually inside `$nu.config-path`
  - running something like
```bash
nushell> http get https://raw.githubusercontent.com/nushell/nushell/main/crates/nu-utils/src/sample_config/default_config.nu | save --force $nu.config-path
```
- how to open the config file, i.e. with
```bash
nushell> config nu
```

amtoine (commented):
a proposition of some more details for those who might not know (1) where to get a default config file and (2) how to open it for edition :yum:

book/configuration.md
sholderbach: There is `config reset` and we prompt at first startup so I don't think pulling from a potentially changing URL is recommended.

amtoine: yeah you're right :thinking: 

then, only `config nu` could be useful i think :yum:

gustavopmaia: Ok, I'll add the config nu part in the docs

PR #801, opened 2023-02-27
Rename $env.PATH to $env.Path to reflect actual shell behavior

debemdeboas:
The documentation states that `$env.PATH` lets the user quote "See the current path". This is not what happens though, the command errors out.

```
❯ $env.PATH
Error: nu::shell::name_not_found (link)

  × Name not found
   ╭─[entry #16:1:1]
 1 │ $env.PATH
   ·      ──┬─
   ·        ╰── did you mean 'Path'?
   ╰────
```

<details><summary>Screenshot</summary>

image

</details>

The correct command *in Windows* to get the current path is `$env.Path`, not `$env.PATH`.

### DIFF
--- a/book/coming_from_bash.md
+++ b/book/coming_from_bash.md
@@ -44,7 +44,7 @@ Note: this table assumes Nu 0.60.0 or later.
 | `stat $(which git)`                  | `stat (which git).path`                          | Use command output as argument for other command                  |
 | `echo /tmp/$RANDOM`                  | `$"/tmp/(random integer)"`                       | Use command output in a string                                    |
 | `cargo b --jobs=$(nproc)`            | `cargo b $"--jobs=(sys \| get cpu \| length)"`   | Use command output in an option                                   |
-| `echo $PATH`                         | `$env.PATH`                                      | See the current path                                              |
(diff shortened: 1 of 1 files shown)


fdncred:
This is a Windows vs *nix thing. On Windows, path is usually `Path`. On MacOS and Linux it's usually `PATH`.

debemdeboas:
You're right. Maybe adding both examples would be beneficial?

debemdeboas:
I changed it to be like so:

image

fdncred:
Looks good but I wouldn't say posix since nushell isn't posix compatible. I'd just say MacOS and Linux.

fdncred:
You could maybe say `Non-Windows` - I'm not sure what the android path looks like but we have people who use it on android.

debemdeboas:
As soon as I get home I will change it to Non-Windows for us to see.

debemdeboas:
Sorry for the delay.
I changed it to (Non-Windows) and removed (Windows) since it was redundant.

image

fdncred:
ok, i don't see the changes in the PR yet.

debemdeboas:
I didn't change it in the PR, I was waiting for your opinion :D

fdncred:
Thanks for helping out. I think it looks good enough for now.

PR #816, opened 2023-03-12
Dashes instead of underscores in command comparisons

bergus:
I know nothing about nu, but https://www.nushell.sh/commands/docs/is-empty.html and https://www.nushell.sh/commands/docs/with-env.html spell them with dashes.


sholderbach:
Awesome thanks!

PR #820, opened 2023-03-13
Fix show_banner config item

martinetd:
the config option to disable banner is show_banner, not show-banner

That does not leave a great impression when the first thing read in documentation is incorrect...


jntrnr:
thanks!

PR #821, opened 2023-03-13
Add description for #8205 and #8014

NotLebedev:
Add description for #8205 and #8014


sholderbach:
Awesome thanks!

We may need to summarize the direction we want to move the output behavior but that can be done with the summary of the other changes.

PR #822, opened 2023-03-13
Releae notes for dirs commands

bobhy:


fdncred:
I'm wondering if we should include this or not in the release notes since std.nu isn't really a thing yet. Granted dirs.nu is in the repository, but I'm just not sure. What do others think?

amtoine:
> I'm wondering if we should include this or not in the release notes since std.nu isn't really a thing yet. Granted dirs.nu is in the repository, but I'm just not sure. What do others think?
>
maybe this should come after a "true" announce of the standard library? :thinking:

kubouch:
Let's put this to the next release notes

bobhy:
Hmm, the dirs work is not tightly integrated with std.nu. we don't even
have to say anything about std.nu right now, if we're not ready.  This
is"just" an alternative implementation and doesn't make any changes to
user's environment (unless user does it themselves). So I'd say it's ok to
ship this pre std (speaking impartially, even though it's my feature).

Regards
Bob

PR #828, opened 2023-03-15
Update command docs to 0.77

Hofer-Julian:
Looking forward to seeing the expected output of commands in the docs :)


Hofer-Julian:
I had to manually fix the docs for xml.

You can find the changes here: https://github.com/nushell/nushell/pull/8462

fdncred:
@Hofer-Julian I landed your xml fix. Do you want to re-run these to make sure it works?

fdncred:
Maybe we can see what @hustcer has to say, he always works on this.

hustcer:
Looks good to me, I think we can land it, thanks

fdncred:
This looks good. I love how the examples show up so nice. Nice work!

However, the next time we run these commands we need to try and make sure that we have all the plugins installed and registered. For example, looking at the help for `query web` has virtually nothing where the command in the repl looks like this.
```
query web --help
execute selector query on html/web

Usage:
  > query web {flags}

Flags:
  -q, --query <String> - selector query
  -m, --as-html - return the query output as html
  -a, --attribute <String> - downselect based on the given attribute
  -t, --as-table <Table> - find table based on column header list
  -i, --inspect - run in inspect mode to provide more information for determining column headers

Examples:
  Retrieve all <header> elements from phoronix.com website
  > http get https://phoronix.com | query web -q 'header'

  Retrieve a html table from Wikipedia and parse it into a nushell table using table headers as guides
  > http get https://en.wikipedia.org/wiki/List_of_cities_in_India_by_population
    | query web -t [Rank City 'Population(2011)[3]' 'Population(2001)' 'State or union territory']

  Pass multiple css selectors to extract several elements within single query, group the query results together and rotate them to create a table
  > http get https://www.nushell.sh | query web -q 'h2, h2 + p' | group 2 | each {rotate --ccw tagline description} | flatten

  Retrieve a specific html attribute instead of the default text
  > http get https://example.org | query web --query a --attribute hre
```

Hofer-Julian:
Mmmh weird, do I have to register plugins every time I run install_all.sh?

hustcer:
The plugins will be registered if you have registered them before

Hofer-Julian:
> The plugins will be registered if you have registered them before

Then the problem probably lies somewhere else

PR #831, opened 2023-03-16
Update plugin docs for v0.77

hustcer:
Update plugin docs for v0.77


hustcer:
Looks like some docs have invalid characters, I will check it later

Issue #815, opened 2023-03-12
Overlay, is it last active or activated 


petrisch:
Hi there

I'm trying to get myself back into translating again some pages into [german](https://github.com/nushell/nushell.github.io/compare/main...petrisch:nushell.github.io:main).

One minor thing:
In `overlays.md` it says:

`...prompt will be preceded by the name of the last active overlay.`

Is it really the last active or the last that "has been activated"?
Or are they understood as the same and I am being picky?

The difference would be in my mind, that they "all" are active, 
but the one prompted is the one that has been activated the last.
Or am I wrong?

PR #832, opened 2023-03-18
DE translation for overlays

petrisch:
I just went for one version for #815 now.


hustcer:
Thanks, let's merge it

PR #833, opened 2023-03-18
docs: Add section on removing ANSI sequences with find command

Phreno:
I was struggeling on this one, not knowing a thing about the ansi stuff.
I think adding this in the doc could be helpfull.

### DIFF
--- a/commands/docs/find.md
+++ b/commands/docs/find.md
@@ -90,3 +90,13 @@ Find value in records
 ╰───┴─────────┴─────────╯
 
 ```
+
+### Removing ANSI Escape Sequences
+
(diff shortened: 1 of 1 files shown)


fdncred:
Unfortunately, it cannot go in the commands because those are systematically generated with a nushell script each release. So, if we were to land this now, the next time we run the script, these changes will be overwritten.

So, this will need to be moved somewhere else or put in the nushell rust code for the find command.

Phreno:
Ohh lord, this was my first pull request on github, and I was scared of misclick for a poor md file, and your telling me that I've to look for rust code.

By the way, I should have guessed that the documentation was generated.
I'll take a look for that.

PR #836, opened 2023-03-19
Add context aware custom completions to the book

thomasgoulet:
I saw no reference of this anywhere in the documentation so I propose we add this to the custom completions page. Here's an [example](https://github.com/nushell/nu_scripts/blob/main/kubernetes/kubernetes.nu#L172) of this in the nu_scripts repository which is where I figured out it's possible to do it.

English isn't my first language don't hesitate to correct me on my grammar and phrasing.


fdncred:
Thanks

PR #837, opened 2023-03-20
DE translation for command_signature

petrisch:
I have done another one over the weekend.
But I am curious. The .vuepress/config.js seems to have changed.
Is there somewhere else now, I have to add this to appear on the site?


hustcer:
Yeah, you can update the side nav in this file: https://github.com/nushell/nushell.github.io/blob/main/.vuepress/configs/sidebar/de.ts

hustcer:
Thanks

PR #838, opened 2023-03-20
add a note about "value / description" completion

amtoine:
i love the new "Custom completion" page, it's really great to learn about all that stuff thanks to #836 :relieved: 

i simply thought there could be a little note about `value / description` based completion, about the fact that only the `value` is used in the actual argument when selecting an option :yum:

### DIFF
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -115,6 +115,7 @@ def my_commits [] {
 ```
 
 > **Note**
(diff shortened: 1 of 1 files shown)


amtoine:
looks like this after an `npm run dev` :ok_hand: 
2023-03-20_19-38-07_3840x1080

PR #843, opened 2023-03-24
feat: docs on the `command_not_found` hook

stevenxxiu:
This PR adds docs on the `command_not_found` hook, with an example in *Arch Linux*.


fdncred:
Thanks

PR #846, opened 2023-04-01
add a warning and a note about changing `show_banner`

amtoine:
Related to https://github.com/nushell/nushell/issues/8698.

This PR adds a **Warning** about the already existing `show_banner: true` in the default config and a **Note** about the idea of throwing an error when fields are defined multiple times in a record.


fdncred:
Thanks but I think we shouldn't land this. I think your last PR that put it at the top should suffice for now.

amtoine:
> Thanks but I think we shouldn't land this. I think your last PR that put it at the top should suffice for now.
>
gotcha :crossed_fingers:

PR #856, opened 2023-04-04
Fix the use of comments

sholderbach:
Reflect the lexing change in https://github.com/nushell/nushell/pull/8151 

Addresses https://github.com/nushell/nushell/issues/8731


fdncred:
good catch!

PR #858, opened 2023-04-04
Fix the broken docs for touch, which now has no -d flag

onthebridgetonowhere:


fdncred:
the documentation for the commands will be auto-generated. so, this change will get overwritten by whatever the command generates.

onthebridgetonowhere:
> 

Gotcha, thank you for explaining that!!!

fdncred:
This should be fixed now that the docs have been regenerated.

PR #862, opened 2023-04-05
use || in closures when configuring prompt

harrysarson:
As of https://www.nushell.sh/blog/2023-04-04-nushell_0_78.html#now-required-in-closures this is required.

I have tested the starship snippet which works locally for me. Read accross test results for oh-my-posh.


fdncred:
Thanks.

PR #864, opened 2023-04-06
Add ppid to ps examples

pingiun:
Note that this is unreleased. This PR should be blocked until https://github.com/nushell/nushell/pull/8750 has landed in a release


hustcer:
Thanks for the PR, `commands/docs/ps.md` need not be modified, all docs in `commands/docs` folder were created by a script, they will be updated after next nushell release.

fdncred:
I think the examples are probably fine to leave here but the commands/docs/ps.md will get overwritten as hustcer has said.

hustcer:
Thanks, we will merge it after nu v0.79 been released

PR #865, opened 2023-04-07
Fix syntax error in blog post

efulleratindeed:
I found a small syntax error while looking at this blog post. I glanced through the rest of the code examples too but nothing else stuck out to me


sholderbach:
Thanks!

PR #866, opened 2023-04-10
fix colors

er2:
I tried these out and got the error message "expected string, found number-like value (hint: use quotes or backticks)" and that worked.

### DIFF
--- a/commands/docs/ansi_gradient.md
+++ b/commands/docs/ansi_gradient.md
@@ -29,24 +29,24 @@ usage: |
 
 draw text in a gradient with foreground start and end colors
 ```shell
-> 'Hello, Nushell! This is a gradient.' | ansi gradient --fgstart 0x40c9ff --fgend 0xe81cff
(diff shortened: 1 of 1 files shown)


amtoine (approved):
that is a recent change.

thanks for the catch :yum:

fdncred:
Thanks for the change but we cannot make changes to the commands this way because the commands are auto generated with a script. What that means is that the next time we run the script this change will be overwritten. In order for this change to remain, the ansi gradient rust code must be changed. If you'd like to make the change there, we'd accept it.

amtoine:
i did not realize it was in the command page, woopsie :eyes:

er2:
I found that here: https://github.com/nushell/nushell/blame/7413ef28243d7586037b8b0b9133e6e7c7d44255/crates/nu-command/src/platform/ansi/gradient.rs#L75

And it looks like it's already fixed. Does something need to be done to regenerate this markdown?

amtoine:
> Does something need to be done to regenerate this markdown?
>
you need to run the `make_docs.nu` script in the root of this repo.
but this is done around each release, so no need to generate that right now, if i have the full flow of the site correctly in mind :yum:

PR #870, opened 2023-04-13
Add info about echo, print and log commands to the book

presidento:


presidento:
You may add the `wait-after-next-release` label to it. (The standard library will be released then.)

book/stdout_stderr_exit_codes.md
```diff
@@ -54,6 +54,24 @@ If we try to run the external `cat` on a file that doesn't exist, we can see wha
 ╰───────────┴─────────────────────────────────────────────╯
 ```
 
+## `echo`, `print`, and `log` commands
+
+The [`echo`](/commands/docs/echo.md) command is mainly for _pipes_. It returns its arguments, ignoring the piped-in value. There is usually little reason to use this over just writing the values as-is.
+
+In contrast, the [`print`](/commands/docs/print.md) command prints the given values to stdout as plain text. It can be used to write to standard error output, as well.
```
amtoine, line 61: maybe add that `print`ed things can not be piped?

for instance
```bash
[1,2,3] | each { |it|
    print $it
} | str join ", "
```
only prints
```bash
1
2
3
```
and not `1, 2, 3`

book/stdout_stderr_exit_codes.md
```diff
@@ -54,6 +54,24 @@ If we try to run the external `cat` on a file that doesn't exist, we can see wha
...
+
+The [standard library](/book/standard_library.md) has commands to write out messages in different levels. For example:
```
amtoine, line 63:
```diff
-The [standard library](/book/standard_library.md) has commands to write out messages in different levels. For example:
+The [standard library](/book/standard_library.md) has commands to write out messages in different logging levels. For example:
```
maybe? :yum:

presidento:
I agree, that this topic is a bit unrelated to the existing chapter, but much more unrelated to other chapters. And this is just a small addition to the other side of stdout and stderr.

I have no strong preference where to put this section, so if you can point me another chapter, I gladly move it there.

amtoine (commented):
thanks for the minor changes :+1:

amtoine (approved):
> I agree, that this topic is a bit unrelated to the existing chapter, but much more unrelated to other chapters. And this is just a small addition to the other side of stdout and stderr.
> 
> I have no strong preference where to put this section, so if you can point me another chapter, I gladly move it there.
>
yeah i see what you mean and i can't think of a better place for this :thinking: 

i think the changes are great, so i'll approve this :+1: 
if another member of the team thinks of a better place, i'll still approve this as before :muscle:

PR #871, opened 2023-04-16
Fix old keybding for help menu

uselesspseudo:
The commit 0ea7a38c21d3c365afdd86f3b9a329938dc99b8 change the keybinding to F1


sholderbach:
Thanks! Good call!

PR #873, opened 2023-04-17
Improve misleading text format on note

Kissaki:
Formatting the asterisk as an inline code block `*` with a `-` dash separator implies that it were a parameter to or syntax to the discussed topic (scope) and mentioned before. However, this is not the case. It is formatted in a misleading manner.

The asterisk seems to be intended as a footnote indicator. But no asterisk exists in the text before, so it does not follow established asterisk footnote practices either.

These issues are solved by removing the asterisk and making it a simple sentence.

To clarify that the paragraph is an additional note the text is formatted in italics.


fdncred:
Agreed. Thanks.

PR #876, opened 2023-04-19
Dataframes

presidento:
Ensure that every Nushell command can be executed.


fdncred:
Thanks. We've been needing this for a while.

PR #877, opened 2023-04-22
Fix ranges in `str substring` examples

Ifthel:
[`str substring` only allows ranges as of 0.78.](https://www.nushell.sh/blog/2023-04-04-nushell_0_78.html#ranges-become-the-standard-range-specifier)
Updates the usage examples for `str substring` in working_with_strings.md for all languages.
Doesn't touch old patch notes.
I couldn't find any other instances of this in the repo (regex: `substring.*\d\s*,\s*\d`), nor any problems with `bytes at`.


hustcer:
Thanks

PR #878, opened 2023-04-22
Update dataframe examples

Ifthel:
Miscellaneous dataframe-related documentation fixes.
- Updates all `x-df` commands to `dfr x` (see nushell/nushell#7998).
- [Adds missing parentheses around captured command output for `let`.](https://www.nushell.sh/blog/2023-04-04-nushell_0_78.html#let-requires-surrounding-parentheses-for-saving-command-output) These were the only instances of such a problem I could find in the entire repo.


hustcer:
Nice 👍

PR #880, opened 2023-04-23
update variable documentation

1Kinoti:
this pr updates variable documentation.

it touches on
- types of variables
- assignment syntax
- question mark operator on cell paths

### DIFF
--- a/book/variables_and_subexpressions.md
+++ b/book/variables_and_subexpressions.md
@@ -1,31 +1,133 @@
 # Variables and Subexpressions
 
-There are two types of evaluation expressions in Nushell: variables and subexpressions. You know that you're looking at an evaluation expression because it begins with a dollar sign (`$`). This indicates that when Nushell gets the value in this position, it will need to run an evaluation step to process the expression and then use the resulting value. Both evaluation expression forms support a simple form and a 'path' form for working with more complex data.
+There are two types of evaluation expressions in Nushell: variables and subexpressions.
+You know that you're looking at an evaluation expression because it begins with a dollar
+sign (`$`). This indicates that when Nushell gets the value in this position, it will
+need to run an evaluation step to process the expression and then use the resulting 
+value. Both evaluation expression forms support a simple form and a 'path' form for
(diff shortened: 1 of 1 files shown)


book/variables_and_subexpressions.md
```diff
@@ -1,31 +1,133 @@
...
+need to run an evaluation step to process the expression and then use the resulting 
+value. Both evaluation expression forms support a simple form and a 'path' form for
+working with more complex data.
 
 ## Variables
 
-The simpler of the two evaluation expressions is the variable. During evaluation, a variable is replaced by its value.
+The simpler of the two evaluation expressions is the variable. During evaluation, a
+variable is replaced by its value. After creating a variable, we can refer to it using
+`$` followed by its name.
```
rgwood, line 14:
```suggestion
The simpler of the two evaluation expressions is the variable. During evaluation, a variable is replaced by its value. After creating a variable, we can refer to it using `$` followed by its name.
```

Can you please fix this and the other places where a sentence or paragraph was split into multiple lines?

1Kinoti: on it

book/variables_and_subexpressions.md
```diff
@@ -1,31 +1,133 @@
...
+nushell will see this as a pipeline with two elements, one of which is not allowed
+in pipleines because it doesn't return a value (`let val = 42`). The correct way to
+declare using a pipeline is to wrap the pipeline in parentheses
+
+```nu
+let val = (42 | math sin)
 ```
 
-## Variable paths
+### Variable Paths
```
rgwood, line 110: Hmm. I hadn't seen the Variable Paths section, I think we should remove it eventually.

[Cell paths](https://github.com/nushell/nushell.github.io/pull/879) aren't really something to do with a _variable_. The cell path `.bar` in `$foo.bar` is the same as the cell path in `(some-command).bar`.

That said, moving/removing this section is probably beyond the scope of this PR; we can do that another time. Thank you for documenting the optional `?` operator!

PR #885, opened 2023-04-24
add the note about the standard library

amtoine:
as there is so much work done and so many things to talk about, i tried to keep that simple :relieved:


amtoine:
thanks to @fdncred's feedback i've made some modifications
- namely make a real introduction to the library, without assuming prior knowledge / goind to external links

amtoine:
@fdncred :blush:

PR #886, opened 2023-04-24
Improve `insert` and `update` examples in working_with_lists.md

Ifthel:
Improves the `insert` and `update` examples in working_with_lists.md by showing their results and adding highlighting.
Copied from the zh-CN version.


hustcer:
Thanks

PR #887, opened 2023-04-24
Use explicit printing in custom_commands.md

Ifthel:
[Non-final values are no longer automatically printed as of 0.78.](https://www.nushell.sh/blog/2023-04-04-nushell_0_78.html#we-no-longer-automatically-print-values-as-part-of-a-script) This updates custom_commands.md accordingly.

I've also updated the zh-CN version's examples so they work, but cannot confirm that the surrounding text is still relevant. It's a bodge.

I wasn't sure whether the tip about piping to null should be kept; I moved it somewhere appropriate, but its example is weak.

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -13,7 +13,7 @@ def greet [name] {
 ```
 
 ::: tip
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

PR #892, opened 2023-04-28
fix the "`std` import" section of the 0.79 changelog

amtoine:
a Nushell user has reported issues in the changelog about the standard library, here is a quick fix :ok_hand:


fdncred (approved):
better

amtoine:
> better
>
it was not even correct before :eyes: 
i was in a hurry the other night i wrote that :scream:

PR #894, opened 2023-04-29
Fixing Tiny Typo - Update aliases.md

Askerad:
Fixing small typo that broke formatting on last few paragraphs of the page


amtoine (approved):
good catch, thanks!

it's true that the [bottom of *Aliases*](https://www.nushell.sh/book/aliases.html#aliases) is not great :open_mouth:

PR #895, opened 2023-04-29
Fixed link in design_notes to how_nushell_code_gets_run

PhilVoel:
Link used to point to book/how_nushell_code_gets_run but since both files are in the same directory, specifying the directory **both** files are in breaks it.


fdncred:
Thanks

PR #897, opened 2023-05-01
update link to job.nu file

maxomatic458:


amtoine (approved):
thanks!

fdncred:
I'm guessing there may be several places that need updates like this, since we reorganized nu_scripts. Thanks!

maxomatic458:
i think that page might still be outdated in some other spots

I cant get the job command to work, running the example command ``job spawn { echo 3 }`` results in this error

```
$ job spawn { echo 3 }
Error: nu::shell::external_command (link)

  × External command failed
    ╭─[job.nu:12:1]
 12 │     let env_path = $nu.env-path
 13 │     let source_code = (view source $command | str trim -l -c '{' | str trim -r -c '}')
    ·                                    ────┬───
    ·                                        ╰── Cannot convert block to a string
 14 │     let job_id = (pueue add -p $"nu --config \"($config_path)\" --env-config \"($env_path)\" -c '($source_code)'")
    ╰────
  help: All arguments to an external command need to be string-compatible```
  
  im on nu version ``0.74.0``

PR #900, opened 2023-05-05
book/coming_from_bash - s/hide/hide-env

tzemanovic:
hi, I tried to use `hide` for an env var and it didn't work


hustcer:
Thanks

PR #914, opened 2023-05-18
Minor improvements to the installation chapter

mrkkrp:
* The page alternates between "we" and "you". It is perhaps better to stick to "you".
* Moved the snippet that demonstrates running Nushell by calling `nu` higher up where it is first mentioned that the main app is called `nu`. The snippet was called `snippets/installation/crates_run_nu.sh`, however there is nothing specific to crates in it, so I renamed it accordingly.
* Added a short explanation what `--release` means when compiling from source.


fdncred (approved):
Thanks

PR #919, opened 2023-05-21
Supply table instead of closure to merge command in tables.md

lavafroth:
If the merge command is supplied an expression in curly braces, nushell errors out expecting input and argument, to be both record or both table.

Tested on nushell 0.80.0

To reproduce this error, run:

```sh
let first = [[a b]; [1 2]]
let second = [[c d]; [3 4]]
$first | merge { $second }
```

We get:

```
× Pipeline mismatch.
   ╭─[entry #5:1:1]
 1 │ $first | merge { $second }
   · ───┬──   ──┬──
   ·    │       ╰── expected: input, and argument, to be both record or both table
   ·    ╰── value originates from here
   ╰────
```

We don't get errors if we use:

```sh
$first | merge $second
```

### DIFF
--- a/cookbook/tables.md
+++ b/cookbook/tables.md
@@ -11,7 +11,7 @@ Examples shown in [`Working with tables`](../book/working_with_tables.md) work f
 ```
 > let first = [[a b]; [1 2] [3 4]]
 > let second = [[c d]; [5 6]]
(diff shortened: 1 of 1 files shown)


amtoine (approved):
sounds right to me, thanks for the fix :blush:

PR #920, opened 2023-05-22
Fix references to match in the "coming from" pages to reflect its current usage in Nu

casually-blue:
I have updated the tables to show what match is equivalent to in other languages. There are a couple places that I simply removed the references since there was no closely equivalent feature.


fdncred:
Thanks

PR #921, opened 2023-05-23
update plugins.md in contributor-book

BrianLondon:
Updated `plugins.md` in the Contributor Book to reflect the current API. The page in its current form appears badly out of date. While the Contributor Book is not linked from within the main docs website, https://www.nushell.sh/contributor-book/plugins.html shows up at the top of web searches for terms like "nushell plugin tutorial" and at least confused me.

I know there are Rust and Python example plugins in the Nushell repo, but I appreciated the simplicity of the example that was in the Contributor Book along with the narrative walk through. I just wished it had reflected the current API. This PR aims to fix that.


fdncred:
I see this as a step forward but some of the old parts are still wrong. I also have not checked the code to ensure it works. However, I don't think that should stop us from landing this.

PR #925, opened 2023-05-25
Remove unnecessary 0.59 binaries

hustcer:
Remove unnecessary 0.59 binaries, It's old and outdated

### DIFF
--- a/blog/2022-03-01-nushell_0_59.md
+++ b/blog/2022-03-01-nushell_0_59.md
@@ -14,15 +14,6 @@ Today, we're releasing a beta release of 0.60 of Nu. This is to get early feedba
 
 <!-- more -->
 
-# Where to get it
-
-Nu 0.59 is available only as pre-built binaries.
-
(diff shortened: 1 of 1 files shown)


fdncred:
This is the only copy we have of these files. Is there somewhere we can place them, like in a release somewhere, so we don't lose them?

hustcer:
If you just want to backup them somewhere and don't care about where they are, I think I can create a release here https://github.com/hustcer/nu-release/releases and upload these binaries

fdncred:
can we put them in the releases of this repo?

hustcer:
> can we put them in the releases of this repo?

Yes, but this repo need to be cleaned, and all the tags will be removed after the clean

fdncred:
what about uploading those binaries as an old release, after it's cleaned?

amtoine (approved):
this is all fine with me if it speeds up `git clone`ing :ok_hand:

Issue #924, opened 2023-05-25
Clean up git pack file


BrianLondon:
A clean sync on this repository requires downloading over a GB of data, almost all of which is the git pack file `.git/objects/pack/pack-da425480b918b7d04013e4ffd99bc2edd76c899f.idx`. The objects taking up the most space in there are the checked in binary files for the 0.59 release that appear to be still bundled with the docs and bunch of what appears to be built js bundles of the form `assets/js/app.9caf494d.js` that should never have been checked in and have since been deleted but are still embedded in the pack file downloaded when cloning the repo.

There are instructions [here](https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-the-git-repository) and [here](https://acquia.my.site.com/s/article/360004334093-Removing-large-files-from-Git-without-losing-history) on how to clean large files out of the git history. This needs to be done by someone with atypical permissions on the repo and can't be fixed with a PR.

hustcer:
We have discussed this matter before and I totally agree with you

hustcer:
```shell
# Nushell, get all files larger than a certain size
git rev-list --objects --all 
  | git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' 
  | sed -n 's/^blob //p' 
  | detect columns -n 
  | rename object size name 
  | sort-by size -n
  | update size {|it| $it.size | into filesize }
  | where size > 1MB
  | uniq-by name
  | reverse
```
Output:
```console
   #                    object                     size                            name
────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   0   2f675f4eceac1b19f2a238bbcb48a85bdee65ba3   15.5 MB   .vuepress/public/binaries/nushell_0_59_linux.tar.gz
   1   bc8064c8b9a55f50aed3f2528ab10d9b3284ad47   13.5 MB   .vuepress/public/binaries/nushell_0_59_macOS.zip
   2   2fd080f6e4bd9ef5b2c0ca7d24c329e3527fad44   10.6 MB   .vuepress/public/binaries/nushell_0_59_windows.msi
   3   c1a1a447e98b60f8e4ee352b5547201c71ee2a52   10.6 MB   .vuepress/public/binaries/nushell_0_59_windows.zip
   4   b6e2f3b5cc5104d956ec70674ab3c4322b90e218    4.3 MB   assets/images/0_19_autocomplete.gif
[...]

PR #926, opened 2023-05-26
Nushell Docs Repo Cleanup Plan

hustcer:
This is a **Tracking PR** for the Nushell Docs Repo Cleanup Task

### TODO:

- [x] Remove the old outdated 0.59 release packages in `.vuepress/public/binaries`: Related PR https://github.com/nushell/nushell.github.io/pull/925
- [x] Backup the original repo: Create an backup repo maybe https://github.com/nushell/nu-docs-backup, which is the mirror of https://github.com/nushell/nushell.github.io
- [ ] Cleanup the original `nushell.github.io` repo
- [x] Change the deploy workflow so that the docs will be deployed without `gh-pages` branch, we won’t keep that branch in docs’ repo any more, Related PR: https://github.com/nushell/nushell.github.io/pull/937
- [x] We can also create a test repo after the clean and deploy it somewhere to make sure everything work as expected
- [ ] Make a force push to update the cleaned repo to the original repo, it CAN NOT be done via a PR
- [ ] Make an announcement to make sure that All the docs' contributors delete their old fork then create a new fork, and delete their local `nushell.github.io` repo and make a fresh clone for future contribution.

### Optional Actions to Slow Down the Increase of Repo Size:

- [x] It would be better if we compress all the images before we do the cleanup work, Related PR: https://github.com/nushell/nushell.github.io/pull/929
- [x] All the newly added images from a PR should be compressed automatically before they get merged, Related PR: https://github.com/nushell/nushell.github.io/pull/932 https://github.com/nushell/nushell.github.io/pull/933  The images will be compressed after they have been merged, it's not good enough, but better than none

### Other Notes

1. After the cleanup only one branch will be kept, that's the `main` branch
2. The commit history will be kept but all the commit ids will be changed after cleanup
3. The [nushell.sh](http://nushell.sh/) site maybe unavailable for a while, from quite a few minutes to maybe an hour
4. We won’t accept any PR when the cleanup process starts until it’s finished

Close https://github.com/nushell/nushell.github.io/issues/924

### DIFF
--- a/i18n.nu
+++ b/i18n.nu
@@ -131,12 +131,12 @@ def main [
         'outdated' => {
             if ($lng | is-empty) {
                 print $'(ansi r)A locale code required, available locales: ($locales), Please try again!(ansi reset)(char nl)'
-                exit --now
+                exit
             }
             let available = ($lng | str downcase) in $locales
             if (not $available) {
                 print $'(ansi r)Unsupported locale, available locales: ($locales), Please try again!(ansi reset)(char nl)'
-                exit --now
+                exit
             }
             check-outdated-translation $lng
         }


amtoine:
looks great to me, as agreed before :+1:

amtoine:
thanks for tackling this @hustcer :relieved:

hustcer:
I think we need not do anything else, after what's have been done, a fresh clone size is `95.92 MB`, It's much smaller than the previous size which is 1 GB+

PR #930, opened 2023-05-26
Try to add a new workflow to compress images on new pull request

hustcer:
Try to add a new workflow to compress images on new pull request


amtoine:
wow we can do that :open_mouth: 
CI is quite amazing :star_struck:

PR #948, opened 2023-06-13
Modernize and add additional info to plugin tutorial

BrianLondon:
In my [previous update](https://github.com/nushell/nushell.github.io/pull/921) to the plugin tutorial a number of other issues were flagged as not fixed. I believe this addresses those as well as expands a section on the low-level protocol that I think is helpful for understanding the Python section.

I've learned a lot while working on these docs but I'm still not super familiar with the technology so I'm very open to feedback on this. I think it addresses the issues flagged as unresolved in the earlier PR.

### DIFF
--- a/contributor-book/plugins.md
+++ b/contributor-book/plugins.md
@@ -6,24 +6,27 @@ title: Plugins
 
 ## Protocol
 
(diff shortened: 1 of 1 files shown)


fdncred:
nice work, thanks!

PR #949, opened 2023-06-14
Remove mentioning ZB and ZiB

hanjunghyuk:
See also: nushell/nushell#9337, nushell/nushell#9427

To be consistent with nushell, this PR removes mention of ZB and ZiB.


amtoine (approved):
thanks a lot, i do not see any more `zb` nor `zib` in the book :ok_hand:

PR #951, opened 2023-06-14
Update aliases.md and setup.md

mh-trimble:
- Add an example how to set the EDITOR variable in env.nu to setup.md
- Update misleading/confusing section about persisting aliases to aliases.md + german translation
- Added some info about pipes in aliases + german translation

### DIFF
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -40,3 +40,11 @@ def uuidgen [] { ^uuidgen | tr A-F a-f }
 ```
 
 See more in the [custom commands](custom_commands.md) section of this book.
+
+Or a more idiomatic example with nushell internal commands
+
+```
+def lsg [] { ls | sort-by type name -i | grid -c | str trim }
+```
+
+displaying all listed files and folders in a grid.
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -30,4 +30,4 @@ Your useable aliases can be seen in `$nu.scope.aliases`.
 
 To make your alias persistent it must be added to your _config.nu_ file.
 
-For more details about how to persist aliases so that they're visible when you start up Nushell, see the [configuration chapter](configuration.md).
+Just put them in your _config.nu_ by running `config nu` and `exec nu` to reload your nushell.
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -28,6 +28,5 @@ Your useable aliases can be seen in `$nu.scope.aliases`.
 
 ## Persisting
 
-To make your alias persistent it must be added to your _config.nu_ file.
-
-Just put them in your _config.nu_ by running `config nu` and `exec nu` to reload your nushell.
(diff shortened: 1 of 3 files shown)


book/aliases.md
```diff
@@ -30,4 +30,4 @@ Your useable aliases can be seen in `$nu.scope.aliases`.
 
 To make your alias persistent it must be added to your _config.nu_ file.
 
-For more details about how to persist aliases so that they're visible when you start up Nushell, see the [configuration chapter](configuration.md).
+Just put them in your _config.nu_ by running `config nu` and `exec nu` to reload your nushell.
```
fdncred, line 33: `exec nu` is only a linux/mac thing and not cross platform. I prefer it to end with ` and restart nushell` or something like that.

mh-trimble: sorry haven't used windows in decades - so forgot that - I'll take care of that!

mh-trimble: ah and I saw there is a german version that I forgot!

book/aliases.md
```diff
@@ -28,6 +28,16 @@ Your useable aliases can be seen in `$nu.scope.aliases`.
...
 ## Persisting
 
-To make your alias persistent it must be added to your _config.nu_ file.
+To make your aliases persistent they must be added to your _config.nu_ file by running `config nu` to open an editor and inserting them.
+Then run `exec nu` on linux/mac to reload your nushell or on windows just restart your nushell.
 
-For more details about how to persist aliases so that they're visible when you start up Nushell, see the [configuration chapter](configuration.md).
+## Piping in aliases
+
+Note that `alias uuidgen = uuidgen | tr A-F a-f` (to make uuidgen on mac behave like linux) won't work.
```
sholderbach, line 36: From a readers perspective I would suggest replacing those two comparitively obscure unix commands with something from nushell's command set or leading in a sentence as not to confuse with command names. (e.g. `tr` wouldn't be an philosophically endorsed nushell solution even though its blazingly fast)

mh-trimble: this is just the use case I had, which I thought would be a simple example - and the point of it is how to use something outside nu. but i am not familiar with what would be more idiomatic - as I am just starting, so if you have some suggestion I'll just work them in.

fdncred: Agreed. I don't really like the `exec` or the other externals. Although I do think it's important to demonstrate how to use externals.

mh-trimble: I'll change the example today!

amtoine: looking forward to it @mh-trimble :yum:

mh-trimble: sorry took me a bit longer - this week was busy. Updated the German version removing the `exec nu` bit and added the more idiomatic example.

fdncred (approved):
Thanks

amtoine (approved):
sounds good to me, thanks @mh-trimble :blush:

PR #952, opened 2023-06-15
Bring shells documentation up to date

bgeron:
- `exit` no longer performs shell functions (despite it staying in `crates/nu-command/src/shells/exit.rs` for now)
- Those functions have been moved to `dexit`
- Add numeric argument to `p` and `n`

### DIFF
--- a/commands/docs/dexit.md
+++ b/commands/docs/dexit.md
@@ -1,0 +1,27 @@
+---
+title: dexit
+categories: |
+  shells
+version: 0.80.0
+shells: |
+  Exits the current Nu shell, and go to the next one
+usage: |
+  Exits the current Nu shell, and go to the next one
+---
+
+# <code>{{ $frontmatter.title }}</code> for shells
+
+<div class='command-title'>{{ $frontmatter.shells }}</div>
+
+## Signature
+
+```> dexit```
+
+
+## Examples
+
+Exits the current shell
+```shell
+> dexit
+
+```
--- a/commands/docs/enter.md
+++ b/commands/docs/enter.md
@@ -15,11 +15,12 @@ usage: |
 
 ## Signature
 
-```> enter (path)```
+```> enter (path)...```
 
 ## Parameters
 
  -  `path`: the path to enter as a new shell
+ -  `...rest`: other paths to add as new shells after this one, but not enter
 
 ## Examples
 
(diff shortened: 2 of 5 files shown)


fdncred:
Typically we don't add docs this way because they get regenerated at every release. I'm wondering if we can generate these docs with the script without manually doing this?

bgeron:
> Typically we don't add docs this way because they get regenerated at every release. I'm wondering if we can generate these docs with the script without manually doing this?

Doesn't look like we currently do that, at least. (With the exception of `enter`. I should do something with that.)

`help` currently gets a little confused, because all the (actual) shells commands are aliases to commands within `std dirs` by a different name. Their parameters are invisible to help.

```
〉help n
Advance to the next directory in the list or wrap to beginning.

Alias: n

Expansion:
  next
〉help std dirs next
Advance to the next directory in the list or wrap to beginning.

Usage:
  > std dirs next (N) 

Flags:

Parameters:
  N <int>: number of positions to move. (optional, default: 1)
```

Ideally a good implementation of autogen for these would be able to see the parameter. Alternatively, as a scrappy workaround we could explain the parameter in the main usage.

(Then again, the parameters of these specifically are super simple and could maybe be explained in the main usage.)

bgeron:
I opened https://github.com/nushell/nushell/pull/9450 to fix the usage for `exit`, to match this PR.

amtoine:
@bgeron 
thanks for the PR on the Nushell side :blush:
it will take effect in the website for next release :ok_hand:

bgeron:
@amtoine Just to confirm, are there plans for a docs generator to get `dexit` in? Currently it is absent from the website at all.

amtoine:
> @amtoine Just to confirm, are there plans for a docs generator to get `dexit` in? Currently it is absent from the website at all.
>
not now because we do not have a way to generate static documentation from a Nushell module, i.e. what the standard library is :thinking: 

hopefully this will come with the `nupm` WIP package manager :relieved:

bgeron:
How about we use the manual docs while there is no autogen? The current autogen doesn’t undo these changes (with the exception of `exit`)

amtoine:
> How about we use the manual docs while there is no autogen? The current autogen doesn’t undo these changes (with the exception of `exit`)
>
we could probably have a "standard library" subdirectory in `commands/` where one put manually generated doc for the moment, but i don't think we want to get rid of `commands/docs/` being automatically generated :+1:

PR #958, opened 2023-06-23
Add multiline regex flag to str replace

davidscholberg:
Documentation updates for https://github.com/nushell/nushell/pull/9508.


fdncred:
unfortunately we cannot accept this PR since the docs/ folder is autogenerated in each release and any changes made here will be overwritten. however, we could accept some comments on the PR for the next release, which is in draft in this repo now.

PR #959, opened 2023-06-23
Fix custom completion scripts

JoaquinTrinanes:
This PR fixes two issues I found while setting up and using custom completions.

### `$fish_completer` deleting input

The completer in the docs works mostly fine, but when there are no more valid completions (for example, `ls dir/that/does/not/exist<TAB>` or `nu --invalid-flag<TAB>`) it returned an empty string. This wouldn't normally be an issue, as `from json` would throw an error and the default completer kick in. However, `'' | split row (char newline)` returns `['']`, a valid JSON, and the last span of the completion was being deleted.

I fixed these issue avoiding `split row` and using `from tsv` instead, avoiding some manual parsing in the process. I used the `--no-infer` flag as a technicality, because some commit hashes where being returned as a number instead of a string.

I found another error while testing this fix, but it happens both in the existing version and the new one. Calling `npm` returns an error that makes completions fallback to show the list of files:

image

Edit: I tried redirecting stderr to ignore the error but it looks like it's preventing the rest of the output to reach nushell:
```nushell
let fish_completer = {|spans: list<string>|
    fish --command $'complete "--do-complete=($spans | str join " ")"' err> /dev/null
    | from tsv --noheaders --no-infer
    | rename value description
}
```
image
image

### Invalid multiple completer

I couldn't get the `$multiple_completer` example to work, so I fixed it using a record that maps the commands to override to their completer, and after retrieving it it `default`s to the default completer (adding `$spans.0` to the record causes an error when there are duplicated values).

In this example, I also made the values of the override record a closure, to be consistent with the other examples, avoiding repetition and being able to pass them around (the most obvious example is that they can be passed as an external completer).

---

This is my first contribution in an open source project (hopefully there's more to come), so please let me know if I did something wrong :) Thank you for the amazing work you do!

### DIFF
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -166,10 +166,9 @@ Multiple completers can be defined as such:
 ```nu
 let multiple_completers = {|spans|
     {
-        $spans.0: { default_completer $spans | from json } # default
-        ls: { ls_completer $spans | from json }
-        git: { git_completer $spans | from json }
-    } | each {|it| do $it}
+        ls: $ls_completer
+        git: $git_completer
+    } | get -i $spans.0 | default $default_completer | do $in $spans
 }
 ```
 
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -177,8 +177,8 @@ This example shows an external completer that uses the `fish` shell's `complete`
 ```nu
 let fish_completer = {|spans|
     fish --command $'complete "--do-complete=($spans | str join " ")"'
-    | from tsv --noheaders --no-infer
-    | rename value description
+    | $"value(char tab)description(char newline)" + $in
+    | from tsv --flexible --no-infer
 }
 ```
 
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -178,10 +178,8 @@ This example shows an external completer that uses the `fish` shell's `complete`
 ```nu
 let fish_completer = {|spans|
     fish --command $'complete "--do-complete=($spans | str join " ")"'
-    | str trim
-    | split row "\n"
-    | each { |line| $line | split column "\t" value description }
-    | flatten
+    | from tsv --noheaders --no-infer
+    | rename value description
 }
 ```
 


book/custom_completions.md
```diff
@@ -166,10 +166,9 @@ Multiple completers can be defined as such:
 ```nu
 let multiple_completers = {|spans|
     {
-        $spans.0: { default_completer $spans | from json } # default
-        ls: { ls_completer $spans | from json }
-        git: { git_completer $spans | from json }
-    } | each {|it| do $it}
+        ls: $ls_completer
+        git: $git_completer
+    } | get -i $spans.0 | default $default_completer | do $in $spans
```
amtoine, line 171: i still do not understand this line...

a record with some keys is created and directly accessed with `get -i`.
it's like writing
```bash
let foo = "a"
{
    $foo: 1
    b: 2
    c: 3
} | get -i $foo
```
:point_right: it works but it's much simpler here to write `1` directly :scream: 

why don't we write, in our real example, 
```bash
{ default_completer $spans | from json } | default $default_completer | do $in $spans
```
?

JoaquinTrinanes: I think we're both expecting different things from this code 😅 

My user story would be in the lines of "_as a user, I want carapace as the default completer. However, some completions are incorrect or don't work as well as others, so I want to override the completer used for specific commands_".

> a record with some keys is created and directly accessed with `get -i`. it's like writing
> 
> ```shell
> let foo = "a"
> {
>     $foo: 1
>     b: 2
>     c: 3
> } | get -i $foo
> ```
> 
> 👉 it works but it's much simpler here to write `1` directly 😱

That works, but what happens if `$foo` was `b`?

```nushell
let foo = "b"
{
    $foo: 1
    b: 2
    c: 3
} | get -i $foo
```

```
Error: nu::shell::not_a_list

  × Record field used twice
   ╭─[entry #7:1:1]
 1 │ {
 2 │ $foo: 1
   · ──┬─
   ·   ╰── field first defined here
 3 │ b: 2
   · ┬
   · ╰── field redefined here
 4 │ c: 3
   ╰────
```

In the PR example, that would mean the completer would work for all cases except for the specific ones we want to override, because we'll be duplicating the value in the record.

For example, I like `fish` completions for git, as carapace ones just output a list of commit hashes. So I can define my record like this:
```nushell
{
    git: $fish_completer
} | default $carapace_completer
```

For any other command not in the record, `$carapace_completer` is used.

> why don't we write, in our real example,
> 
> ```shell
> { default_completer $spans | from json } | default $default_completer | do $in $spans
> ```
> 
> ?

I don't quite get this part 😅 If I understand correctly, `{ default_completer $spans | from json }` would already return completions, so this wouldn't work? If it returns a value and `default` doesn't do anything, `do $in $spans` would cause an error, because `$in` is a list of `value, description` records.

That code is similar to what would be written if I wanted to have a fallback completer if the default one doesn't return anything. In fact, I already have that in my config:

```nushell
    {
      __zoxide_z: $zoxide_completer
      asdf: $fish_completer
      git: $git_completer
      gpg: $fish_completer
      nu: $fish_completer
      sed: $fish_completer
      yadm: $yadm_completer
      yay: $fish_completer
    } | get -i $spans.0 | default $default_completer | do $in $spans # up to this point, normal completions
    | if (($in | is-empty) and (not ($fallback_completer | is-empty))) { do $fallback_completer $spans } else { $in } # fallback to another completer if the result is empty
```

I think the concept of fallback completer and completer override might have been mixed here, this PR specifically adds completer override 😁

amtoine: but if `$foo` is `b`, that would crash the completer right?
i'm confused, i've never encountered such a structure in Nushell code :yum: 

agree about the closure being non-`null` and thus `default` would probably not do anything :+1: 

> ```nushell
> {
>     __zoxide_z: $zoxide_completer
>     asdf: $fish_completer
>     git: $git_completer
>     gpg: $fish_completer
>     nu: $fish_completer
>     sed: $fish_completer
>     yadm: $yadm_completer
>     yay: $fish_completer
> } | get -i $spans.0 | default $default_completer | do $in $spans # up to this point, normal completions
> | if (($in | is-empty) and (not ($fallback_completer | is-empty))) { do $fallback_completer $spans } else { $in } # fallback to another completer if the result is empty
> ```
>
this makes more sense to me because there is no `$spans.0` in the keys defining the record.

this is what is bugging me here, i do not see why you'd do that because
- if `$spans.0` is not an explicit key in the record then `get -i $spans.0` selects it immediately and thus the record does not look usefull to me 
- if `$spans.0` is a key of the record, then this all throws an error 

i might just miss something obvious there :scream:

JoaquinTrinanes: Okay let's take a step back 😁 

> this makes more sense to me because there is no $spans.0 in the keys defining the record.

But I don't set `$spans.0` in the record 🤔 In fact, I deleted it from the previously existing docs because it didn't work

> - if $spans.0 is not an explicit key in the record then get -i $spans.0 selects it immediately and thus the record does not look usefull to me

If `$spans.0` is not in the record `get -i $spans.0` will return null only for the items not set in the record. That means there's no override for this value, and `default` kicks in and returns the default one.

> - if $spans.0 is a key of the record, then this all throws an error
If we had the record defined like this
```nushell
{
    $spans.0: ...
   git: ...
}
```
Then yeah it would be a duplicate key and an error would be thrown. However, I don't add it as a key, that's why I need the `-i` flag in `get`, to avoid errors.

Lets do a MWE to showcase how this works:
 
- `$spans` is a list of strings. For example, if I wrote `git check<TAB>` it would be `[git check]`
- That means `$spans.0` can be assumed to be the name of the program that's going to be executed. In this example, `git`

I want all my completions to use carapace, with the exception of `git`. In this case, I want to use the fish completer.
 My config would look like this:
```nushell
let fish_completer = {|spans|
    fish --command $'complete "--do-complete=($spans | str join " ")"' ...
}


let carapace_completer = {|spans|
    carapace ...
}

let external_completer = {|spans|
    {
        git: $fish_completer
    } | default $spans.0 | do $in $spans
}
```

#### Case 1: Complete git
- `$spans`: `[git check]`
- `$spans.0`: `git`
- Output of `{...} | get -i $spans.0`: `$fish_completer`
- Output of `default $spans.0`: `$fish_completer` (no changes)
- Equivalent call: `do $fish_completer $spans`

#### Case 2: Complete cargo
- `$spans`: `[cargo  install]`
- `$spans.0`: `cargo`
- Output of `{...} | get -i $spans.0`: `null`
- Output of `default $carapace_completer`: `$carapace_completer` (default value, as previous one is `null`)
- Equivalent call: `do $carapace_completer $spans`

---

If you want to speak more comfortably about this I'll be happy to chat on Discord if you have more questions ✨ I _know_ this works because it's what I use on a daily basis, but I'm probably explaining it badly 😅

fdncred: Great explanation @JoaquinTrinanes! This explanation is more clear than any documentation or Discord conversations I've seen on this subject for nushell. Would love you to add to our docs once this PR gets sorted out. ❤️

amtoine: yeeeah you removed the `$spans.0` part in the record definition, i was so confused i must have forgotten about that part :laughing: 

i think we agree on all of this in the end, thanks @JoaquinTrinanes :ok_hand:

amtoine: this snippet has been an issue for a very long time and always confused me :eyes: 

as @fdncred suggested, maybe we could add a tiny example next to the snippet to make it clear what it does?
otherwise, with all this, i think we're good to go, you tell me if you want to that in this very PR or in another @JoaquinTrinanes :wink:

JoaquinTrinanes: Thank you @fdncred and @amtoine ! 💚 

I'd love to keep helping after this 😁 

What kind of example would you like to be added @amtoine ?  Something explaining what the command does or how to use it in the config and with other completers? 🤔

JoaquinTrinanes:
Ah, another day another issue with the completers :)

I was testing out `atuin` and noticed that I wasn't getting any completion for subcommands _unless_ I started writing it
```nushell
> atuin <TAB> # nothing, list files
> atuin h<TAB> # autocompletes 'help' and 'history'
```

The cause for this is `from tsv`. This is what the fish completions look like (`bat -A` shows hidden characters):
image

And this is what happens when trying to parse it:
image

That's why the `--flexible` flag exists, but instead of detecting 2 columns it only outputs the first one:
image

I don't know if this is a bug or a feature 😅 but the problem can be circunvented, for example, appending a tab at the end of each line
image

So, what do you think about this? Should I just leave it as-is? Open an issue regarding `from tsv`? Add the workaround to the docs?

book/custom_completions.md
amtoine: > I'd love to keep helping after this grin
> 
and we love contributions :wink: 

> What kind of example would you like to be added @amtoine ? Something explaining what the command does or how to use it in the config and with other completers? thinking
>
maybe yes?
we must just make sure it won't clutter the completer section, so at least it should be in a dimmed note, something like that :thinking:

amtoine:
@JoaquinTrinanes 
i've never used TSV and i never use tabs, so i probably won't be super helpful here :eyes:

fdncred:
I'd probably leave `to tsv` alone. I like your alternate solution best `$"value(char tab)description(char newline)" + $out | from tsv --flexible`

JoaquinTrinanes:
> I'd probably leave `to tsv` alone. I like your alternate solution best `$"value(char tab)description(char newline)" + $out | from tsv --flexible`

Pushed the fix in 7b0504ad3acfb76f5c5308d1795f5b00a5d12ff4! @fdncred 

FYI, I'll be away until next week so I'll be unable to update the PR with examples until then. We can merge the PR now (I'd create a new one later) or I'll just add them here in a couple of days, whatever you prefer 😋 

>  i've never used TSV and i never use tabs, so i probably won't be super helpful here 👀

I never use tabs... by choice 🥲

fdncred:
Thanks. When you're available I'd like to have some of the notes here as documentation somewhere either as comments or in our docs site.

PR #960, opened 2023-06-23
Fix direnv not using transformed values

JoaquinTrinanes:
When using `direnv` like the docs indicated I noticed that I no longer had completions for binaries in the `PATH`. Turns out, `direnv` outputs the variables to load as a string (which is normal) and they're being loaded as-is, without undergoing the transformation specified in `$env.ENV_CONVERSIONS`.

The code I changed gets the subset of variables that need to undergo transformation, applies the transformation and merges them with the original values.

I'm not 100% happy with the implementation as it's not that idiomatic, but I couldn't find an easy way to conditionally update a record and returning a new record in the process. I tried the following approaches:

```nushell
# Returns a table with values only in the diagonal
$direnv | items {|key, value| ... { $key: $transformed_value } }
```

```nushell
# Returns a record that maps the index on the table to 1-item record
$direnv | items {|key, value| ... { $key: $transformed_value } } | into record
```

```nushell
# This works! However, it's not very intuitive to read
$direnv | items {|key, value| ... { $key: $transformed_value } } | into record | flatten | into record

# a fully functioning one-liner would look like this
$direnv
    | items {|key,value| { $key: (if $key in ($env.ENV_CONVERSIONS | columns) { do ($env.ENV_CONVERSIONS | get $key | get from_string) $value } else { $value } ) } }
    | into record
    | flatten
    | into record
```

While at it, I also took the chance to change indentation to 4 spaces (the other files I checked used it) and simplify the default value instead of accessing the length of the record.

I'm open to any changes to this, curious to see what other way there might be to achieve this in an easy and readable way! 😁

### DIFF
--- a/cookbook/direnv.md
+++ b/cookbook/direnv.md
@@ -19,6 +19,12 @@ $env.config = {
     pre_prompt: [{ ||
         let direnv = (direnv export json | from json | default {})
         let env_to_convert = ($direnv | transpose key value | where key in ($env.ENV_CONVERSIONS | columns))
+        let converted_values = ($env_to_convert | each {|it|
+            let convert = ($env.ENV_CONVERSIONS | get $it.key | get from_string)
+            let value = (do $convert $it.value)
+            { $it.key: $value }
+        } | into record)
+        $direnv | merge $converted_values | load-env
     }]
   }
 }
--- a/cookbook/direnv.md
+++ b/cookbook/direnv.md
@@ -18,13 +18,20 @@ $env.config = {
   hooks: {
     pre_prompt: [{ ||
         let direnv = (direnv export json | from json | default {})
-        let env_to_convert = ($direnv | transpose key value | where key in $env.ENV_CONVERSIONS)
-        let converted_values = ($env_to_convert | each {|it|
-            let convert = ($env.ENV_CONVERSIONS | get $it.key | get from_string)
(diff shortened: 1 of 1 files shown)


fdncred:
I don't use `direnv` but when I saw your diagonal comment it made me think of `transpose -ird`. Maybe something like this would work `$direnv | items {|key, value| ... { $key: $transformed_value } } | transpose -ird`?

JoaquinTrinanes:
@fdncred THANK YOU! That worked 😁 

I had to build the table differently, creating a `key` and `value` column instead of a list of records like I did before 😁 

I had to add an `is-empty` check because this expression returns `[] transpose -ird` returns an empty list instead of an empty record, which is invalid input for `load-env`. I'll remember this trick in the future, it's not the first time I wanted to do a transformation similar to this one ✨

fdncred:
Thanks

lcnbr:
I'm getting the following error with this change:

```
 329 │         } | transpose -ird | load-env
     ·             ────┬────
     ·                 ╰── command doesn't support list<string> input
 330 │     }]
```
Any idea why?

fdncred:
@lcnbr I can think of 2 things that may cause this.
1. The script is written for 0.85.0 or 0.85.1 of nushell and you're using an older version
2. The script was written for an older version and you're using a newer version of nushell

PR #961, opened 2023-06-25
Remove reference to `build-string` from `working_with_strings.md`

eopb:
`build-string` was removed in https://github.com/nushell/nushell/pull/7144


sholderbach:
Thanks! Well spotted

fdncred:
Thanks! I wonder if we should make a build-string for the stdlib... 🤔

PR #966, opened 2023-06-28
Release notes for `0.83`

amtoine:
Please add your new features and breaking changes to the release notes
by opening PRs against the `release-notes-0.83` branch.

- [x] look at interesting contributions
- [x] write all the sections
- [x] order the sections by interest
- [x] add the breaking changes
- [x] detail the breaking changes
- [x] add the full changelog

### DIFF
--- a/blog/2023-07-25-nushell_0_83.md
+++ b/blog/2023-07-25-nushell_0_83.md
@@ -43,10 +43,167 @@ As part of this release, we also publish a set of optional plugins you can insta
 - [#9587](https://github.com/nushell/nushell/pull/9587) disallow blocks as first-class values
 
 # Full changelog
-<!-- TODO
-    paste the output of
-    ```nu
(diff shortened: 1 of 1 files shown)


amtoine:
i had a look at all the PRs since last release and added in raw tables and sections those that might be of some interest in 996fa120 :relieved:

sholderbach:
Thanks for collecting the current state, I don't think the tables are a particularly helpful presentation to the users so we should change them into more explicit explanations.

amtoine:
> Thanks for collecting the current state, I don't think the tables are a particularly helpful presentation to the users so we should change them into more explicit explanations.
>
yes, of course!

i should have added that this is a temporary document, just listing what is interesting imo :relieved:

amtoine:
same goes with the breaking changes, they are raw right now and should be explained a bit more :wink:

PR #968, opened 2023-06-30
Added useful example to `http options` documentation

bgmort:
This PR adds a more real-world example of how the new `http options` is used, as well as some additional information in the description, including a link to the MDN CORS page.


sholderbach:
Thank you for trying to improve nushell's documentation.

As the source of our command help is the command implementation itself (that gets converted to pages for the commands via a script after the release), please make these changes on the command directly: https://github.com/nushell/nushell/blob/main/crates/nu-command/src/network/http/options.rs

PR #972, opened 2023-07-01
replace `let-env FOO = ...` by `$env.FOO = ...`

amtoine:
related to nushell/nushell#9574

this PR moves from using `let-env FOO = ...` to using `$env.FOO = ...` in all the website.

## procedure
this was an automatic commit, i've simply run the following command :+1: 
```nushell
sd --string-mode "let-env " '$env.' book/**/*
```

cc/ @jntrnr

### DIFF
--- a/book/3rdpartyprompts.md
+++ b/book/3rdpartyprompts.md
@@ -21,7 +21,7 @@ If you like [oh-my-posh](https://ohmyposh.dev/), you can use oh-my-posh with Nus
 3. Set the PROMPT_COMMAND in ~/.config/nushell/config.nu(or the path output by `$nu.config-path`), change `M365Princess.omp.json` to whatever you like [Themes demo](https://ohmyposh.dev/docs/themes).
 
 ```shell
-> let-env PROMPT_COMMAND = { oh-my-posh --config ~/.poshthemes/M365Princess.omp.json }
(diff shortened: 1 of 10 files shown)


fdncred:
I'm kind of on the fence about landing this now because it's such a huge breaking release and we haven't published a release yet that contains these changes. I could go either way.

amtoine:
> I'm kind of on the fence about landing this now because it's such a huge breaking release and we haven't published a release yet that contains these changes. I could go either way.
>
it can wait for the release, at least the change is done here and can be used anytime :relieved: 

on the other hand, changing from `let-env FOO = ...` to `$env.FOO = ...` on the website is not a breaking change right now because both syntaxes are supported, right?

fdncred:
I'd call it a breaking change because anyone who previously used it will now be broken.

amtoine:
but now, `let-env` has been removed, so this is a breaking change for the greater good :vulcan_salute: :wink:

fdncred:
it's a breaking change for the same reaons this https://github.com/nushell/nushell/pull/9574 is a breaking change

amtoine:
do you mean that if we change this *now*, people using the latest stable release will have a broken doc?

if so, yeah, we need to wait for the next release :+1:

jntrnr:
@fdncred they would not have broken docs because `$env.FOO = ...` has been supported for a few releases.

fdncred:
@jntrnr removing `let-env` was the breaking change I was referring to.

PR #973, opened 2023-07-02
add a more complete example to persist aliases in `config.nu`

amtoine:
should solve https://github.com/nushell/nushell/discussions/9586 and probably other issues as well

this PR makes it clear how to add an alias to the persistent config in `config.nu`.


fdncred:
Thanks!

PR #974, opened 2023-07-04
Remove `&&` and `||`

Hofer-Julian:
They have been replaced with `and` and `or`


amtoine (approved):
great catch, thanks @Hofer-Julian :relieved:

PR #975, opened 2023-07-07
fix a broken link in plugins page in zh-CN

Equationzhao:
image

fix a broken link in https://www.nushell.sh/zh-CN/book/plugins.html#%E7%A4%BA%E4%BE%8B


sholderbach:
Thank you very much!

PR #976, opened 2023-07-09
remove all `| table` from the book

amtoine:
related to 
- ~~https://github.com/nushell/nushell/pull/9633~~
- https://github.com/nushell/nushell/pull/9676
- https://github.com/nushell/nushell/issues/9630

with the `display_output` out of the way by default in https://github.com/nushell/nushell/pull/9633, we do not need anymore to pass big tables in `table` to make sure the hook does not trigger `table -e`

Commands used:
```nushell
sd ' ?\| table$' '' **/*
git restore commands/
```


book/pipelines.md
```diff
@@ -79,7 +79,7 @@ In effect, the command:
 And the pipeline:
 
 ```
-> ls | table
+> ls
```
jntrnr, line 82: please don't change this one as it explains how `| table` is added for you automatically.

amtoine: very good catch, thanks @jntrnr :scream: 

should be fixed in f3fd7eba :+1:

amtoine: i've also added a note about `ls` and `ls | table` being *one and the same* in 17597f96, this is not true and caused more confusion in https://github.com/nushell/nushell/issues/9630 :thinking:

amtoine:
now that nushell/nushell#9676 has landed, i think this is ready :yum:

Issue #977, opened 2023-07-10
node hook fails


amtoine:
i was writing and commiting to #976 and i stumbled upon this annowing error when running `git commit`

```shell
npm WARN config tmp This setting is no longer used.  npm stores temporary files in a special
npm WARN config location in the cache, and they are managed by
npm WARN config     [`cacache`](http://npm.im/cacache).
(node:32596) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/amtoine/documents/repos/github.com/amtoine/nushell.github.io/node_modules/.bin/lint-staged:3
import fs from 'node:fs'
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1175:20)
    at Module._compile (node:internal/modules/cjs/loader:1219:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1309:10)
    at Module.load (node:internal/modules/cjs/loader:1113:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
    at node:internal/main/run_main_module:23:47

Node.js v20.3.0
husky - pre-commit hook exited with code 1 (error)
```

## reproduce
- checkout on the latest `main`, e.g. e6ff7836
- run `git commit --allow-empty`

hustcer:
You can add `-n` to skip the hook temporarily，I will check it later

hustcer:
@amtoine You can give this PR https://github.com/nushell/nushell.github.io/pull/978 a try

PR #978, opened 2023-07-11
Use lefthook instead of husky and lint-staged for git hooks

hustcer:
Do the same thing, with less dependencies


hustcer:
> @hustcer, so this should close #977, right?

Yeap

PR #982, opened 2023-07-17
refactor `ssh-agent` cookbook example

amtoine:
this PR hopefully makes the `ssh-agent` example in the cookbook easier to read :innocent: 

i've also removed the note about `ssh-add` as it does not appear to be useful at all :thinking: 
i use this `ssh-agent` everyday without ever running `ssh-add` :confused:

### DIFF
--- a/cookbook/misc.md
+++ b/cookbook/misc.md
@@ -17,8 +17,6 @@ ssh-agent -c
 > **Note**  
 > `transpose -i -r -d` can also be replaced with `transpose --header-row | into record`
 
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Seems to be good, (man are those `transpose` options confusing for my taste)

amtoine:
> man are those `transpose` options confusing for my taste
>
yeah i agree :laughing: 
i always have to play with the combinations of `-r`, `-i` and `-d` to get what i want :eyes:

PR #986, opened 2023-07-18
Update nushell_map_imperative.md

Heidar-An:
Some changes to the nushell imperative map, _maybe I will try to finish it when i'm procrastinating again_

### DIFF
--- a/book/nushell_map_imperative.md
+++ b/book/nushell_map_imperative.md
@@ -10,7 +10,7 @@ Note: this table assumes Nu 0.43 or later.
 | math avg     | statistics.mean               |                                                     |                         |                                               |
 | calc, = math | math operators                | math operators                                      | math operators          | math operators                                |
 | count        | len                           | size, length                                        | length                  | len                                           |
-| cp           | shutil.copy                   |                                                     |                         |                                               |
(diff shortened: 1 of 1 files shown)


book/nushell_map_imperative.md
```diff
@@ -24,37 +24,37 @@ Note: this table assumes Nu 0.43 or later.
...
+| pwd          | os.getcwd                     |                                                     |                         | env::current_dir                              |
 | range        | range                         | .., until, downTo, step                             | iota                    | ..                                            |
 | reduce       | functools.reduce              | reduce                                              | reduce                  | fold, rfold, scan                             |
 | reject       | del                           |                                                     |                         |                                               |
-| rename       | shutil.move                   |                                                     |                         |                                               |
+| rename       | shutil.move                   |                                                     |                         | fs::rename                                    |
```
sholderbach, line 51: This is a mistake on the python side already. `rename` doesn't affect the file system but replaces the keys of a table/records

Heidar-An: oh right, yeah, i just looked at the python command. i'll change both now, should I rename keys in hashmaps for both?

Heidar-An:
can we merge now?

PR #992, opened 2023-07-29
tweak `overlay new` tip

gregokent:
In the tip for `Overlays are Recordable` the text shows how to create a scratch overlay to record changes into before showing a less verbose version using `overlay new`.

However the more verbose version also uses `overlay new` before it's introduced, which I presume is meant to be `overlay use` and this change updates it for the English and German translation. The Chinese translation appears to be what I believe to be the correct usage and the other translations don't seem to have that section yet and have been left alone.

### DIFF
--- a/book/overlays.md
+++ b/book/overlays.md
@@ -135,7 +135,7 @@ The solution can be to create a new empty overlay that would be used just for re
 
 (spam)> module scratchpad { }
 
-(spam)> overlay new scratchpad
+(spam)> overlay use scratchpad
 
 (scratchpad)> def eggs [] { "eggs" }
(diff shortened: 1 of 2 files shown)


fdncred:
I'm not following. `overlay new` seems to be a valid command. are you saying `overlay new` doesn't work or just that it should be `overlay use` in these two places?

gregokent:
The latter. I'm just saying that in that one specific spot (in both languages) it should be `overlay use` as `overlay new` doesn't get introduced as a concept until just after the more verbose example.

fdncred:
ok, thanks!

PR #993, opened 2023-07-29
fix(docs): Link to Command Reference led to HTTP 404

LeoniePhiline:
Hi y'all!

Not sure if this is right. I just noticed that the "Command reference" link at the bottom of  is broken.

I tried fixing it - see diff.

However, there seem to be more places, like in the [table of contents](https://github.com/nushell/nushell.github.io/blob/main/book/table_of_contents.md?plain=1#L40). (Edit: table of contents now also fixed in this PR.)

See also [these search results](https://github.com/search?q=repo%3Anushell%2Fnushell.github.io%20command_reference.md&type=code).

Apparently, in some language versions (japanese, brazilian portugese, german, chinese), `/book/command_reference.md` [exists](https://github.com/nushell/nushell.github.io/blob/main/de/book/command_reference.md?plain=1), while in the default (english) version it [does not exist](https://github.com/nushell/nushell.github.io/tree/main/book) and seems to be placed in [`/commands/README.md`](https://github.com/nushell/nushell.github.io/blob/main/commands/README.md?plain=1) now.

I am confused. :)


hustcer:
Good catch, thanks

PR #994, opened 2023-07-30
patch: release notes for 0.83.1

amtoine:
related to https://github.com/nushell/nushell/issues/9812

this is a small blog post / release note for the patch release following 0.83.0 and whose issues are tracked in https://github.com/nushell/nushell/issues/9812 :+1: 

## :warning: TODO
finish the section about `let-env` not being brought back to life :eyes:


fdncred:
Thanks for starting notes.

I think we'll be naming it 0.83.1 right? since .1 hasn't been released yet. Also, I think it would be good to run your script that generates the list of new PRs for a release, kind of making it like a normal release but we just call it a hotfix/breakfix/patch.

PR #997, opened 2023-07-31
Update testing examples

oatovar:
This PR updates the examples in the testing section so that they function. With the recent updates to testing, the examples were no longer working. Primarily, they were still recommending the usage of `test_<name>` to recognize tests instead of using the test annotations. Also, the `use` directives were no longer accurate, and could also be condensed. I've updated the example so that it could be copied and run successfully.


amtoine (approved):
this looks much better, thanks for updating this long out-of-date documentation :pray:

PR #1001, opened 2023-08-06
Cookbook cleanup

rgwood:
We were chatting about the cookbook on Discord and it really needs some TLC; there's a lot of overlap with the Book and many things are out of date. I want to hide the cookbook entirely and move useful content to blog posts, but that's contentious. I don't have the energy to fight that battle right now so here are some small tweaks instead.

1. Remove an out-of-date example for copying content to the clipboard
2. Simplify the `help` page; I don't think we should be showing users a really long command that would probably not be used in practice (`help commands | where command_type != builtin | first 10 | drop column 2`). Also, I don't think that using `help` for a subcommand is different enough that it needs its own example.
3. Delete the "Native Shell Programs" page, [that's covered in the Book](http://www.nushell.sh/book/escaping.html)
4. Rename the Miscellaneous page to "ssh-agent", the `ssh-agent` content is the most important content on that page
5. Clean up the System page, it incorrectly says that `kill` is not a Nu built-in

The cookbook still needs a lot of love, I haven't reviewed every page.


cookbook/system.md
```diff
@@ -121,34 +121,5 @@ Output
 
 Notes:
 
-- `kill` is Linux/Unix specific command, it is not built-in to nu.
-- filtering with the `where` command, as shown above, is case-sensitive.
-
----
-
-### Pipeline content to clipboard
```
amtoine, line 129: what about mentioning `std clip` instead?

amtoine (commented):
i think these are good changes and a great step forward, thanks @rgwood :blush: 

also the changelog is super great and it was dead easy to review :pray:

cookbook/system.md
amtoine: might better be in the [*Standard library* section](https://www.nushell.sh/book/standard_library.html) :thinking:

rgwood: I don't know how to document stdlib functionality properly, because I don’t really understand the stdlib. I’ll leave that to someone else for now

PR #1002, opened 2023-08-10
Specify alternative to deprecated `str_collect`

AntoniosBarotsis:
Added suggestion for using [`str join`](https://www.nushell.sh/commands/docs/str_join.html) instead of [`str collect`](https://www.nushell.sh/commands/docs/str_collect.html).

Wrote this in the same style as the [`let-env`](https://www.nushell.sh/commands/docs/let-env.html) deprecation comment.

### DIFF
--- a/commands/docs/str_collect.md
+++ b/commands/docs/str_collect.md
@@ -4,7 +4,7 @@ categories: |
   deprecated
 version: 0.83.0
 deprecated: |
-  Deprecated command.
+  Deprecated command. Use `str join` instead.
 usage: |
   Deprecated command.
 ---


AntoniosBarotsis:
Thanks for the quick reply! Should I make this change to the main repo somewhere then? I was just now bumping my shell's version and some of my theme's broke so it would be useful to have the "use this instead" hint :)

amtoine:
`str collect` has been removed completely recently, i think this page should disappear with the next release when we regenerate the website pages :yum:

AntoniosBarotsis:
Wouldn't this be somewhat confusing for anyone that was previously using it though 🤔

fdncred:
@AntoniosBarotsis Considering it was deprecated September 11, 2022, we think it's ok to remove this command now. https://github.com/nushell/nushell/pull/6531

amtoine:
> Wouldn't this be somewhat confusing for anyone that was previously using it though thinking
>
these commands have been removed in https://github.com/nushell/nushell/pull/9840, so after the last *stable* release.
most people will probably use this release and not the latest revision of Nushell and building themselves, thus they will still see this page :ok_hand: 

i'm personaly building from source but then i know using the website might have some inconsistencies with the latest Nushell :yum: 

i hope this answers your concerns :relieved:

PR #999, opened 2023-08-04
remove last mentions to `let-env`

amtoine:
this PR
- replaces the `let-env` with `$env.` in the foreign languages
- fixes the English "configuration" page

## TODO
there are still mentions to `let-env` in the following files
- `zh-CN/book/configuration.md`
- `zh-CN/book/modules.md`
- `zh-CN/book/environment.md`
- `de/book/konfiguration.md`
- `de/book/environment.md`

PR #1004, opened 2023-08-14
Finish `let-env` removal in German translation

sholderbach:
Follow-up to #999


amtoine (commented):
my german is a bit rusty, i'll trust you on this one :wink:

PR #1006, opened 2023-08-15
Update command to list aliases

conqp:
https://github.com/nushell/nushell/issues/9011#issuecomment-1679486217


fdncred:
You could also use `scope aliases`

PR #1009, opened 2023-08-21
Some improvements to `make_docs.nu`

hustcer:
Try to add the following features to `make_docs.nu` for docs generation:

- [x] Add subcommands list to the docs of the root command
- [x] Add subcommand type: Custom, Plugin or Builtin
- [x] Add Input/output types for each command
- [x] Add build tips for commands from `dataframe` feature
- [x] Add build tips for commands from `extra` feature

Check the [example docs](https://github.com/nushell/nushell.github.io/pull/1009/files) for more detail

**Note:** 
Only a few command docs were updated here as examples, and all the docs will be regenerated after the next release.

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -193,8 +193,14 @@ $"($example.description)
     # Typically a root command that has sub commands should be one word command
     let one_word_cmd = ($command.name | split row ' ' | length) == 1
     let sub_commands = if $one_word_cmd { scope commands | where name =~ $'^($command.name) ' } else { [] }
+    let type_mapping = {is_builtin: 'Builtin', is_plugin: 'Plugin', is_custom: 'Custom'}
     let sub_commands = if $one_word_cmd and ($sub_commands | length) > 0 {
-        let commands = $sub_commands | select name usage | update name { $"[`($in)`]\(/commands/docs/($in | safe-path).md\)" } | to md --pretty
+        let commands = $sub_commands
+            | select name usage is_builtin is_plugin is_custom
+            | upsert type {|it| $type_mapping | columns | each {|t| if ($it | get $t) { $type_mapping | get $t } } | str join ',' }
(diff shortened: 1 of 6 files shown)


make_docs.nu
```diff
@@ -12,6 +12,21 @@ def safe-path [] {
   $in | str replace --all '\?' '' | str replace --all ' ' '_'
 }
 
+# NOTE: You should have nushell source code directory sit along with the docs' directory to make it work
+# Get commands in `extra` crate by parsing the source code
+def get-extra-cmds [] {
+    if ('../nushell/crates/nu-cmd-extra/src' | path exists) {
+        cd ..
+        glob nushell/crates/nu-cmd-extra/src/**/*.rs
+            | each { $in | open -r | parse -r 'fn name\(&self\) -> &str \{\n\s+"(?P<name>[^"]+)"\n\s+\}' }
```
fdncred, line 21: I was just playing around with this regex and saw that it does not work in Windows. You'd have to have something more like this:

```nushell
glob crates/nu-cmd-extra/src/**/*.rs | 
  each { 
    $in | 
    open -r | 
    str replace '"' '' -a | 
    parse -r 'fn name\(&self\) -> &str \{[\r|\n]\s+(?P<name>.+)[\r|\n]\s+\}'
  } | flatten
```
I think this would work cross-platform but I haven't been able to test it in mac or linux.

Not saying you need to change it if it already works where you run it.

hustcer: Thanks, I will check it tomorrow

hustcer: I have just updated the regex of `parse`, it works on windows and mac.

> I'm kind of wondering if it's better to grab all the commands this way so we don't have to worry about which platform the script was ran on.

`parse` can help to grab all the commands' name, however, to create the docs we need signatures, input/output, examples, categories, etc. I'm afraid it won't work, at least the regexes would be difficult to maintain

PR #1010, opened 2023-08-22
Use '--locked' in cargo install snippet for dataframe feature

jwarlander:
Update documentation to reflect the requirement for the `--locked` flag when using `cargo install` with the `dataframe` feature.

This problem has been mentioned in the following issues, but was only fixed in the scripts in the nushell repository:
- nushell/nushell#10014
- nushell/nushell#10084


hustcer:
Good catch, thanks

PR #1017, opened 2023-08-23
Make $env.PATH setup more orthogonal.

otakutyrant:


amtoine:
@otakutyrant 
i'm not sure i see what this change brings :confused: 
could you explain a bit more in the PR description? :innocent:

otakutyrant:
Converting `$env.PATH` to list and appending new path into the PATH are orthogonal. So I decouple the code so that users can copy `$env.PATH = ($env.PATH | split row (char esep))` to their config directly. And whether copy and modify the following line `$env.PATH = ($env.PATH | append '/some/path')` is up to them.

amtoine:
ooh i see what you mean @otakutyrant, thanks for the clarification :+1: 

i'm not opposed to that change, even though i'm not sure it's that critical :yum: 

i personally do not like these lines to be split...
maybe writing
```nushell
$env.PATH = (
    $env.PATH
    | split row (char esep)
    | append '/some/path'
)
```
would be a good middle-ground?

otakutyrant:
But what you suggests lose orthogonality again. I learned "code orthogonality" from Pragamtic Programmerm, and you could consult the first section of this: https://medium.com/codex/the-pragmatic-programmer-ep-3-a-pragmatic-practice-part-ii-1193c5c3bfac

fdncred:
I don't think we're going in this direction. Thanks for trying to help.

PR #1025, opened 2023-08-25
nu cheat sheet added

lomm28:
Cheat sheet section as per discussion in https://github.com/nushell/nushell/issues/442


package-lock.json
amtoine, line 1: 4746 deletions, is that normal? :scream: :laughing:

amtoine (commented):
thanks for starting to tackle this :pray: 

in order to make this easier to maintain, i'm wondering if we should move these to snippets and have some tests that we can run every now and then :thinking:

fdncred:
Cool! I love the idea of a cheat sheet. Thanks! However, I'm not sure what's going on with the package-lock.json changes.

lomm28:
I believe `package-lock.json` deletions are due to these lines in `package.json`:
```json
{
  ...
  "lefthook": "^1.4.7",
  "prettier": "^3.0.0",
  ...
}
```
it automatically updates these packages to minor versions so `package-lock.json` gets updated correspondingly. `lefthook` is now 1.4.8 and prettier is `3.0.2`. I didn't do anything special only ran `npm install`.

fdncred:
ok, looks like there are some typos to.

amtoine:
the first typo in https://github.com/nushell/nushell.github.io/actions/runs/5978468827/job/16221561337?pr=1025 looks right, but i'm not sure about the others :sweat_smile:

lomm28:
> in order to make this easier to maintain, i'm wondering if we should move these to snippets and have some tests that we can run every now and then 🤔

 @amtoine I can look into this as well... not sure what kind of tests should be added for the snippets though

fdncred:
Let's move forward with this as-is and come back in a separate PR after investigating whether having snippets is necessary. I can understand the reason for wanting them, but it's going to create like 100 little snippet files. I'm not sure that's easier to maintain than one md.

fdncred:
Thanks @lomm28

PR #1026, opened 2023-08-26
Error style section

JoaquinTrinanes:
Related to https://github.com/nushell/nushell/pull/10122, this PR creates an accessibility subsection explaining recommended settings for screen reader users.

### DIFF
--- a/book/coloring_and_theming.md
+++ b/book/coloring_and_theming.md
@@ -496,7 +496,7 @@ if you want to go full-tilt on theming, you'll want to theme all the items I men
 
 ### Working on light background terminal
 
-Nushell's default config file contains a light theme definition, if you are working on a light background terminal, you can applied light theme easily.
+Nushell's default config file contains a light theme definition, if you are working on a light background terminal, you can apply the light theme easily.
 
 ```shell
 # in $nu.config-file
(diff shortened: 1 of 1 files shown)


book/coloring_and_theming.md
```diff
@@ -517,3 +517,20 @@ $env.config = {
   ...
 }
 ```
+
+## Accessibility
+
+It's often desired to have the minimum amount of decorations when using a screen reader. In those cases, it's possible to disable borders and other decorations for both table and errors with the following options:
+
+```shell
+# in $nu.config-file
```
amtoine, line 526:
```suggestion
# in $nu.config-path
```
:wink:

amtoine (commented):
thanks a lot :pray: 

let's wait for nushell/nushell#10122 and then i think we can land this :ok_hand:

book/coloring_and_theming.md
amtoine: ooh there were more `$nu.config-file` in that page :scream: 
thanks for fixing the other ones :pray:

JoaquinTrinanes: No problem! They were the reason I wrote that in the first place 😌

JoaquinTrinanes:
I'm wondering if we should add indications to disable the banner in the accessibility section, to avoid all the noise caused by
```
     __  ,
 .--()°'.' Welcome to Nushell,
'|, . ,'   based on the nu language,
 !_-(_\    where all data is structured!
```

fdncred:
Thanks! Let's leave the banner for now.

PR #1030, opened 2023-08-29
Clarify how to check config file paths

adamchalmers:
I was confused about where Nushell keeps its config files, and I read this page twice trying to figure it out. Somehow I didn't understand that `echo $nu.env-path` was the answer.

Here's my attempt to make the docs a little clearer.

### DIFF
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -4,7 +4,7 @@
 
 Nushell uses a configuration system that loads+runs two Nushell script files at launch time:
 First, `env.nu`, then `config.nu`.
-Paths to these files can be found by calling `echo $nu.env-path` and `echo $nu.config-path`.
+You can check where Nushell is reading these config files from by running `echo $nu.env-path` and `echo $nu.config-path`.
 `env.nu` is meant to define the environment variables which are then available within `config.nu`.
 `config.nu` can be used to add definitions, aliases, and more to the global namespace.
 
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -4,7 +4,7 @@
 
 Nushell uses a configuration system that loads and runs two Nushell script files at launch time:
 
- - `env.nu` is used to define environment variables. These typically get used in the second config file, which is...
+ - `env.nu` is used to define environment variables. These typically get used in the second config file, config.nu.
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Thanks, that makes it both clearer and seems to improve the reading flow!

book/configuration.md
```diff
@@ -4,7 +4,7 @@
 
 Nushell uses a configuration system that loads+runs two Nushell script files at launch time:
 First, `env.nu`, then `config.nu`.
-Paths to these files can be found by calling `echo $nu.env-path` and `echo $nu.config-path`.
+You can check where Nushell is reading these config files from by running `echo $nu.env-path` and `echo $nu.config-path`.
```
sholderbach, line 7: In principle you don't need to use `echo` with Nushell (for the most part it is a no-op)

```suggestion
You can check where Nushell is reading these config files from by running `$nu.env-path` and `$nu.config-path`.
```

adamchalmers: I agree... but I suspect the start of this page is aimed at beginners coming from bash. Should we try to accommodate beginners used to Bash, or get them used to idiomatic Nushell?

adamchalmers: OK, I updated the PR and I think it's more clear now.

book/configuration.md
```diff
@@ -3,10 +3,16 @@
 ## Nushell Configuration with `env.nu` and `config.nu`
 
 Nushell uses a configuration system that loads+runs two Nushell script files at launch time:
-First, `env.nu`, then `config.nu`.
-Paths to these files can be found by calling `echo $nu.env-path` and `echo $nu.config-path`.
-`env.nu` is meant to define the environment variables which are then available within `config.nu`.
-`config.nu` can be used to add definitions, aliases, and more to the global namespace.
+
+ - `env.nu` is used to define environment variables.
```
fdncred, line 7: It's an important distinction here that these environmental variables are available within config.nu. This means that env.nu gets sourced first.

fdncred:
I'm not going to hold back this PR for this comment that I'm about to make, but there are really 3 config files. 
1. `login.nu` which is used when nushell is set as your login shell. 
2. `env.nu` which gets loaded first, 
3. and `config.nu`.

sholderbach:
Let's ship it for now, as this is a definite improvement!

Thanks @adamchalmers and @fdncred

PR #1031, opened 2023-08-30
Link to escape when discussing escapes

adamchalmers:
The book chapter on configuration mentions ^ syntax before the chapter on escapes. I think it'd be helpful to link to that chapter so the reader can learn more if they want to.


fdncred:
Thanks!

PR #1032, opened 2023-08-30
Move launch instructions earlier in page.

follower:
The instructions were in the Package Managers section which meant they were easily missed if someone[0] only read as far as the binary download link.

This is a drive-by contribution which I'm unlikely to follow-up on so feel free to make any changes as you see fit.

(For completeness, it might be worth adding this information to the start of the Quick Tour page too: http://www.nushell.sh/book/quick_tour.html )

[0] Me. I'm the someone. :)


hustcer:
I think this change is reasonable, but I'd also like to hear others' opinions.

fdncred:
no objections

PR #1033, opened 2023-08-31
Update loading_data.md, correct link to http command

narve:
The section on http mentions the http get command but the link points to  (probably deprecreated/renamed) fetch instead of http


hustcer:
Thanks

Issue #1041, opened 2023-09-02
Make generated command pages point to the right place for editing


alsuren:
Unfortunately, any change to commands/docs gets rejected because those docs are generated from the commands themselves with each new release. If you want to change the help for a command like this, you'll need to change the rust code and then when we auto-generate, it'll appear in here.

_Originally posted by @fdncred in https://github.com/nushell/nushell.github.io/issues/1040#issuecomment-1703838807_
            

Currently https://www.nushell.sh/commands/docs/keybindings.html has a link to `Edit this page on GitHub` which points at https://github.com/nushell/nushell.github.io/edit/main/commands/docs/keybindings.md . I followed its instructions and opened https://github.com/nushell/nushell.github.io/issues/1040 but then I had to close it and open https://github.com/nushell/nushell/pull/10193 instead.

Options include:
a) make the generated keybindings.md contain a `<!-- comment -->` pointing at the location in the rust code where the source of truth is (even just pointing at the repo would probably be enough).
b) somehow disable the mdbook-generated `Edit this page on GitHub` link to avoid confusion. (not sure if it's possible to do on just hte commands/docs subtree. I've not looked)

fdncred:
I believe the CI failures are intended to tell people about making links in commands/docs

PR #970, opened 2023-07-01
TEST: try to change a command documentation file

amtoine:
related to https://github.com/nushell/nushell.github.io/pull/969

## :warning: THIS IS A TEST, DO NOT MERGE

PR #1042, opened 2023-09-02
Point out where command docs should be edited

alsuren:
Fixes https://github.com/nushell/nushell.github.io/issues/1041 in the most minimal way possible.

Hopefully this will signpost people towards the right place to contribute before they've gone to the effort of formatting their markdown and making PR descriptions. It should also reduce the impact of the CI check being broken (https://github.com/nushell/nushell.github.io/pull/970).

- [ ] I tried running `nu make_docs.nu` (`npm run update`) but I ran into https://github.com/nushell/nushell/issues/10184 (so I will have to wait until that is fixed and released before trying again I guess?)


sholderbach (approved):
Thanks, sounds like a good addition!

PR #1043, opened 2023-09-04
Fix example history command pipeline

ofek:
```
❯ history | wrap cmd | where cmd =~ cargo
Error: nu::shell::type_mismatch

  × Type mismatch during operation.
   ╭─[entry #23:1:1]
 1 │ history | wrap cmd | where cmd =~ cargo
   · ───┬───                        ─┬ ──┬──
   ·    │                            │   ╰── string
   ·    │                            ╰── type mismatch for operator
   ·    ╰── record<start_timestamp: string, command: string, cwd: string, duration: duration, exit_status: int>
   ╰────
```

### DIFF
--- a/commands/docs/history.md
+++ b/commands/docs/history.md
@@ -45,7 +45,7 @@ Show last 5 commands you have ran
 
 Search all the commands from history that contains 'cargo'
 ```shell
-> history | wrap cmd | where cmd =~ cargo
+> history | wrap record | where record.command =~ cargo | get record.command
 
 ```
 


hustcer:
Thanks, and the docs were generated by `make_docs.nu` and it's not recommend to modify it manually, it's better to fixed it here: https://github.com/nushell/nushell/blob/e5145358eb4a13436eccae58c7e07acc9666b661/crates/nu-cli/src/commands/history.rs#L142
BTW: `wrap` may be  unnecessary here? `history | where command =~ cargo | get command` should works too

fdncred:
Agreed. I don't think `wrap` is required anymore.

ofek:
Thanks for the feedback! I opened https://github.com/nushell/nushell/pull/10220

PR #1045, opened 2023-09-04
add a link to the contributor book at the end of the plugins chapter

stormasm:
The contributor book has a great chapter on how plugins work...

However, when reading the plugins chapter in the nushell book there is no reference to it.

You have to search for it to find it...

I added a link to the end of the plugins chapter so people can read more stuff about plugins...

I found the chapter in the contributor book incredibly insightful.

### DIFF
--- a/book/plugins.md
+++ b/book/plugins.md
@@ -49,3 +49,7 @@ The simplest way to debug a plugin is to print to stderr; plugins' standard erro
 ## Help
 
 Nu's plugin documentation is a work in progress. If you're unsure about something, the #plugins channel on [the Nu Discord](https://discord.gg/NtAbbGn) is a great place to ask questions!
+
+## More details
+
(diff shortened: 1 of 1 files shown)


fdncred:
Yup, someone recently updated that plugins section. I'd like to get the entire contributor book up to snuff at some point.

stormasm:
https://github.com/stormasm/nu_plugin_len/tree/span-refactor

I updated the rust code for the book as well but will not land it until
we do the next release...

stormasm:
> Yup, someone recently updated that plugins section. I'd like to get the entire contributor book up to snuff at some point.

They did a great job !  And after reading that section of the contributor book on plugins I was like "no one will ever find this information"

The contributor book does not even have a TOC so you have to individually go to each section ?

We have lots more fun work to do 😄

PR #1048, opened 2023-09-06
Document that `open` looks up `from`s in scope

sholderbach:
That we check for the extension and then look up the matching `from` subcommand has been an open secret used by plugin authors. Let's document it.


fdncred:
Thanks Stefan for documenting this magic. 🤣

PR #1050, opened 2023-09-09
Update min required node version and some node modules

hustcer:
1. Set min required node version to v18.12.0-LTS, as v16 LTS support will goes to end, see: https://endoflife.date/nodejs
2. Upgrade some node modules
3. Upgrade some outdated Github Actions


fdncred:
Thanks @hustcer for continually working/maintaining our docs repo!!!! ❤️

PR #1054, opened 2023-09-10
Update variables_and_subexpressions.md - remove obsolete section

connorjs:
Replaces unknown `sin` with `sqrt` for the “Variable values” example.

---

I encountered an error in the [Variable values guide](https://www.nushell.sh/book/variables_and_subexpressions.html#variable-values).

```
~> let val = (42 | math sin)                                                                                                                         09/10/2023 01:58:00 PM
Error: nu::parser::extra_positional

  × Extra positional argument.
   ╭─[entry #108:1:1]
 1 │ let val = (42 | math sin)
   ·                      ─┬─
   ·                       ╰── extra positional argument
   ╰────
  help: Usage: math
```

I chose `math sqrt`, but feel free to edit to any math subcommand.

### DIFF
--- a/book/variables_and_subexpressions.md
+++ b/book/variables_and_subexpressions.md
@@ -94,20 +94,6 @@ It is common for some scripts to declare variables that start with `$`. This is 
 # identical to `let var = 42`
 ```
 
-### Variable Values
-
-A common issue that users run into is trying to declare a variable using a pipeline as the value like so:
-
-```nu
-let val = 42 | math sqrt
-```
-
-This is an error, because the pipe command is used to separate pipeline elements, so Nushell will see this as a pipeline with two elements, one of which is not allowed in pipelines because it doesn't return a value (`let val = 42`). The correct way to declare using a pipeline is to wrap the pipeline in parentheses
-
-```nu
-let val = (42 | math sqrt)
-```
-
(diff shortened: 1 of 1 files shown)


sholderbach (changes requested):
Thanks for trying to fix the snippet.

Since a few releases this whole subsection is actually outdated. `let` assignments will accept a pipline even if it is not delimited with parentheses.

fdncred:
I'm not sure what you mean `unknown sin`.
```nushell
❯ 42 | math sin
-0.9165215479156338
```
Maybe part of these docs need to be left in place and just say that `math sin` is available in the extra feature.

sholderbach:
> Maybe part of these docs need to be left in place and just say that `math sin` is available in the extra feature.

The section doesn't make sense anymore so we can't slide in some extra explanations for unrelated stuff here.

connorjs:
I just came across the “extra” features when reading other issues. https://github.com/nushell/nushell/tree/main/crates/nu-cmd-extra mentions these will likely be dropped before 1.0, so I think docs should not use those features. (Although mentioning extras somewhere may be nice, especially for using nu for scripting/Data science/math use cases.)

> Since a few releases this whole subsection is actually outdated.

I will defer to y’all on this then!

PR #1057, opened 2023-09-13
Support for standard input in script with shebang

dnsem:
I found out that `stdin` is not available when running a script with shebang, and after some investigation I found a solution.


fdncred:
Thanks!

PR #1058, opened 2023-09-14
Change direnv example to upload $env.PATH as a list

Gryff:
First of all thanks for the direnv example!
It got me unblocked, but I missed that I couldn't see the path in the list format when I typed `$env.PATH`.

I'm new to nushell, so if there's a better way to do this than the `no_changes` `changes` thing, let me know and I'll change it. I couldn't find like a `$my_thing | is-not-empty`.

### DIFF
--- a/cookbook/direnv.md
+++ b/cookbook/direnv.md
@@ -18,8 +18,13 @@ $env.config = {
   hooks: {
     pre_prompt: [{ ||
       let direnv = (direnv export json | from json)
-      let direnv = if ($direnv | length) == 1 { $direnv } else { {} }
-      $direnv | load-env
+      let no_changes = $direnv | is-empty
+      let changes = not $no_changes
+
+      if $changes {
+        let direnv = $direnv | upsert PATH {|direnv_result| ($direnv_result.PATH | split row ":") }
+        $direnv | load-env
+      }
     }]
   }
 }
--- a/cookbook/direnv.md
+++ b/cookbook/direnv.md
@@ -21,7 +21,7 @@ $env.config = {
       let no_changes = $direnv | is-empty
       let changes = not $no_changes
 
-      if $changes {
+      if not ($direnv | is-empty) {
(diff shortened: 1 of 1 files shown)


cookbook/direnv.md
```diff
@@ -18,8 +18,13 @@ $env.config = {
   hooks: {
     pre_prompt: [{ ||
       let direnv = (direnv export json | from json)
-      let direnv = if ($direnv | length) == 1 { $direnv } else { {} }
-      $direnv | load-env
+      let no_changes = $direnv | is-empty
+      let changes = not $no_changes
+
+      if $changes {
```
fdncred, line 24:
```suggestion
      if not ($direnv | is-empty) {
```
I think the above will work. Then I'd just delete the no_changes and changes lines.

amtoine: sounds simple enough to me that it does not require two intermediate variables, agree with @fdncred :+1:

cookbook/direnv.md
```diff
@@ -18,8 +18,13 @@ $env.config = {
...
+        let direnv = $direnv | upsert PATH {|direnv_result| ($direnv_result.PATH | split row ":") }
```
amtoine, line 25: i would remove the parentheses here and use `it` to have a bit less clutter on the line
```suggestion
        let direnv = $direnv | upsert PATH {|it| $it | split row ":"}
```

Gryff: I agree it looks better, just need to modify `$it` to `$it.PATH`
```suggestion
        let direnv = $direnv | upsert PATH {|it| $it.PATH | split row ":" }
```

fdncred:
I personally don't like things like `upsert PATH {|it| $it.PATH | split row ":" }`. To me, it's not useful to name the closure variable with a generic variable that means nothing. I prefer my variables to have meaning like `upsert PATH {|direnv_result| $direnv_result.PATH | split row ":" }`.

If we're just going for brevity, then you could do `upsert PATH { $in.PATH | split row : }`.

But let's land this anyway.

PR #1060, opened 2023-09-14
Changed Python plugin example to be more portable

edhowland:
…This PR will ssolve Issue 1059. Change Python example in Plugins chapter shebang line to be more portable.


sholderbach:
Thanks @edhowland!

PR #1062, opened 2023-09-18
DE translation for hooks and background_tasks

petrisch:


hustcer:
Thanks, let's land it

PR #1065, opened 2023-09-19
Fix formatting in `de/book/hooks`

sholderbach:


petrisch:
@sholderbach thanks for the fix! 🥳

PR #1067, opened 2023-09-19
Fix Rust example in contributor-book/plugins.md that does not compile in recent Nushell versions

edhowland:
Changes the Rust example in the file:
contributor-book/plugins.md
to compile with recent versions of Nushell.
Issue number 1066      details the history of thechanges that caused the example
in the web page to no longer compile.
This PR modifies './src/main.rs' around lines 16 and 17 to use the new 
'input_output_type' instead of the previous separate 'input_type', and
'output_type' functions in the nu-protocol crate.
Fixes ./src/main.rs in the example to be:

```rust
            .input_output_type(Type::String, Type::Int)
```

Also changes the sample Cargo.toml to pin the 'nu-plugin' and 'nu-protocol'
crates to 0.84.0:

```toml
[dependencies]
nu-plugin = "0.84.0" # These version numbers may differ
nu-protocol = "0.84.0"
```


contributor-book/plugins.md
```diff
@@ -44,8 +44,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-nu-plugin = "0.80.0" # These version numbers may differ
-nu-protocol = "0.80.0"
+nu-plugin = "0.84.0" # These version numbers may differ
+nu-protocol = "0.84.0"
```
amtoine, line 48: maybe we can just wait for tomorrow to bump this to `0.85.0` directly?

edhowland: Bumped nu-plugin and nu-protocol dependant crates in sample Cargo.toml to version 0.85.0 each.

fdncred (approved):
Looks good @edhowland. Thanks.

PR #1074, opened 2023-09-22
Add information on transient prompt

ysthakur:
Since [this PR](https://github.com/nushell/nushell/pull/10391) to add transient prompts to Nushell was merged, this PR adds some docs on how to use the newly added environment variables to the website.


amtoine (approved):
very interesting feature :open_mouth: 

i could follow the steps there and got it working and then off again as before, sounds very good to me, thanks for the complete feature :pray:

PR #1076, opened 2023-09-26
Update external_completers.md

marcelarie:
use `enable` instead of `enabled`


sholderbach:
Thanks for the correction!

PR #1078, opened 2023-09-28
Add real Nushell syntax highlighting support

hustcer:
## Try to Add real Nushell syntax highlighting support to Nushell official site:

It may take a bit longer though let's just get started, the detailed plan:

- [ ] Add Nushell syntax highlighting support to [Shiki](https://github.com/shikijs/shiki/pull/522);
- [ ] Use `Shiki` as `Vuepress`'s syntax highlighter;
- [ ] Update all Nu code blocks, wrap them with <code>\```nushell  \```</code>  or <code>\```nu  \```</code>


fdncred:
Nice work @hustcer. I'll be excited to see this (if 522 ever lands). I appreciate all your attention to our docs. Thank you so much!

fdncred:
What's the deal with shiki? Do they not accept PRs very often?

hustcer:
> What's the deal with shiki? Do they not accept PRs very often?

They may accept it after a while, no one is taking care of the PRs now. And we can't wait for them for the uncertainty, we can add a patch as a workaround.

sholderbach:
Thanks for pushing this forward @hustcer 

Are there special rules in the grammar for how a table or prompt should look like? 

Looking at a random assortment of prompts sometimes it highlights the code after a `> ` and sometimes it doesn't.

hustcer:
@sholderbach I'm simply use [The Same Nu Grammar as the VS Code Plugin Here](https://github.com/nushell/vscode-nushell-lang/blob/main/syntaxes/nushell.tmLanguage.json), maybe @fdncred knows?

fdncred:
It's a regex fest for the grammar. It should behave identically, as far as syntax checking, as the extension. So, you could test by putting a `> pipe | line` in vscode and it should work the same. I'm not sure what that would do? You won't get sqigglys because that comes from nushell integration with the extension.

sholderbach:
The page looks generally fine even when prefixed with the `> `. I think the weird ones I observed were pages that did not yet have the new annotations.

fdncred:
Also, the tables look a bit weird because the column names seem green-ish. Wondering if they should be in their own code fence?

hustcer:
> Also, the tables look a bit weird because the column names seem green-ish. Wondering if they should be in their own code fence?

It's better to update the grammar file to make those strings with `# |` in it is not taken as a comment.

PR #1049, opened 2023-09-07
Simplify external completer

JoaquinTrinanes:
This PR modernizes the external completer docs by using `match` instead of `{ ... } | get -i ... | default ...`.

It's:
- Simpler to read and understand
- A nice way to get folks reading the docs to learn about `match` (<3)
- A textbook example of when to use `match`

PR #1081, opened 2023-09-28
Fix syntax errors

Roger-Roger-debug:
I tried copying the examples but got errors like
```
 163 │     match $spans.0
     ·                   ▲
     ·                   ╰── missing match_block
 164 │     {
     ╰────
```

I just got started using nushell so this might be wrong.

### DIFF
--- a/cookbook/external_completers.md
+++ b/cookbook/external_completers.md
@@ -46,8 +46,8 @@ This completer is not usable for allmost every other command, so it's recommende
 
 ```nu
 {
-    z: $zoxide_completer
-    zi: $zoxide_completer
+    z => $zoxide_completer
+    zi => $zoxide_completer
 }
 ```
 
@@ -57,8 +57,8 @@ This completer is not usable for allmost every other command, so it's recommende
 >
 > ```nu
 > {
->     __zoxide_z: $zoxide_completer
->     __zoxide_zi: $zoxide_completer
+>     __zoxide_z => $zoxide_completer
+>     __zoxide_zi => $zoxide_completer
 > }
 > ```
 
@@ -68,8 +68,7 @@ Sometimes, a single external completer is not flexible enough. Luckily, as many 
 
 ```nu
 let multiple_completers = {|spans|
-    match $spans.0
-    {
+    match $spans.0 {
         ls => $ls_completer
         git => $git_completer
         _ => $default_completer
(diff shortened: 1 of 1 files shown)


cookbook/external_completers.md
```diff
@@ -46,8 +46,8 @@ This completer is not usable for allmost every other command, so it's recommende
 
 ```nu
 {
-    z: $zoxide_completer
-    zi: $zoxide_completer
+    z => $zoxide_completer
+    zi => $zoxide_completer
```
sholderbach, line 50: Mhh this is not valid Nu code outside of a `match`. There may be some work necessary to get this section up to date. Previously, before `match` existed things like this where handled with records for which this is the correct syntax.

cookbook/external_completers.md
```diff
@@ -68,8 +68,7 @@ Sometimes, a single external completer is not flexible enough. Luckily, as many
 
 ```nu
 let multiple_completers = {|spans|
-    match $spans.0
-    {
+    match $spans.0 {
```
sholderbach, line 71: Good catch!

sholderbach (commented):
Thnanks for giving this a good look!

I think the snippets in the sections preceding "Multiple completer" don't make as much sense after #1049?

cookbook/external_completers.md
Roger-Roger-debug: I'm still a bit lost with everything. Just reading the docs it says that "This completer is not usable for allmost every other command, so it's recommended to add it as an override in the [multiple completer](https://www.nushell.sh/cookbook/external_completers.html#multiple-completer):". I copy-pasted it to the appropriate place and got the syntax error.

JoaquinTrinanes: No wonder it doesn't work, those snippets where written when the completer options were a `record | get | default` instead of a `match` :sweat_smile: 

As far as I can tell these changes should be enough, but maybe it should be made clearer where this configuration goes, and emphasize that it's a match statement?

Or just writing it as

```shell
match $spans.0 {
    ...
   __zoxide_z | __zoxide_zi => $zoxide_completer
    ...
}
```

Btw, found a typo at `This completer is not usable for alLmost ...` 

Happy to lend a hand rewriting this if needed :) As far as I can see no other snippets need updating

Roger-Roger-debug: I fixed the typo and added zoxide to the full example, I hope this is what you had in mind.

PR #1083, opened 2023-09-30
Expand `append` examples and add `match` example

savente93:
As discussed in https://github.com/nushell/nushell/issues/10486 Some more examples of the various ways to append things to lists/strings/tables would be appreciated. In addition, `str append` was a nice exmaple of how to use `match` and also pattern match on types. 

This PR adds more examples on the appending in the appropriate sections (specifically more on using the ++ operator which was previously not mentioned).

`str append` is also added to the cookbook and expanded upon somewhat. In the linked issue "vectorisation" was also mentioned but I wasn't quite sure how/what part of the `str append` that refers to so I didn't mention that. If someone can explain that part to me and thinks it would be useful to also add, I'd be happy to do so. 

Finally I think instructions on how to add new pages to the cookbook were outdated (at least as far as I could tell) so I updated that as well in accordance with how I did it. If this is incorrect, Let me know how it should be done instead and I'll happily change it up

### DIFF
--- a/book/working_with_strings.md
+++ b/book/working_with_strings.md
@@ -134,6 +134,39 @@ You can place the `^` sigil in front of any string (including a variable) to hav
 
 You can also use the [`run-external`](/commands/docs/run-external.md) command for this purpose, which provides additional flags and options.
 
+## Appending and Prepending to strings
+
+There are various ways to pre, or append strings. If you want to add something to teh beginning of each string closures are a good option:
+
+```sh
+ ['foo', 'bar'] | each {|s| '~/' ++ $s} # ~/foo, ~/bar
+['foo', 'bar'] | each {|s| '~/' + $s} # ~/foo, ~/bar
+```
+
+You can also use a regex to replace the beginning or end of a string:
+
+```sh
+['foo', 'bar'] | str replace -r '^' '~/'# ~/foo, ~/bar
+['foo', 'bar'] | str replace -r '$' '~/'# foo~/, bar~/
+```
+
+If you want to get one string out of the end then `str join` is your friend:
+
+```sh
+"hello" | append "world!" | str join " " # hello world!
+```
+
+You can also use reduce:
+
+````sh
+1..10 | reduce -f "" {|it, acc| $acc + ($it | into string) + " + "} # 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10 +
+``
+
+
+though in the cases of strings, especially if you don't have to opperate on the strings, it's usually easier and more correct (notice the extra + at the end in the example above) to use `str join`.
+
+Finally you could also use string interpolation, but that is complex enough that it is covered in it's own subsection below.
+
 ## String interpolation
 
 More complex string use cases also need a new form of string: string interpolation. This is a way of building text from both raw text and the result of running expressions. String interpolation combines the results together, giving you a new string.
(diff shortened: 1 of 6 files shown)


savente93:
@amtoine Would you mind reviewing? (you mentioned in the issue I could ping you)

fdncred (approved):
Very nice improvements to our documentation. Thanks!

fdncred:
I fixed a couple typos. Good work!

book/working_with_strings.md
```diff
@@ -146,7 +179,7 @@ For example, let's say we have a variable called `$name` and we want to greet th
 > let name = "Alice"
 > $"greetings, ($name)"
 greetings, Alice
-```
+````
```
amtoine, line 182: this looks incorrect, right?

cookbook/pattern_matching.md
```diff
@@ -0,0 +1,86 @@
...
+  } else if ( $c in ['yellow', "purple"]){
+   "vibrant"
+  }  else {
+   "innovative"
+  }
+ }
+}
+```
+
+As you can see you can also use comound expressions in match statements (in this case used with `|`). Also notice the `_` case at the end, this is called the default arm and is used.
```
amtoine, line 48: did you mean *command* instead of *comound*? :yum:

cookbook/pattern_matching.md
```diff
@@ -0,0 +1,86 @@
...
+
+```sh
+{one: 1 two: 2} | describe
+record<one: int, two: int>
+
+[{a: 1 b: 2} {a: 2 b:3 }] | describe
+table<a: int, b: int>
+```
+
+Together with `match` and some clever regex use you can do quite powerful type matching. For example, let's say we wanted to implement a `str append` function that would work on both strings and lists. On strings it would work as expected, on lists of strings, it should append the same string to each list. Using `match` one might do that like so:
```
amtoine, line 62: maybe? :innocent: 
```suggestion
Together with `match` and some clever regex use you can do quite powerful type matching. For example, let's say we wanted to implement a `str append` function that would work on both strings and lists. On strings it would work as expected, on lists of strings, it should append the same string to each element of the list. Using `match` one might do that like so:
```

amtoine (changes requested):
i like that a lot :pray: 

i have a few minor remarks and, if you fix the CI for typos, i think we'll be able to land this soon :blush:

savente93:
Thanks for the comments both! should be all good now :blush:

cookbook/pattern_matching.md
amtoine: done in https://github.com/nushell/nushell.github.io/pull/1083/commits/0bad6309b64e456bc89f194c660d1d8899961839

book/working_with_strings.md
amtoine: done in https://github.com/nushell/nushell.github.io/pull/1083/commits/32067f1dd7f61b1f008e54b401f2991ff9641f76

cookbook/pattern_matching.md
amtoine: https://github.com/nushell/nushell.github.io/pull/1083/commits/32067f1dd7f61b1f008e54b401f2991ff9641f76

amtoine (approved):
thanks for the quick fixes and making our documentation better, much appreciated :pray: 

i don't see anything more to change, let's land this :partying_face:
in case there is a typo or something off, we can fix them later :relieved:

PR #1096, opened 2023-10-04
Improve command docs

Hofer-Julian:
Get closer to `help command` in some aspects, which I think is more accessible in some aspects.
Also, now the short option is included for named arguments.

Before:
before

After:
after

Comparison to `help bytes add`:

```
nu ❯ help bytes add
Add specified bytes to the input.

Search terms: append, truncate, padding

Usage:
  > bytes add {flags} <data> ...(rest)

Flags:
  -i, --index <Int> - index to insert binary data
  -e, --end - add to the end of binary

Parameters:
  data <binary>: the binary to add
  ...rest <cell-path>: for a data structure input, add bytes to the data at the given cell paths

Input/output types:
  ╭───┬──────────────┬──────────────╮
  │ # │    input     │    output    │
  ├───┼──────────────┼──────────────┤
  │ 0 │ binary       │ binary       │
  │ 1 │ list<binary> │ list<binary> │
  │ 2 │ table        │ table        │
  │ 3 │ record       │ record       │
  ╰───┴──────────────┴──────────────╯
```


fdncred:
You'd think our docs would be nearly identical to our help output but I guess not. Looks like an improvement to me. Thanks!

hustcer:
Looks better indeed, let's land it. Thanks

PR #1097, opened 2023-10-05
fixed alias normalization

dedebenui:
changed `split words` to `split row " "` because `split words` gets rid of too much. For example, `"-foo__bar baz" | split words` outputs the list `[foo, bar, daz]` while `"-foo__bar baz" | split row " "` outputs `[-foo__bar, baz]`, which is what we want

This was an issue as I was trying to setup completion for zoxide and this code was splitting `__zoxide_z` into `[zoxide, z]`

### DIFF
--- a/cookbook/external_completers.md
+++ b/cookbook/external_completers.md
@@ -96,7 +96,7 @@ let expanded_alias = (scope aliases | where name == $spans.0 | get -i 0 | get -i
 # overwrite
 let spans = (if $expanded_alias != null  {
     # put the first word of the expanded alias first in the span
-    $spans | skip 1 | prepend ($expanded_alias | split words)
+    $spans | skip 1 | prepend ($expanded_alias | split row " ")
 } else { $spans })
 ```
 


fdncred:
Thanks for figuring this out. I created `split words` for a very specific use case, which was to do this one particular test easier.
```
// curl -sL "https://www.gutenberg.org/files/11/11-0.txt" | tr '[:upper:]' '[:lower:]' | grep -oE "[a-z\']{1,}" | ^sort | ^uniq -c | ^sort -nr | head -n 10
// benchmark INCLUDING DOWNLOAD: 1sec 253ms 91µs 511ns
//    1839 the
//     942 and
//     811 to
//     695 a
//     638 of
//     610 it
//     553 she
//     546 i
//     486 you
//     462 said
```
with nushell
```
// current
// $alice | str downcase | split words | uniq -c | sort-by count --reverse | first 10
// benchmark: 1sec 481ms 604µs 700ns
// ╭───┬───────┬───────╮
// │ # │ value │ count │
// ├───┼───────┼───────┤
// │ 0 │ the   │  1839 │
// │ 1 │ and   │   942 │
// │ 2 │ to    │   811 │
// │ 3 │ a     │   695 │
// │ 4 │ of    │   638 │
// │ 5 │ it    │   610 │
// │ 6 │ she   │   553 │
// │ 7 │ i     │   546 │
// │ 8 │ you   │   486 │
// │ 9 │ said  │   462 │
// ├───┼───────┼───────┤
// │ # │ value │ count │
// ╰───┴───────┴───────╯
```

If the commented out parameters were added, it may make it work in more general cases.

https://github.com/nushell/nushell/blob/afdb68dc7175b1c6b5f4ca8f6a20b55036cb2375/crates/nu-command/src/strings/split/words.rs#L30-L44

Issue #460, opened 2022-05-29
Inconsistancies in documentation


petrisch:
When dealing with the [german](https://github.com/petrisch/nushell.github.io/blob/main/de/book/thinking_in_nushell.md) translation, I found a few things that are inconsistent even in the english version.

in [types_of_data](https://github.com/nushell/nushell.github.io/blob/main/book/types_of_data.md) we talk about floats in the first section.
Later the term is `decimals`. From `help into decimals` I can see, that it's been used quite as a synonym. So I think we could just use `decimal` and later explain that it's also a float, would that be ok?

Another thing is, that in environement.md are quite a few links to the html file instead of the .md file.
Is that on purpose or could it be cleaned up?

sholderbach:
The preferred way for VuePress is to use relative links to the markdown files. e.g. https://github.com/nushell/nushell.github.io/pull/388/files

Kissaki:
Looks like PR #1099 fixed the html links to md.

---

What's still open about floats vs decimals? How do we want to improve that?

https://www.nushell.sh/book/types_of_data.html

Nushell seems to word it float only. `1.2 | describe`, `42 | into float`. `help into 

```nu
1.2 | describe
# => float
```

```nu
help into float | find decim
# => │ 0 │ Search terms: convert, number, floating, decimal │
```

I guess it makes sense to also mention decimal in the book for discoverability even without float familiarity. With that in mind, I didn't see an improvement on the current docs.

*(I don't like the mixed title and broken anchor between md while it works website on website (with and without minus for slash) but I don't see an obvious better alternative.)*

PR #1099, opened 2023-10-06
Fixing #460 changing .html links to .md links

petrisch:
This should fix at least one part of my naive issue so that we can close it.


amtoine:
would this PR completely close https://github.com/nushell/nushell.github.io/issues/460? :yum:

petrisch:
@amtoine no as i said only partially.
But since i created it and there really shouldn't be two issues in one, we should close it.
But you can also keep it and i will fix it later.

sholderbach (approved):
Good call!

PR #1100, opened 2023-10-07
Added "Use hooks to export state via environment variables" Section in setup.md

cptpiepmatz:
@amtoine meant that something like this could be a nice addition to the cookbook.

I personally use this as the example I gave in here.

### DIFF
--- a/cookbook/setup.md
+++ b/cookbook/setup.md
@@ -136,9 +136,9 @@ currently activated.
 ```nu
 # set NU_OVERLAYS with overlay list, useful for starship prompt
 $env.config.hooks.pre_prompt = ($env.config.hooks.pre_prompt | append {||
-  let nu_overlays = (overlay list | range 1.. | str join ", ")
-  if ($nu_overlays | str length | into bool) {
-    $env.NU_OVERLAYS = $nu_overlays
+  let overlays = overlay list | range 1..
+  if not ($overlays | is-empty) {
+    $env.NU_OVERLAYS = $overlays | str join ", "
   } else {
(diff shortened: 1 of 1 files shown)


cookbook/setup.md
```diff
@@ -114,3 +114,43 @@ or
...
+
+```nu
+# set NU_OVERLAYS with overlay list, useful for starship prompt
+$env.config.hooks.pre_prompt = ($env.config.hooks.pre_prompt | append {||
+  let nu_overlays = (overlay list | range 1.. | str join ", ")
+  if ($nu_overlays | str length | into bool) {
+    $env.NU_OVERLAYS = $nu_overlays
+  } else {
+    $env.NU_OVERLAYS = null
+  }
```
amtoine, line 144: i might just propose the following change
```suggestion
  let overlays = overlay list | range 1..
  if not ($overlays | is-empty) {
    $env.NU_OVERLAYS = $overlays | str join ", "
  } else {
    $env.NU_OVERLAYS = null
  }
```
i find it a bit clearer without the `str length | into bool` but i'll let you decide :yum: 
it's really not a strong request at all :wink:

cptpiepmatz: It is way more clear what the intention is, so this is nice 👍🏻

amtoine: thanks for the quick response :blush:

amtoine (approved):
thanks for this @cptpiepmatz :pray: 

let's run the CI for good measure and then land this PR :muscle:

amtoine:
it's all :green_circle:, let's gooooo :partying_face:

PR #1104, opened 2023-10-08
prevent same entry in $env.PATH

joergsch:
if one extends PATH in `env.nu` and start a different shell before launching nu, one might start to collect the same path multiple times in PATH these lines construct the new value (based on home directory and the separator based on the os)

### DIFF
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -144,6 +144,14 @@ This will append `/some/path` to the end of PATH; you can also use [`prepend`](/
 
 Note the `split row (char esep)` step. We need to add it because in `env.nu`, the environment variables inherited from the host process are still strings. The conversion step of environment variables to Nushell values happens after reading the config files (see also the [Environment](environment.html#environment-variable-conversions) section). After that, for example in the Nushell REPL when `PATH`/`Path` is a list , you can use [`append`](/commands/docs/append.md)/[`prepend`](/commands/docs/prepend.md) directly.
 
+To prepend a new path only if not already listed, one can add to `env.nu`:
+```nu
+# create a new string holding the desired path
+$env.my_path = ( [ $env.HOME, "bin" ] | str join (char psep) )
+# return $env.PATH if $env.my_path is already listed, return $env.PATH with $env.my_path prepended otherwise
+$env.PATH = ( if ( $env.PATH | split row (char esep) | any { |p| $p == $env.my_path } ) { $env.PATH } else { $env.PATH | prepend $env.my_path } )
+```
+
 ### Homebrew
 
 [Homebrew](https://brew.sh/) is a popular package manager that often requires PATH configuration. To add it to your Nushell PATH:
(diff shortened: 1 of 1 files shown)


fdncred:
Seems like this part `$env.my_path = ...` should be `let my_path = ...`. I'm not sure why we'd create an env var just to add it to the `$env.PATH`.

joergsch:
> Seems like this part `$env.my_path = ...` should be `let my_path = ...`. I'm not sure why we'd create an env var just to add it to the `$env.PATH`.

There is no benefit creating an additional env var.
my bad (still learning nu)

book/configuration.md
```diff
@@ -144,6 +144,14 @@ This will append `/some/path` to the end of PATH; you can also use [`prepend`](/
 
 Note the `split row (char esep)` step. We need to add it because in `env.nu`, the environment variables inherited from the host process are still strings. The conversion step of environment variables to Nushell values happens after reading the config files (see also the [Environment](environment.html#environment-variable-conversions) section). After that, for example in the Nushell REPL when `PATH`/`Path` is a list , you can use [`append`](/commands/docs/append.md)/[`prepend`](/commands/docs/prepend.md) directly.
 
+To prepend a new path only if not already listed, one can add to `env.nu`:
+```nu
+# create a new string holding the desired path
+$env.my_path = ( [ $env.HOME, "bin" ] | str join (char psep) )
```
amtoine, line 150: same remark as @fdncred plus another change on `str join` i propose :yum: 
```suggestion
let my_path = ( $nu.home-path | path join "bin" )
```

book/configuration.md
```diff
@@ -144,6 +144,14 @@ This will append `/some/path` to the end of PATH; you can also use [`prepend`](/
...
+# return $env.PATH if $env.my_path is already listed, return $env.PATH with $env.my_path prepended otherwise
+$env.PATH = ( if ( $env.PATH | split row (char esep) | any { |p| $p == $env.my_path } ) { $env.PATH } else { $env.PATH | prepend $env.my_path } )
```
amtoine, line 152: i think this can be simplified quite a bit :innocent: 
```suggestion
$env.PATH = ( if $my_path in $env.PATH { $env.PATH } else { $env.PATH | prepend $my_path } )
```

amtoine (changes requested):
thanks for adding this, that can be quite handy to not have multiple times the same path in `$env.PATH` for some applications :yum: 

just two remarks and i'd be good to go :relieved:

amtoine:
@joergsch 
and if you want to go extra fancy, you can define a more general custom command for this :smirk:
```nushell
def prepend-if-not-in [element: any]: list<any> -> list<any> {
    if $element in $in { $in } else { $in | prepend $element }
}

$env.PATH = ($env.PATH | prepend-if-not-in ($nu.home-path | path join "bin"))
```

amtoine (approved):

amtoine:
thanks for this new section @joergsch, let's land this :pray:

PR #1106, opened 2023-10-09
book/overlays: fix tip div formatting

KAAtheWiseGit:
The `:::` div closer was located as a last part of the list, which broke formatting.

image


fdncred:
Thanks!

PR #1107, opened 2023-10-10
fix variable name

joergsch:
variable was named `second`, but should be named `third` to match the example


sholderbach (approved):
Thanks!

PR #1108, opened 2023-10-10
repeat variable assignment

joergsch:
one might start in the middle of the section
the variable scores is defined above
the example might not result the expected results, hence ensuring assignment is done before


sholderbach (approved):
Good call!

PR #1117, opened 2023-10-20
Update types_of_data.md with "into float"

hatunike:
Update `Decimals` text with the recent `into float` change

Fixes broken link and clarifies correct command.


fdncred:
Thanks!

PR #1119, opened 2023-10-21
update the Direnv section in the cookbook

amtoine:
related to
- https://github.com/nushell/nu_scripts/pull/628
- https://github.com/direnv/direnv/pull/1175

## description
i propose to update the Direnv hook, as in https://github.com/nushell/nu_scripts/pull/628 and https://github.com/direnv/direnv/pull/1175 :yum: 

i've seen a few people complaining about this hook in the cookbook, so here is both a working and a much simpler hook to run Direnv in Nushell :innocent:


amtoine:
this was still not working and lead to an issue on Discord, i'll land the patch :relieved:

PR #1120, opened 2023-10-22
m: fix output of command example when displaying expanded/collapsed table

tkokasih:
Fix output example for expanded and collapsed view of table


fdncred:
The CI is red because we don't allow you to change a command's examples. Those have to be changed in rust code since this command documentation is autogenerated.

tkokasih:
Thanks for the information, looking at the source code [here](https://github.com/nushell/nushell/blob/05efd735b9c02f0dc516bdc750f7be2809c74484/crates/nu-command/src/viewers/table.rs#L157), I think I'm out my depth to fix this, for now.

I will close this PR. Hopefully someone else can fix this.

PR #1123, opened 2023-10-23
add translation for bash "command 2>&1 | less"

rukai:
closes https://github.com/nushell/nushell/issues/10805

This is a very common thing to want to do and isnt documented anywhere that I could find.
I think an addition to this table would be the best way to communicate how to achieve this.

The pipe could go to anything but I chose less as thats the most common destination and it illustrates the point better than using something generic like `command 2>&1 | other_command`

I had to adjust the table width, so you will want to review with "hide whitespace" enabled.


amtoine (approved):
sounds good to me, thanks :pray: 

and EXTRA THANKS for the "hide whitespaces" to review the commit, very helpful :heart:

PR #1124, opened 2023-10-23
Update list of "types" i.e. syntax shapes for `def`s

sholderbach:


amtoine (approved):
thanks @sholderbach for updating 21 releases :pray: :yum:

PR #1127, opened 2023-10-25
Add a note on theming Reedline’s menus

CBenoit:
This adds a note on how Reedline’s menus are themed in the chapter dedicated on coloring and theming.
I found myself looking for this information at the wrong place, and thought it was sensible to at least mention it.


fdncred:
Thanks!

PR #1128, opened 2023-10-26
Book scripts with sub-commands

Roba1993:
This adds a section to the script to describe how to use sub.commands


fdncred:
Thanks!

PR #1130, opened 2023-10-28
Change Ambiguous Example in `select` documentation

stellarpower:
Using 'a' and 'b' for both the keys and the values leaves the reader guessing as to how the command works. Changed the values to be numeric.

---

As a sidenote, I think some of these toy examples might be easier to understand if they were a little more realistic. Personally I find it very easy to gloss over things like "x", "y", "z" or "1", "2", "3", and when they are rather short, it is hard to see what is happening. E.g. some others I have seen use tables with one row. I think two-three minimum helps with working out dimensionality and what we are doing.

For object-orientation, examples are often real-world items that are easy to visualise - such as vehicles, animals, etc. I think the examples starting form `ls` etc. are also good springboards. It helps people think in terms of real data, and that connects how you might realistically use a command. Selecting processes or stats from my machine would also work nicely.

### DIFF
--- a/commands/docs/select.md
+++ b/commands/docs/select.md
@@ -38,20 +38,20 @@ usage: |
 
 Select a column in a table
 ```nu
-> [{a: a b: b}] | select a
+> [{a: 1 b: 2}] | select a
 ╭───┬───╮
 │ # │ a │
 ├───┼───┤
-│ 0 │ a │
+│ 0 │ 1 │
 ╰───┴───╯
 
 ```
 
 Select a field in a record
 ```nu
-> {a: a b: b} | select a
+> {a: 1 b: 2} | select a
 ╭───┬───╮
-│ a │ a │
+│ a │ 1 │
 ╰───┴───╯
 ```
 


fdncred:
The red CI means you can't change a command's help. If you want that change to occur, you'll have to change the rust code because the command documentation is auto-generated from the commands.

stellarpower:
Okay, thanks. I had initially thought it would surely be auto-generated from a tag, but this was what the github search brought up. I guess if it's using the .io hosting that figures. I'll go and find the lines.

amtoine:
@fdncred 
not sure the red CI here is correct, it looks like the CI tried to run the step but failed :thinking: 
just like when i tried this as an experiment in https://github.com/nushell/nushell.github.io/pull/970 :confused:

fdncred:
@amtoine ok. I just see this in the CI and figured it was doing whatever it was supposed to do. Maybe it's just broken?

body: '{"body":"Awesome that you are trying to improve the documentation of a nushell command!\\n\\nAs we autogenerate these command pages from the help information defined with the command after each release:\\n- If you already updated the command help, you are already done!\\n- Else, please look for the command in the [source tree](https://github.com/nushell/nushell/tree/main/crates) and make your changes directly there.\\n\\nThanks for helping out!"}',

PR #1132, opened 2023-10-31
Update table.md

brad-burch:
Simply updating to match current output


hustcer:
Thanks, however, the command docs were generated by a script. It's better to fix it here: https://github.com/nushell/nushell/blob/main/crates/nu-command/src/viewers/table.rs#L133

amtoine:
yeah, thanks for trying to make the documentation better @brad-burch :pray: 

we cannot accept such a PR as @hustcer already said :confused: 
i'll close this PR :relieved:

PR #1133, opened 2023-10-31
Update rename.md

brad-burch:
Noticed the documentation example fails, which is probably a bug (it should probably be able to take a list<string> or a record).  But this example is functionally equivalent, except renaming the second column for clarity in targeting columns besides the first.


fdncred:
I'm not following. This works
```
❯ [[a, b, c]; [1, 2, 3]] | rename --column { a: ham }
╭#┬ham┬b┬c╮
│0│  1│2│3│
╰─┴───┴─┴─╯
```

brad-burch:
I apologize, on v0.85 and trying to learn NuShell and I love it!  Updated to v0.86 and it is indeed fixed (you guys move fast).  Also noticed you generate these, so I will look over in the Rust logic for these examples.  Thank you!

PR #1136, opened 2023-11-05
Add Argument Type Interpretation to Scripts

noverby:
Documents behavior behind issue: https://github.com/nushell/nushell/issues/10887

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -73,6 +73,34 @@ def main [x: int] {
 110
 ```
 
+## Argument Type Interpretation
+
+By default, arguments provided to a script are interpreted with the type `Type::Any`, implying that they are not constrained to a specific data type and can be dynamically interpreted as fitting any of the available data types during script execution. 
+
+In the previous example, `main [x: int]` denotes that the argument x should possess an integer data type. However, if arguments are not explicitly typed, they will be parsed according to their apparent data type. 
+
+For example:
+
+```nu
+# implicit_type.nu
+def main [x] {
+  $"Hello ($x | describe) ($x)"
+}
+
+# explicit_type.nu
+def main [x: string] {
+  $"Hello ($x | describe) ($x)"
+}
+```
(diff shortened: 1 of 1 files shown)


book/scripts.md
```diff
@@ -73,6 +73,32 @@ def main [x: int] {
 110
 ```
 
+## Argument Type Interpretation
+
+By default, arguments provided to a script are interpreted without a specified data type. In the previous example, `main [x: int]`` denotes that the argument x should possess an integer data type. However, if arguments are not explicitly typed, they will be parsed according to their apparent data type.
```
fdncred, line 78: Maybe an extra backtick?

```suggestion
By default, arguments provided to a script are interpreted without a specified data type. In the previous example, `main [x: int]` denotes that the argument x should possess an integer data type. However, if arguments are not explicitly typed, they will be parsed according to their apparent data type. 
```

fdncred:
> By default, arguments provided to a script are interpreted without a specified data type

This is kind of right. Arguments without a specified type are considered `Type::Any`, which is a valid type but can be any of the available types.
```rust
pub enum Type {
    Any,
    Binary,
    Block,
    Bool,
    CellPath,
    Closure,
    Custom(String),
    Date,
    Duration,
    Error,
    Filesize,
    Float,
    Int,
    List(Box<Type>),
    ListStream,
    MatchPattern,
    #[default]
    Nothing,
    Number,
    Range,
    Record(Vec<(String, Type)>),
    Signature,
    String,
    Table(Vec<(String, Type)>),
}
```

noverby:
@fdncred Is this better?

By default, arguments provided to a script are interpreted with the type `Type::Any`, implying that they are not constrained to a specific data type and can be dynamically interpreted as fitting any of the available data types during script execution.

fdncred:
sounds good to me.

fdncred:
Thanks!

PR #1137, opened 2023-11-08
Add Pipelines section about stdout output

cosineblast:
This pull request provides documentation for the semantics of nushell with regard to command stdout output, specially to address some of the misleading behavior that manifests itself in the issue https://github.com/nushell/nushell/issues/10175.

### DIFF
--- a/book/pipelines.md
+++ b/book/pipelines.md
@@ -165,7 +165,7 @@ For instance, the above example sets the `$env.config.table.mode` configuration 
 ```
 
 Because Nushell `$env` variables are [scoped](http://localhost:8080/book/environment.html#scoping), this means that the `table` command in the example is not affected by the
-environemnt modification inside the `do` block and the data will not be shown with the applied configuration.
+environment modification inside the `do` block and the data will not be shown with the applied configuration.
 
 When displaying data early is desired, one may use `print` command:
 
--- a/book/pipelines.md
+++ b/book/pipelines.md
@@ -85,12 +85,14 @@ And the pipeline:
 Are one and the same.
 
 > **Note**
-> the sentence *are one and the same* above only applies for the graphical output in the shell,
+> the sentence _are one and the same_ above only applies for the graphical output in the shell,
 > it does not mean the two data structures are them same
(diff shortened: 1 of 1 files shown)


book/pipelines.md
```diff
@@ -138,3 +140,35 @@ drwxr-xr-x@  4 pengs  admin   128 14 Nov 13:42 en
...
+Knowing when data is displayed is important when using configuration variables that affect the display output of commands such as `table`.
+
+```nu
+> do { $env.config.table.mode = none; ls }
+```
+
+For instance, the above example sets the `$env.config.table.mode` configuration variable to `none`, which causes the `table` command to render data without additional borders. However, as it was shown earlier, the command is effectively equivalent to
+
+```nu
+> do { $env.config.table.mode = none; ls } | table
```
amtoine, line 164: should we add the working example as well for completion?
```nushell
do { $env.config.table.mode = none; ls | table }
```

book/pipelines.md
```diff
@@ -138,3 +140,35 @@ drwxr-xr-x@  4 pengs  admin   128 14 Nov 13:42 en
...
+```
+
+Because Nushell `$env` variables are [scoped](http://localhost:8080/book/environment.html#scoping), this means that the `table` command in the example is not affected by the
+environment modification inside the `do` block and the data will not be shown with the applied configuration.
+
+When displaying data early is desired, one may use `print` command:
+
+```nu
+> do { $env.config.table.mode = none; print(ls) }
```
amtoine, line 173:
```suggestion
> do { $env.config.table.mode = none; print (ls) }
```

amtoine (changes requested):
that's quite interesting each time i read about that :open_mouth: :pray:

book/pipelines.md
cosineblast: Sure

cosineblast:
Done

amtoine (approved):
thanks @cosineblast for clarifying all this non-trivial non-bug behaviour :pray:

PR #1138, opened 2023-11-08
Cookbook ssh agent - alternatives fixing zombie agents

wanesty:
the initial example has a pretty big flaw if simply copy-pased in one's `env.nu`: every time the user opens a new instance of nu it will start a new instance of ssh-agent, which won't exit when nu does

so here's a nushell and two non-nushell ways to achieve a similar result without said flaw

--- 

also the nushell exemple only adds 2-4ms of startup time if the ssh-agent is seen as already running
(R7 5800x & system on nvme)

### DIFF
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -27,11 +27,11 @@ You can work around this behavior by checking if a ssh-agent is already running 
 
 ```nushell
 # exemple for env.nu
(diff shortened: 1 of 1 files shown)


fdncred:
I'm also wondering if having nushell as a login shell and putting it in login.nu would allow it to only run once?

fdncred:
Thanks

PR #1139, opened 2023-11-09
Clarify Dumping files into directory

Traister101:
The created completions directory in the example cannot itself be added to NU_LIB_DIRS but it's parent directory instead. In our example we create the directory in $nu.default-config-dir which should be added to NU_LIB_DIRS

### DIFF
--- a/book/modules.md
+++ b/book/modules.md
@@ -464,8 +464,8 @@ Now you've set up a directory where you can put your completion files and you sh
 
 > **Note**
 > This will use the file name (in our example `git` from `git.nu`) as the modules name. This means some completions might not work if the definition has the base command in it's name.
-> For example a completion like this `export extern 'git push' [] {}` in our `git.nu` will result in a completion like this `git git push`. If you have this style of completion you must instead
-> import like this `use completions git *` which will import the exported definitions of the submoudule `git`. 
+> For example a completion like this `export extern 'git push' []` in our `git.nu` will result in a completion like this `git git push`. If you have this style of completion you must instead
+> import like this `use completions git *` which will import the exported definitions of the submoudule `git`. Or simply rename `git push` to just `push`.
 
 ### Setting environment + aliases (conda style)
 
--- a/book/modules.md
+++ b/book/modules.md
@@ -463,8 +463,9 @@ $env.NU_LIB_DIRS = [
 Now you've set up a directory where you can put your completion files and you should have some Git completions the next time you start Nushell
 
 > **Note**
-> This will use the file name (in our example `git` from `git.nu`) as the modules name. This means completions like the ones found in the [nu scripts](https://github.com/nushell/nu_scripts) repo won't work
-> as you will end up with completions like `git-completions git ...`. You should instead do `use completions git-completions *` which will import the exported definitions of the submodule `git-completions` 
+> This will use the file name (in our example `git` from `git.nu`) as the modules name. This means some completions might not work if the definition has the base command in it's name.
+> For example a complition like this `export extern 'git push' [] {}` in our `git.nu` will result in a completion like this `git git push`. If you have this style of completion you must instead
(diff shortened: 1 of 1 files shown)


Traister101:
Could also change the guide to instead have the directory created inside `scripts` which is added to NU_LIB_DIRS by default removing the need to add a new directory to the environment variable at all

amtoine:
@Traister101 
to be honest i think i would even rewrite the section entirely as follows :eyes:

Traister101:
Definitely think it's better than what's currently there (my changes included). Since it's so much more through however I think that "git" being used as the module name should be mentioned. It's not an issue in the example but if you drop completions from the [nu scripts reop](https://github.com/nushell/nu_scripts) in they won't work. As an example dropping the `git-completions.nu` file in will result in completions like `git-completions git ...` the second `git` is from how the completions are defined, they have `git` prefixed at the front as it's expected to be imported (`use`) file by file.

Basically the completions in the nu script repo are fundamentally incompatible with this drop in style. Seeing as they are probably the most popular if not the only collection of the things I think it's probably good to mention so it doesn't cause confusion (like it did with me)

I *do* like the full rewrite you've done however. How should that be handled? Have very little experience actually collaborating with people through git. I assume I just make a new commit keeping the pr?

Traister101:
@amtoine Should I have pinged you? I thought you'd get notified since you commented but maybe not?

amtoine:
> Definitely think it's better than what's currently there (my changes included). Since it's so much more through however I think that "git" being used as the module name should be mentioned. It's not an issue in the example but if you drop completions from the [nu scripts reop](https://github.com/nushell/nu_scripts) in they won't work. As an example dropping the `git-completions.nu` file in will result in completions like `git-completions git ...` the second `git` is from how the completions are defined, they have `git` prefixed at the front as it's expected to be imported (`use`) file by file.
> 
> Basically the completions in the nu script repo are fundamentally incompatible with this drop in style. Seeing as they are probably the most popular if not the only collection of the things I think it's probably good to mention so it doesn't cause confusion (like it did with me)
> 
yeah that's a good point!
so maybe taking the example of the `nu_scripts` directly would be more profitable for the long term :thinking: 
you'd mention the `nu_scripts` completion directory, briefly how to install it and then the proper `use` call to import the Git completions without the `git-completions` namespace prefix :+1: 
alternately, you could just define `git`, `git checkout` and `git add` in the snippet above and `use completions git *` in `config.nu` which should be much more similar to what you'd do with the `nu_scripts`.

i'd leave that up to you, honestly i like both ways:
- the real life scenario with the real `nu_scripts`
- the simpler and yet complete example with a slightly modified snippet from above

:wink: 

> I _do_ like the full rewrite you've done however. How should that be handled? Have very little experience actually collaborating with people through git. I assume I just make a new commit keeping the pr?
>
yep, you can just commit on top of the PR branch with the new chapter :yum: 
no need to rebase and force push :innocent:

Traister101:
Gah why did I chose to do this from GitHub instead of locally

Traister101:
Okay @amtoine should be done.
I should really just not use GitHub to edit stuff. I've regretted it every time so far.

kubouch:
Looks good to me, I just wouldn't make the Note nu_scripts-specific because the contents of nu_scripts might change (e.g., someone could rewrite the git-completions to work with the module directory style).

Traister101:
Hmm that's a good point. Could use a particular commit but it'd probably be best to be more generic in the note.

kubouch:
OK, thank you! It now looks much better. I added a few clarifications and fixed some minor things.

amtoine:
> Okay @amtoine should be done. I should really just not use GitHub to edit stuff. I've regretted it every time so far.
>
ehe no worries :wink: 

i like Git in the CLI as well :relieved: 

thanks for this better chapter about directory modules :pray:

PR #1140, opened 2023-11-11
Fix broken link

MarikaChlebowska:
Fix for link to `localhost`


amtoine (approved):
good catch, the new link looks better indeed :pray:

PR #1142, opened 2023-11-13
Add a snippet to suggest using the ? operator for environmental variables

132ikl:
I didn't initially realize the `?` operator existed (there's only two sections in the book about this, both of which are easy to miss!), and originally accomplished this using `$env | get --ignore-errors FOO` as suggested by `help default`. This seems like a common use case for the `?` operator, so I added a little snippet to guide users towards using `?`

### DIFF
--- a/book/environment.md
+++ b/book/environment.md
@@ -95,6 +95,21 @@ Error: nu::shell::column_not_found
 
 > $env.FOO? | describe
 nothing
+
+> $env.FOO? | default "BAR"
+BAR
+```
+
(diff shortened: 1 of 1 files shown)


fdncred:
You can also do `"FOO" in $env` to check for an env var.

132ikl:
Do you think it would be worth adding another snippet to mention that? I also thought about providing an example using `default` but I wasn't sure if that would clutter up the page too much.

fdncred:
I'd support an `in` example and a `default` example.

fdncred:
Thanks!

PR #1150, opened 2023-11-19
Add a cookbook comparing jq and nushell

arnau:
Hi, this PR adds a new cookbook comparing jq and nushell as suggested by @fdncred in https://github.com/arnau/seachess/issues/35

I translated my original blog post using a hopefully  better outline for incrementally introduing new Nushell commands and concepts.

I would appreciate an extra eye on the written English. I'm not a native speaker so I'm sure I've done at least some silly mistakes. 🙇

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -147,6 +147,7 @@ export const sidebarEn: SidebarConfig = {
         'ssh_agent',
         'tables',
         'polars_v_pandas_v_nushell',
+        'jq_v_nushell',
       ],
     },
   ],
--- a/cookbook/jq_v_nushell.md
+++ b/cookbook/jq_v_nushell.md
@@ -4,7 +4,7 @@ title: jq vs Nushell
 
 # jq vs Nushell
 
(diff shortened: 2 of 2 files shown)


fdncred (approved):
This looks fantastic! Thank you so much. I read it and tweaked it a little bit. We really appreciate you adding this great learning utility.

arnau:
@fdncred I see you removed the `>` which I added because I saw them used in other cookbooks. I definitely prefer not having them around. Should we perhaps remove them from all other cookbooks? happy to raise a PR for it.

amtoine:
@arnau 
i'd say it depends :wink: 

here i find it better without them, completely agree :ok_hand: 
a place where it's usefull, we you have a command and it's return value in the same cell
```nushell
> [1, 2, 3] | math sum
6
```

arnau:
@amtoine fair. I guess I prefer to split the return value out. More verbose and takes much move vertical space though 🤷

amtoine:
> @amtoine fair. I guess I prefer to split the return value out. More verbose and takes much move vertical space though 🤷
>
i reviewed your PR :wink: 
i'm open to other feedback on that question, at least the two ways share the same information, so that's the most important :pray:

fdncred:
I just said this in the other PR but I'll say it here too. My preferences is to have the command in a code block without the `>` and the output in a separate code block. If we have them both the command and the output in one code block, then it should keep the `>`.

PR #1151, opened 2023-11-19
Unifies all cookbook codeblocks to use the same style.

arnau:
The unified style favours the removal of any prompt indicator such as `>`.

When appropriate I split `nu` blocks and output blocks as well.

And fixed an example in pattern_matching.md

### DIFF
--- a/cookbook/README.md
+++ b/cookbook/README.md
@@ -11,7 +11,7 @@ where you will find interesting scripts written for Nushell.
 And if you are looking for cool oneliners like this one
 
 ```nu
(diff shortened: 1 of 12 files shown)


amtoine (commented):
pretty neat :+1: 

i just think having `>` for a command and it's output is fine, e.g.
```nushell
> command
output
```
but i'm fine with either way :relieved:

fdncred:
I removed all the `>` in @arnau's PR because of how his was designed. You have the jq command, then the nushell command, then the output. So, I like that style without the `>` so it's easier to copy-n-paste.

PR #1152, opened 2023-11-20
Add `parse` examples to *Parsing*

ppenguin:
Add more examples how to conveniently parse tabular text output from legacy shell utils.


amtoine (approved):
looks pretty good thanks :pray:

PR #1157, opened 2023-11-27
Replace usage of removed `date format` with `format date` in examples

MarikaChlebowska:
I've noticed one of the examples doesn't work because of removed command so I replace usage of it in all the examples with the new one. It's still part of automatically generated documentation but that should be changed in the main nushell repo which should be fixed as well to avoid confusion.


fdncred:
Good catch. Thanks!

PR #1159, opened 2023-11-29
Improve workaround for carapace ERR output

stfacc:
In case of an unknown flag somewhere in the span, carapace outputs the last element of the span followed by the string "ERR". If the flag itself is not the last element, the regex won't match the error message. Instead, check the exact expected value.

Test with
```nu
> chmod -x foo<TAB>
```


fdncred:
Thanks

PR #1164, opened 2023-12-05
Fix append to mutable variable operator

KAAtheWiseGit:
Replaced an outdated `**=` operator with the current `++=`.


fdncred:
You're on a roll today. Thanks so much! We appreciate all your hard work to make nushell better.

Issue #1035, opened 2023-08-30
Documentation for 'def' keyword signature should include optional syntax for the return type of the custom command


edhowland:
### Describe the bug

Custom commands defined using the 'def' keyword allow, in addition to specifying
the  optional parameter types, also allow for specifying an optional return
type. E.g.

```sh
# returns the square of the parameter
def square [x: int] -> int {
  $x * $x
}
```

The '-> <type>'  is not mentioned in any of the documentation for either
the 'def' keyword or the Nushell book chapter on  Custom Commands:

## The def keyword

https://www.nushell.sh/commands/docs/def.html

Note: The same text as in the command reference for the 'def' keyword is output
when you type:

```sh
help def
```

and should also include the optional '-> <type> part of the signature.

## The Custom Command chapter

https://www.nushell.sh/book/custom_commands.html

In the command reference or the help text there should be:

1. Part of the signature between the parameters and the body should be the return type.
2. Examples of using the 'def' keyword should include an example with the '-> <type>
  * Perhaps like the square example above.

In the chapter on "Custom Commands",
there should be a  Heading 2 section called something like

## Specifying the return type of your command

Custom commands can optionally specify their return type which helps the Nushell
compiler to check your usage when calling your custom command especially in an expression.

Eg. Say we just want to format our greeting like this:

```sh
def fmt-greet [name: string] -> string {
  $"Hello ($name)"
}```

Then for another command that takes a greeting as input:

```sh
def print-greeting [greeting: string] {
  echo $"When you greet someone, you should say: '($greet)'"
}

# then to call it
print-greet (fmt-greet "Fred")
```

### How to reproduce

This is just a documentation ommision. There is no steps to reproduce this bug. Just  add the missing documentation.

### Expected behavior

The documentation on the web site for the Nushell book and the command reference should be updated to include the optional return type syntax for defining custom commands. This should also be included in the help text for the 'def' keyword.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.84.0                                                            |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | macos-aarch64                                                     |
| build_target       | aarch64-apple-darwin                                              |
| rust_version       | rustc 1.71.1 (eb26296b5 2023-08-03) (built from a source tarball) |
| cargo_version      | cargo 1.71.2 (1a737af0c 2023-08-07)                               |
[...]

PR #1167, opened 2023-12-10
Update command signature doc to modern syntax

OrionOth:
While implementing a custom alias, I found myself unable to provide an explicit command signature as described in the documentation. A bit of crawling the main repo's commit history led to my discovery that the syntax as described in the documentation was indeed outdated.
After a bit of experimentation I was able to discover the correct annotation syntax.
Ideally, in a slightly longer timescale than I am willing to deal with right now, this change shouldn't be needed; I believe the information on this page would serve better integrated with the rest of the documentation. It seems as though it would fit better there after the syntax change.

### DIFF
--- a/book/command_signature.md
+++ b/book/command_signature.md
@@ -1,16 +1,12 @@
 # Command signature
 
-nu commands can be given explicit signatures; take [`histogram`](/commands/docs/histogram.md) as an example, the signature is like this:
+nu commands can be given explicit signatures; take [`str stats`](/commands/docs/str_stats.md) as an example, the signature is like this:
 
 ```nu
-def histogram [
-  column-name?: string,
-  frequency-column-name?: string,
-  --percentage-type (-t): string = "normalize"
-]: list<any> -> table { }
+def "str stats" []: string -> record { }
 ```
 
-The type names between the `:` and the opening curly brace `{` describe the command's input/output pipeline types. The input type for a given pipeline, in this case `list<any>`, is given before the `->`; and the output type `table` is given after `->`. There can be multiple input/output types. If there are multiple input/output types, they can be placed within brackets and separated with commas, as in [`str join`](/commands/docs/str_join.md):
+The type names between the `:` and the opening curly brace `{` describe the command's input/output pipeline types. The input type for a given pipeline, in this case `string`, is given before the `->`; and the output type `record` is given after `->`. There can be multiple input/output types. If there are multiple input/output types, they can be placed within brackets and separated with commas, as in [`str join`](/commands/docs/str_join.md):
 
 ```nu
 def "str join" [separator?: string]: [list -> string, string -> string] { }
--- a/book/command_signature.md
+++ b/book/command_signature.md
@@ -1,24 +1,22 @@
 # Command signature
 
-nu commands contain a signature section, take [`str distance`](/commands/docs/str_distance.md) as example, the signature is like this:
+nu commands can be given explicit signatures; take [`str distance`](/commands/docs/str_distance.md) as an example, the signature is like this:
 
 ```nu
-Signatures(Cell paths are supported):
-  <string> | str distance <string> -> <int>
+def "str distance" [comapare-string: string, ...rest: cell-path]: [string -> int, table -> table, record -> record]
 ```
 
-The first type name before `|` describes the type of input pipeline. The command name is followed by the required argument type(s) for the command. The output type is `int` and given after `->`.
+The type names in the brackets after the `:` describe the command's input/output pipeline types. The input type for a given pipeline, in this case `string`, is given before the `->`; and the output type `int` is given after `->`. There can be multiple input/output types.
 
-`(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply an operation at the given cell path(s) in a nested structure or table, and replace the column or field with the result, like:  `ls | str distance 'nushell' 'name'`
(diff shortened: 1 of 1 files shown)


OrionOth:
Oops! I forgot to check for issues about this before I committed! Looks like this will close #1035 (at least the web documentation) after another minor change.

book/command_signature.md
```diff
@@ -1,24 +1,22 @@
 # Command signature
 
-nu commands contain a signature section, take [`str distance`](/commands/docs/str_distance.md) as example, the signature is like this:
+nu commands can be given explicit signatures; take [`str distance`](/commands/docs/str_distance.md) as an example, the signature is like this:
 
 ```nu
-Signatures(Cell paths are supported):
-  <string> | str distance <string> -> <int>
+def "str distance" [compare-string: string, ...rest: cell-path]: string -> int { }
```
amtoine, line 6: to follow completely the i/o types of `str distance`, i think it should be
```suggestion
def "str distance" [compare-string: string, ...rest: cell-path]: [string -> int, table -> table, record -> record] { }
```

amtoine: EDIT: aah i see you mention the full signature below :+1: 
maybe just add here that this is a simplified signature?

book/command_signature.md
```diff
@@ -1,24 +1,22 @@
...
 ```
 
-The first type name before `|` describes the type of input pipeline. The command name is followed by the required argument type(s) for the command. The output type is `int` and given after `->`.
+The type names after after the `:` describe the command's input/output pipeline types. The input type for a given pipeline, in this case `string`, is given before the `->`; and the output type `int` is given after `->`. There can be multiple input/output types. If there are multiple input/output types, they can be placed within brackets and separated with commas: `str distance`'s signature actually has the `[string -> int, table -> table, record -> record]` pipeline types.
```
amtoine, line 9:
```suggestion
The type names after the `:` describe the command's input/output pipeline types. The input type for a given pipeline, in this case `string`, is given before the `->`; and the output type `int` is given after `->`. There can be multiple input/output types. If there are multiple input/output types, they can be placed within brackets and separated with commas: `str distance`'s signature actually has the `[string -> int, table -> table, record -> record]` pipeline types.
```

book/command_signature.md
```diff
@@ -1,24 +1,22 @@
...
-`(Cell paths are supported)` indicates that you can provide cell paths for `str distance` to apply an operation at the given cell path(s) in a nested structure or table, and replace the column or field with the result, like:  `ls | str distance 'nushell' 'name'`
+The `cell-path` type indicates that you can provide cell paths for `str distance` to apply an operation at the given cell path(s) in a nested structure or table, and replace the column or field with the result, like: `ls | str distance 'nushell' 'name'`
 
-Here is another one example, [`str join`](/commands/docs/str_join.md):
+Here is another example, [`str join`](/commands/docs/str_join.md):
 
 ```nu
-Signatures:
-  list<string> | str join <string?> -> <string>
+def "str join" [separator?: string]: [list<string> -> string, string -> string] { }
```
amtoine, line 16:
```suggestion
def "str join" [separator?: string]: [list<any> -> string, string -> string] { }
```

book/command_signature.md
```diff
@@ -1,24 +1,22 @@
...
 ```
 
-It says that [`str join`](/commands/docs/str_join.md) command expect input pipeline is a list of string, and take optional `string` type argument, finally the output type is `string`.
+It says that the [`str join`](/commands/docs/str_join.md) command takes an optional `string` type argument, and an input pipeline of either a list of `string`s with output type of `string`, or a single `string`, again with output type of `string`.
```
amtoine, line 19:
```suggestion
It says that the [`str join`](/commands/docs/str_join.md) command takes an optional `string` type argument, and an input pipeline of either a list of `string`s with output type of `any`thing, or a single `string`, again with output type of `string`.
```

amtoine (changes requested):
that's a pretty nice update thanks :pray: 

i have just a few tweaks to ask in the review below :innocent: 

also i'm wondering about the _abstract signature syntax_, should it be something like
```nushell
"str join" [separator?: string]: [list<string> -> string, string -> string]
```
i.e. without the `def` or the `{ }`?

:point_right: maybe the `def` is useful in case you use `def --env` or `def --wrapped` which would change the signature :+1:

OrionOth:
Thanks for the suggestions @amtoine! I couldn't find your email for a DCO, though, sorry about that.

book/command_signature.md
OrionOth: I replaced `list<string>` with `list` in `42ea6a8` to clarify in text that it implies `list<any>`.

book/command_signature.md
OrionOth: I'm not sure if you meant to edit the input type instead, but either way I've clarified this a bit more in `[42ea6a8`, based on the current published web docs.

OrionOth:
Unfortunately it does not like there is a subcommand within the `str` family with `<nothing> -> `<nothing>` pipeline type, so that one will have to remain off-theme (for now! 👀). Perhaps `str` itself, but that simply takes `nothing` as input, it does have pipeline output.
Thanks again @amtoine!

amtoine:
> Unfortunately it does not like there is a subcommand within the `str` family with `<nothing> -> <nothing>` pipeline type, so that one will have to remain off-theme (for now! 👀). Perhaps `str` itself, but that simply takes `nothing` as input, it does have pipeline output.
>
not sure what `<nothing> -> <nothing>` is refering too :open_mouth: 

> Thanks again @amtoine!
>
anytime :wink:

amtoine (approved):

OrionOth:
Haha, now the German translation, at least, is outdated...

PR #1169, opened 2023-12-11
update the SSH agent snippet

amtoine:
i did a logout on my machine and noticed the SSH agent i took from the book stopped working :open_mouth: 
i fixed it and thought it would be great to update the book :innocent: 

## description
- uses `snake_case` which is i think the convention with Nushell
- uses [do --env](https://github.com/nushell/nushell.github.io/compare/main...amtoine:fix-ssh-snippet?expand=1#diff-e98bd788993dd2508080f08a338ade8ac95c61eda25455a4a5f8aad2a09fde21R29) to avoid leaking internal definitions to the REPL main scope
- defines the SSH agent path in a [cross-platform way](https://github.com/nushell/nushell.github.io/compare/main...amtoine:fix-ssh-snippet?expand=1#diff-e98bd788993dd2508080f08a338ade8ac95c61eda25455a4a5f8aad2a09fde21R31)
- when the SSH agent file is there
  - if the process is running => [load the environment and early return](https://github.com/nushell/nushell.github.io/compare/main...amtoine:fix-ssh-snippet?expand=1#diff-e98bd788993dd2508080f08a338ade8ac95c61eda25455a4a5f8aad2a09fde21R38) 
  - if the process is not there, e.g. after a logout without reboot => [remove the SSH agent file](https://github.com/nushell/nushell.github.io/compare/main...amtoine:fix-ssh-snippet?expand=1#diff-e98bd788993dd2508080f08a338ade8ac95c61eda25455a4a5f8aad2a09fde21R40)
- create the agent by default

### DIFF
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -26,22 +26,29 @@ See the workarounds.
 You can work around this behavior by checking if a ssh-agent is already running on your user, and start one if none is:
 
 ```nushell
-# exemple for env.nu
-let sshAgentFilePath = $"/tmp/ssh-agent-($env.USER).nuon"
+do --env {
+    let ssh_agent_file = (
+        $nu.temp-path | path join $"ssh-agent-($env.USER? | default $env.USERNAME).nuon"
+    )
 
-if ($sshAgentFilePath | path exists) and ($"/proc/((open $sshAgentFilePath).SSH_AGENT_PID)" | path exists) {
-    # loading it
-    load-env (open $sshAgentFilePath)
-} else {
-    # creating it
-    ^ssh-agent -c
+    if ($ssh_agent_file | path exists) {
+        let ssh_agent_env = open ($ssh_agent_file)
+        if ($"/proc/($ssh_agent_env.SSH_AGENT_PID)" | path exists) {
+            load-env $ssh_agent_env
+            return
+        } else {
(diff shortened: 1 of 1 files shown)


fdncred:
@amtoine i'm assuming you forgot about this PR. is it ready to land?

amtoine:
@fdncred 
aah yeah, it looks like i forgot :open_mouth: 

as it looks exactly the same as the one i use everyday, i guess it's ready to land :wink:

fdncred:
lol. Thanks!

cookbook/ssh_agent.md
```diff
@@ -26,22 +26,29 @@ See the workarounds.
...
-
-if ($sshAgentFilePath | path exists) and ($"/proc/((open $sshAgentFilePath).SSH_AGENT_PID)" | path exists) {
-    # loading it
-    load-env (open $sshAgentFilePath)
-} else {
-    # creating it
-    ^ssh-agent -c
+do --env {
+    let ssh_agent_file = (
+        $nu.temp-path | path join $"ssh-agent-($env.USER? | default $env.USERNAME).nuon"
```
wanesty, line 31: you wrote `$env.USER? | default $env.USERNAME` for crossplatform compatibility but `$env.USERNAME` doesn't exists on linux, so i think both `$env` user variables would need a `?` and thus maybe a check

PR #1170, opened 2023-12-11
Update `max_history_size` to grouped option `max_size`

TheJiahao:
`max_history_size` in `$env.config` doesn't work. Updated it to `max_size` in history subrecord according to https://www.nushell.sh/blog/2022-11-29-nushell-0.72.html


amtoine:
good catch!
looking at the [default configuration](https://github.com/nushell/nushell/blob/main/crates/nu-utils/src/sample_config/default_config.nu#L199), it should be what your changes show, thanks :pray:

PR #1176, opened 2023-12-16
Add information on spread operator

ysthakur:
Adds information on spread operator in records, lists, and command calls. Should it all go on the "Operators" page?


book/types_of_data.md
```diff
@@ -245,6 +245,21 @@ However, if a record has a key name that can't be expressed as a bare string, or
 false
 ```
 
+To make a copy of a record with new fields, you can use the spread operator (`...`):
+
+```nu
```
amtoine, line 250: i think it should be `nushell` for the syntax highlighting to kick in, or do we support `nu`?

amtoine (commented):
> Should it all go on the "Operators" page?
>
i don't think it's that critical for now :thinking: 
the documentation will likely go under a pretty huge refactor before 1.0 so i'd say the most important is to have all the information somewhere :relieved:

book/types_of_data.md
ysthakur: The other code blocks in the same file used `nu`, so I think both work

amtoine: okey, gotcha :ok_hand:

fdncred:
Just to say what I said in discord, "I'd like to have one page that documents all the features of the spread operator and then a minor blurb on the pages with lists and records that link to the more thorough documentation."

fdncred:
Thanks! Excited to play with this feature along with the docs.

PR #1177, opened 2023-12-17
Typo in history menu section

braddunbar:
I'm fairly sure this should be `history_menu` instead of `help_menu`.


fdncred:
Thanks

PR #1178, opened 2023-12-17
Sync zh-CN shells_in_shells.

MediosZ:
Sync the zh-CN shells_in_shells with the main book to avoid misleading behavior.
Small fix but means a lot.


amtoine:
@hustcer will probably be able to read the changes :yum:

hustcer:
Thanks

PR #1179, opened 2023-12-19
I've revised the Japanese translation

YukiOnodera:
I've revised the Japanese translation as it was somewhat off, ensuring it now feels more natural and accurate.


hustcer:
Thanks

PR #1180, opened 2023-12-19
logo contest announcement

amtoine:
## TODO
- [x] the themes of the contest in 06cda2a4
- [x] the dates in c081238f


amtoine:
> I'd still like all submissions from email to be logged into a nushell repo somewhere. I'd like to be 100% transparent on submissions (except for submissions that are "bad" like porn, etc) and think having them in a repo would be helpful.
>
we will share them all after the preselection for sure!

or do you mean candidates should submit their work as PRs against a GitHub repo instead of by email?

fdncred:
I'm fine with candidates emailing submissions but I'm saying that I think there also needs to be some task to take those submissions and submit them to a repo somewhere for public viewing.

PR #1182, opened 2023-12-21
add information about switches

WindSoilder:
follow-up to
- https://github.com/nushell/nushell/pull/11365

### DIFF
--- a/book/custom_commands.md
+++ b/book/custom_commands.md
@@ -301,7 +301,7 @@ You can also assign it to true/false to enable/disable the flag too:
 
 But note that this is not the behavior you want: `> greet -a 10 --switch false`, here the value `false` will pass as a positional argument.
 
-To avoid confusing, it's not allowed to annotate a boolean type on a flag:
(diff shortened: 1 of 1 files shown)


book/custom_commands.md
```diff
@@ -292,6 +292,25 @@ Or just without the switch flag:
...
+You can also assign it to true/false to enable/disable the flag too:
+
+```nu
+> greet -a 10 --switch=false
+> greet -a 10 --switch=true
+```
+
+But note that this is not the behavior you want: `> greet -a 10 --switch false`, here the value `false` will pass as a positional argument.
+
+To avoid confusing, it's not allowed to annotate a boolean type on a flag:
```
amtoine, line 304: i would say
```suggestion
To avoid confusion, it's not allowed to annotate a boolean type on a flag:
```

amtoine (changes requested):
this is a nice addition :ok_hand: 

just a minor thing and then we're good to go :yum:

fdncred:
Thanks Wind!

PR #1188, opened 2024-01-01
Add info on spread operator in commands for 0.89 release notes

ysthakur:
Added a section about the spread operator being supported in command calls


fdncred:
Thanks

PR #1190, opened 2024-01-01
Update aliases adding new section

zolodev:
Adding a new section of how to "replace" a command with an alias.

This might not be "best practice" but I found this solution like many other had asked before me on Discord.
That I thought could be good to have in the documentation.

This is my first commit to this projects so I thought I keep it simple to begin with.

### DIFF
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -71,7 +71,7 @@ alias core-ls = ls    # This will create a new alias core-ls for ls
 Now you can use `core-ls` as `ls` in your nu-programming. You will see further down how to use `core-ls`.
 
 The reason you need to use alias is because, unlike `def`, aliases are position-dependent. So, you need to "back up" the old command first with an alias, before re-defining it.
-If you do not backup the command and you replace the command using `def` you can end up gtting a recursion error.
+If you do not backup the command and you replace the command using `def` you get a recursion error.
 
 ```nu
 def ls [] { ls }; ls    # Do *NOT* do this! This will throw a recursion error
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -69,3 +69,25 @@ alias ls = do { ls | sort-by type name -i} # Needs to be below other alias like 
 ```
 The reason why `ls` needs to be below alias `ll` is because we are "overriding" the normal `ls` with this function that does not take any arguments.
 And the ls is used with flag `ls -l` in the `ll` alias.
+
+Another way to do it is to shadow the `ls` command. This way you do not use the "hacky" way using `do` command.
+
+```nu
+# An escape hatch to have access to the original ls command
+alias core-ls = ls
+
+# Call the built-in ls command with a path parameter
+def old-ls [path] {
+  core-ls $path | sort-by type name -i
+}
+
+# Shadow the ls command so that you always have the sort type you want
+def ls [path?] {
+  if $path == null {
+    old-ls .
+  } else {
+    old-ls $path
+  }
+}
+```
+
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -60,17 +60,36 @@ def lsg [] { ls | sort-by type name -i | grid -c | str trim }
 
 displaying all listed files and folders in a grid.
 
-## Replacing aliases
-> Caution! When replacing commands like below, it is not possible to pass flags to the command.
+## Replacing existing commands using aliases
+
+> Caution! When replacing commands it is best to "back up" the command first and avoid recursion error.
+
+How to back up a command like `ls`:
+```nu
+alias core-ls = ls    # This will create a new alias core-ls for ls
+```
+Now you can use `core-ls` as `ls` in your nu-programming. You will see further down how to use `core-ls`.
+
+The reason you need to use alias is because, unlike `def`, aliases are position-dependent. So, you need to "back up" the old command first with an alias, before re-defining it.
+If you do not backup the command and you replace the command using `def` you can end up gtting a recursion error.
 
 ```nu
-alias ll = do { ls -l | sort-by type name -i}
-alias ls = do { ls | sort-by type name -i} # Needs to be below other alias like `ll` when using ls
-```
-The reason why `ls` needs to be below alias `ll` is because we are "overriding" the normal `ls` with this function that does not take any arguments.
-And the ls is used with flag `ls -l` in the `ll` alias.
+def ls [] { ls }; ls    # Do *NOT* do this! This will throw a recursion error
 
-Another way to do it is to shadow the `ls` command. This way you do not use the "hacky" way using `do` command.
+#output:
+#Error: nu::shell::recursion_limit_reached
+#
+#  × Recursion limit (50) reached
+#     ╭─[C:\Users\zolodev\AppData\Roaming\nushell\config.nu:807:1]
+# 807 │
+# 808 │ def ls [] { ls }; ls
+#     ·           ───┬──
+#     ·              ╰── This called itself too many times
+#     ╰────
+```
+
+The recommended way to replace an existing command is to shadow the command.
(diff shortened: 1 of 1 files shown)


fdncred:
These aliases don't allow you to pass parameters. It might be good to mention that. I also see this, using `do`, as kind of a hack to our current system that may not be around forever.

zolodev:
> These aliases don't allow you to pass parameters. It might be good to mention that. I also see this, using `do`, as kind of a hack to our current system that may not be around forever.

Yes, I just thought it seems like many have asked about how to "replace" ls command (myself included), that was why I thought I should contribute to the docs about "how" to do it.

Regarding passing parameters, I thought the "caution" text would be enough on line 64. 
```suggestion
> Caution! When replacing commands like below, it is not possible to pass flags to the command.
```

Do you have a better suggestion to how I can change this to be better or more clear?
My new suggestion might be better:
```suggestion
> Caution! Alias like these don't allow you to pass parameters.
```

kubouch:
I wouldn't encourage the `do` workarounds because it doesn't pass parameters to `ls`. It's better to use custom commands if you want to wrap something more complex, such as a pipeline.

zolodev:
> I wouldn't encourage the `do` workarounds because it doesn't pass parameters to `ls`. It's better to use custom commands if you want to wrap something more complex, such as a pipeline.

I agree, if that was possible I think it would not be a problem. Like @fdncred said as well, `do` is more of a hack, but until any other way to "replace" a command like `ls`, is it not a valid way?

Reason why I think it could be a good idea to have it in the documentation, is because I noticed that this question have been asked several times on Discord. And until there is a "better" way to do it, would it not be better to document it? There is always a possibility to update the documentation when this changes. Or at least that was my thought. 😄 

But I am kind of new to nushell, just found it yesterday and thought that I would contribute with what I think is important Quality-of-Life improvements.

fdncred:
I'm fine with your new suggestion. It's probably good enough for now.

I haven't looked at this section, but the "best practice" is probably to write a custom command since our aliases don't allow you to do this. Maybe something like this?

```nushell
# An escape hatch to have access to the original ls command
alias core-ls = ls

# Call the built-in ls command with a path parameter
def old-ls [path] {
  core-ls $path | sort-by type name -i
}

# Shadow the ls command so that you always have the sort type you want
def ls [path?] {
  if $path == null {
    old-ls .
  } else {
    old-ls $path
  }
}
```
I'm not saying this is the best way to do things, just another way.

zolodev:
> I'm fine with your new suggestion. It's probably good enough for now.
> 
> I haven't looked at this section, but the "best practice" is probably to write a custom command since our aliases don't allow you to do this. Maybe something like this?
> 
> ```nushell
> # An escape hatch to have access to the original ls command
> alias core-ls = ls
> 
> # Call the built-in ls command with a path parameter
> def old-ls [path] {
>   core-ls $path | sort-by type name -i
> }
> 
> # Shadow the ls command so that you always have the sort type you want
> def ls [path?] {
>   if $path == null {
>     old-ls .
>   } else {
>     old-ls $path
>   }
> }
> ```
> 
> I'm not saying this is the best way to do things, just another way.

I like that solution!

Now I think it is "to much" for a simple solution. But I think we can add this as another solution along side, to let the user decide on what is a better option ❓

kubouch:
Maybe the text could more clearly formulate that it's about replacing some already defined command. The reason you need to go via alias is that, unlike `def`s, aliases are position-dependent. So, you need to “back up” the old command first with an alias, before re-defining it. If you used only `def`s, you'd end up with infinite recursion (try `def ls [] { ls }; ls`).

We already have a section about pipelines in aliases https://www.nushell.sh/book/aliases.html#piping-in-aliases, so I wouldn't mention the `do` hack here at all, this section is about a different topic. The `do` hack can be mentioned in the linked section as a potential workaround, with a note that in those cases you probably want to use `def` as using alias+do does not bring any benefits.

zolodev:
> Maybe the text could more clearly formulate that it's about replacing some already defined command. The reason you need to go via alias is that, unlike `def`s, aliases are position-dependent. So, you need to “back up” the old command first with an alias, before re-defining it. If you used only `def`s, you'd end up with infinite recursion (try `def ls [] { ls }; ls`).

Yes, I will update the text to try clarifying it more that it replaces an existing defined command. I will try to elaborate about trying to avoid the recursion as well. That was something myself included ended up doing. That got me searching for a simple way to "replace" ls with `alias ls = do { ls | sort-by type name -i}`. I have worked with bash before and trying to learn the ropes of nushell (and still learning). So I tried to "think bash" with what I wanted to do, and trying to "one line" the command. 
 
> We already have a section about pipelines in aliases https://www.nushell.sh/book/aliases.html#piping-in-aliases, so I wouldn't mention the `do` hack here at all, this section is about a different topic. The `do` hack can be mentioned in the linked section as a potential workaround, with a note that in those cases you probably want to use `def` as using alias+do does not bring any benefits.

This is what I think will get to many coming from bash, due to the "simplicity" using aliases in bash, where you can use piping in aliases. And while it is technical possible to still use `do` to achieve this in nushell, would it not be a good idea to add it to the documentation in the mean time ❓ So new users that want to replace `ls` can do just that without the need to ask the question in Discord in the future? It is possible to "update" the documentation the day when either `do` command is obsolete or a better solution is in place. 😇  

I think the documentation should bring value, and I think the documentation should also help the user to achieve the goal of what they are trying to do. Otherwise the documentation will probably end up on someones blog and the users learn to search in other places then the official documentation. But here we can control the information and also enlighten the user to be caution to use the `do` command for example. I will probably use the solution @fdncred shared that I think is a better solution then the other that I found on Discord. So I will edit and remove the `do` part from the text as you suggest. 😄

Thank you @kubouch for the feedback.

kubouch:
I think it's OK now. It could be condensed more, but we can merge it. Docs improvement PRs are always welcome.

The alias-do method can be mentioned in the previous section because people keep trying it. So, we could have a short alias-do example followed by explaining that `def` is considered more idiomatic because you can add flags to it etc., unlike alias-do.

Note that you can one-line `def` as well:
```nushell
alias ls = do { ls | sort-by type name -i }
def ls [] { ls | sort-by type name -i }
```
It's actually shorter than the alias-do in this case!

kubouch:
Thanks!

zolodev:
Thank you!

I will try to add more documentation in the future. The more I learn the more I can contribute. Cheers!

zolodev:
> I think it's OK now. It could be condensed more, but we can merge it. Docs improvement PRs are always welcome.
> 
> The alias-do method can be mentioned in the previous section because people keep trying it. So, we could have a short alias-do example followed by explaining that `def` is considered more idiomatic because you can add flags to it etc., unlike alias-do.
> 
> Note that you can one-line `def` as well:
> 
> ```nushell
> alias ls = do { ls | sort-by type name -i }
> def ls [] { ls | sort-by type name -i }
> ```
> 
> It's actually shorter than the alias-do in this case!

Sorry for the late update @kubouch , but I have been busy with other things latley.
However when I run the "one-line `def` I get an `recursion_limit_reached` error.
```shell
Error: nu::shell::recursion_limit_reached

  × Recursion limit (50) reached
     ╭─[C:\Users\zolodev\AppData\Roaming\nushell\config.nu:798:1]
 798 │ #}
 799 │ def ls [] { ls | sort-by type name -i }
     ·           ──────────────┬──────────────
     ·                         ╰── This called itself too many times
 800 │
     ╰────
```

So it seems that it is not possible to "one-line" even with a `def`.
I just wanted to give you an update about this. :) Even if this PR is already merged.

kubouch:
Yeah, you're right, you'd need to alias `ls` beforehand or something. What I meant that you can write a one-liner `def`s in general and it's not longer than `alias` if it's very simple.

PR #1192, opened 2024-01-03
update the list of commands which are interpreted by CMD

yukitomoda:
Update the list of commands which are interpreted by CMD

Here is the current full list:

https://github.com/nushell/nushell/blob/49ab559992fa3c17a86a66c18b40fb8bb9c01859/crates/nu-command/src/system/run_external.rs#L205C3-L207


fdncred:
Thanks!

PR #1193, opened 2024-01-03
external completer: fixed cookbook example to exchange the first span entry only

aserowy:
Hi,

@rsteube and i integrated the cookbook example for alias expansion in carapace rsteube/carapace-bin#2102. We (mostly not me) found, that nushell expands the alias partially and returns for eg `alias gco = git checkout` a span with `[ "gc", "checkout" ]`. Thus, the cookbook example replaces `gc` with `git` and `checkout`.

To replace `gc` only, we added `take 1` to ensure the alias gets replaced with `git`.

Kind regards
Alexander

### DIFF
--- a/cookbook/external_completers.md
+++ b/cookbook/external_completers.md
@@ -96,7 +96,7 @@ let expanded_alias = (scope aliases | where name == $spans.0 | get -i 0 | get -i
 # overwrite
 let spans = (if $expanded_alias != null  {
     # put the first word of the expanded alias first in the span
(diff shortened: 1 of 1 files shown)


fdncred:
thanks

clavia-lang:
There are no errors in examples.

```nu
alias gco = git checkout
```

When we run `gco master`, the current logic is:
1. `skip 1`: removes the first element of the `spans` array, which is `gco`.
2. `take 1`: takes the first the element of the result `[git checkout]`, which is `git`.
3. The result should be `git check master`, but actually `git master`.

PR #1197, opened 2024-01-06
JA update shells_in_shells.md

yukitomoda:
Update the Japanese translation of shells_in_shells.md catching up with English ver.


fdncred:
thanks

PR #1206, opened 2024-01-11
Adding range related info.

### LINKED ISSUE #404: Specify type of Range
In https://www.nushell.sh/book/types_of_data.html#ranges

Replace "A range is a way of expressing a sequence of values from start to finish."

To "A range is a way of expressing a sequence of INTEGER values from start to finish."

### LINKED ISSUE #1191: Step value of range is not mentioned in document
Range can accept a step value like this

```nu
❯ 1..3..10
╭───┬───╮
│ 0 │ 1 │
│ 1 │ 3 │
│ 2 │ 5 │
│ 3 │ 7 │
│ 4 │ 9 │
╰───┴───╯
```

But this is not documented.
https://github.com/nushell/nushell.github.io/blob/5954009c85289cb4e9e6d2d209da1f44a582ec13/book/types_of_data.md?plain=1#L156

TWSiO:
Resolves #1191 and resolves #404

Added a section about the middle "step" range parameter, changed it to "sequence of integer or float values", added a tip about creating similar sequences for chars and dates, and added a warning about displaying open ended ranges.

Hopefully the range step section makes sense or at least isn't too confusing. I tried to be careful with how I presented it but it was a bit tricky.

---

After seeing that the range design is still in flux here nushell/nushell#11086, I considered holding off on adding this and waiting until it's more nailed down, however I figured increasing awareness of it might help us get more feedback about the current design. We can always change it when the time comes.

### DIFF
--- a/book/types_of_data.md
+++ b/book/types_of_data.md
@@ -155,7 +155,19 @@ true
 
 ## Ranges
 
-A range is a way of expressing a sequence of values from start to finish. They take the form \<start\>..\<end\>. For example, the range `1..3` means the numbers 1, 2, and 3.
+A range is a way of expressing a sequence of integer or float values from start to finish. They take the form \<start\>..\<end\>. For example, the range `1..3` means the numbers 1, 2, and 3.
+
+::: tip
+
+You can also easily create lists of characters with a form similar to ranges with the command [`seq char`](/commands/docs/seq_char.html) as well as with dates using the [`seq date`](/commands/docs/seq_date.html) command.
+
+:::
+
+### Specifying the step
+
+You can specify the step of a range with the form \<start\>..\<second\>..\<end\>, where the step between values in the range is the distance between the \<start\> and \<second\> values, which numerically is \<second\> - \<start\>. For example, the range `2..5..11` means the numbers 2, 5, 8, and 11 because the step is \<second\> - \<first\> = 5 - 2 = 3. The third value is 5 + 3 = 8 and the fourth value is 8 + 3 = 11.
+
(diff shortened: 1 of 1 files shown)


TWSiO:
I suppose the tip about `seq char` doesn't necessarily capture a char equivalent to a range which is more of an iterator than a list. Something like was mentioned here might be closer https://github.com/nushell/nushell.github.io/issues/404#issuecomment-1100711792.

Although, this section is covering the basic information about these types rather than getting into the nitty gritty about things like that it's sort of a number iterator, so maybe it's alright to have an imperfect equivalent that is still simple.

TWSiO:
Forgot to add some screenshots to show what the changes actually look like on the site and to provide something a bit more readable.

Screenshot_2024-01-11_17-31-07

---

Screenshot_2024-01-11_17-31-35

book/types_of_data.md
```diff
@@ -155,7 +155,19 @@ true
 
 ## Ranges
 
-A range is a way of expressing a sequence of values from start to finish. They take the form \<start\>..\<end\>. For example, the range `1..3` means the numbers 1, 2, and 3.
+A range is a way of expressing a sequence of integer or float values from start to finish. They take the form \<start\>..\<end\>. For example, the range `1..3` means the numbers 1, 2, and 3.
+
+::: tip
+
+You can also easily create lists of characters with a form similar to ranges with the command [`seq char`](/commands/docs/seq_char.html) as well as with dates using the [`seq date`](/commands/docs/seq_date.html) command.
```
fdncred, line 162: maybe mention `seq char` too?

fdncred (changes requested):
I think this will be a great help. I just asked for one small change. Thanks!

book/types_of_data.md
TWSiO: I think it does. Or did you mean it should provide an example or explain them or something?

fdncred: lol, i have no idea what i meant. sorry for the noise.

fdncred (approved):

fdncred:
Thanks.

PR #1208, opened 2024-01-13
explicit print as nu no longer automatically prints output of every line

### LINKED ISSUE #1163: Confusing example on script definition order
## Issue

I'm currently reading the book which [claims](https://www.nushell.sh/book/scripts.html) that 

> there is no requirement that definitions have to come before the parts of the script that call the definitions [...]

while showing these scripts as examples of how definition order does not matter:
```nushell
def greet [name] {
  ["hello" $name]
}

greet "world"
```
```nushell
greet "world"

def greet [name] {
  ["hello" $name]
}
```
This choice of examples is a bit unfortunate in my opinion as, when running both scripts, they have different output.
My understanding is that this is caused by the last expression in a script being printed which evaluates to `["hello" "world"]` and `nothing` respectively.
However, this confused me initially and I would expect this to trip up some other new users as well.

## Possible improvements

1. Changing the script to e.g.
```nushell
def greet [name] {
  ["hello" $name]
}
greet "world" | null
```
or
```nushell
def greet [name] {
  ["hello" $name]
}
let foo = greet "world"
$foo
```
would result in the same output when moving the command invocation line around but might raise other questions.
```nushell
def greet [name] {
  ["hello" $name]
}
def greet-world [] {
  greet "world"
}
greet-world
```
and then moving the definition of `greet-world` to the top is probably the best example I can think of right now.

2. Alternatively, the different output could be explained in a small note. I'm not even sure sure if the book explains the last expression of a script being printed up to that point. It's definitely explained in a note for functions but I don't remember reading that the same applies for scripts. I wouldn't rule out that I simply missed that part, though.

Inveracity:
# Change

This PR updates the script examples to explicity `print` output.

# Reason

As per this comment https://github.com/nushell/nushell.github.io/issues/1163#issuecomment-1841613415, Nu no longer automatically print output of every line.

closes #1163

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -17,7 +17,7 @@ Let's look at an example script file:
 ```nu
 # myscript.nu
 def greet [name] {
-  ["hello" $name]
+  print ["hello" $name]
 }
 
 greet "world"
@@ -31,7 +31,7 @@ In the above, first `greet` is defined by the Nushell interpreter. This allows u
 greet "world"
 
 def greet [name] {
-  ["hello" $name]
+  print ["hello" $name]
 }
 ```
 


fdncred:
I'm not sure it works like that now. This seems to work fine.
```nushell
def greet [name] {
  ["hello" $name]
}
```
```
❯ greet moon
╭───┬───────╮
│ 0 │ hello │
│ 1 │ moon  │
╰───┴───────╯
```
However, this is really how I think it's meant to work currently.
```nushell
def greet2 [name] {
  ["hello" $name]
  $name
}
```
```
❯ greet2 moon
moon
```

You'll notice that `["hello" $name]` doesn't print out in greet2. This is because we only return/print the last item.

But if you try greet3, you'll see that both print again.
```nushell
def greet3 [name] {
  echo ["hello" $name]
  echo $name
}
```
```
❯ greet2 moon
╭───┬───────╮
│ 0 │ hello │
│ 1 │ moon  │
╰───┴───────╯

moon
```

Inveracity:
Ah I see

When I initially started out writing functions I called them multiple times in a script and was confused as to why I only got the output of the last function call

```nu
def greet [name] {
  ["hello" $name]
}

greet "one"
greet "two"
```
only outputs
```
╭───┬───────╮
│ 0 │ hello │
│ 1 │ two   │
╰───┴───────╯
```
without the explicit `print`

It wasn't until I found that comment I linked to, that it made sense.

I can close this if it's not relevant to have the explicit `print`, but personally I think it would help some newcomers as I ran head first into this 😄

Inveracity:
It literally just struck me that it's an implicit return 🤦 now I understand. Started learning Nu yesterday haha

PR #1211, opened 2024-01-16
Fix deprecation in external_completers.md

flying-sheep:
See https://github.com/nushell/nushell/pull/11547


fdncred:
Thanks again!

PR #1212, opened 2024-01-16
JA Translate quick_tour.md to ja from commit: fcf0d7c8

yukitomoda:
Add a Japanese translation of quick_tour.md


hustcer:
Thanks for continuing to update and advance the translation of the Japanese version docs!

PR #1213, opened 2024-01-18
fix(exec.md): remove `{flags}` since there is no flag for exec

qdm12:
- there is no flag for `exec`
- it's confusing since it could mean flags to the command, which is not the case

### DIFF
--- a/commands/docs/exec.md
+++ b/commands/docs/exec.md
@@ -17,7 +17,7 @@ feature: default
 
 ## Signature
 
-```> exec {flags} (command)```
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for trying to help. This CI is red because commands can't be modified in our help since they're autogenerated from the command itself. If changes need to happen to commands, it needs to happen in the rust code.

qdm12:
That's awesome, I'll have a look in the code (probably tomorrow)

PR #1219, opened 2024-01-20
Remove outdated commands from explore.md

samoylovfp:
commands
```
:config
:show-config
:tweak
```
were removed 
in https://github.com/nushell/nushell/pull/10259

keeping the documentation up to date


fdncred:
Good catch. Thanks!

PR #1220, opened 2024-01-20
Update `scope`'s syntax in external_completers.md

SLASHLogin:
Update `scope`'s syntax to reflect current syntax and match command from above the file.


fdncred:
Thanks!

PR #1222, opened 2024-01-21
Adding a page on control flow (and small misspelling)

### LINKED ISSUE #1207: A page on control flow
Looking through the book I didn't see anywhere that introduced `if` or most of the other control flow commands.

I suggest adding a page about control flow commands.

I think most people might not consider looking in the commands section about them since I think most people think of them as "part of the language" or more primitive "special forms" like range or the operators that wouldn't be included in the commands which people might be thinking of as "built in functions" (since I think control flow is presented separately in a lot of other languages as well). Plus it's also just helpful to gather up commands with similar use cases and letting people know what their names are so they don't have to go searching through the list of command names for the right command.

The page can also cover most of the error functionality mentioned in #772.

Here's a list of commands I could think to include.

- `if`
- `match`
  - At least a basic brief summary. Can point to the cookbook pattern matching page for more details.
- `for`
- `while`
- `loop`
- `break`
- `continue`
- `error make`
- `try`
- `return`

TWSiO:
Resolves #1207

As the title says, it adds a page on control flow in Nu (and fixes a small misspelling on the pattern matching cookbook page I spotted while working on this).

The PR is in draft because I want to read through it again tomorrow with fresh eyes, but otherwise I think it's ready for review.

<details>
<summary>Long screenshot of the page</summary>

Control Flow Nushell

</details>

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -43,6 +43,7 @@ export const sidebarEn: SidebarConfig = {
         '/book/aliases.md',
         '/book/operators.md',
         '/book/variables_and_subexpressions.md',
+        '/book/control_flow.md',
         '/book/scripts.md',
         '/book/modules.md',
         '/book/overlays.md',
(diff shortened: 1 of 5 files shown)


TWSiO:
Made a couple small changes, and it could probably use a few more, but I'll call it there and submit it for review.

<details>
<summary>Long screenshot of the page</summary>

Screenshot Control Flow Nushell

</details>

fdncred:
great adds!

PR #1224, opened 2024-01-21
Add a section describing how to change table outputs

crasite:
I've been trying out Nu for a while. Some problem I have is that Nu will try to expand the table structure unless I pipe 
it with `table` command. I've looked into [Working with table](https://www.nushell.sh/book/working_with_tables.html) and [Configuration](https://www.nushell.sh/book/configuration.html) on how to change this default behavior without success. I found a solution by accidently finding this nushell/nushell#7292 issues. So I think that we should have some documentation explaining how to configure it more explicitly.


fdncred:
Looks right to me. Thanks!

PR #1226, opened 2024-01-22
add env fallback example

c4lliope:


fdncred:
Maybe this is what you're trying to do? This gets the $env.FOO variable if it exists and if it doesn't returns ABC.
```nushell
$env.FOO? | default "ABC"
```

c4lliope:
Yeah! This looks good.
Is this the recommended approach for default arguments also, or can you pass a default in the argument braces...
```nu
def loud-echo [phrase:string (DEFAULT?) ] { echo $"Yo! ($phrase)" }
```

fdncred:
I'm not sure the `optional` operator `?` works with variables.

PR #1227, opened 2024-01-23
Add note about string interpolation at parse time

ysthakur:
Since string interpolation at parse time has been merged (https://github.com/nushell/nushell/pull/11562) but it has unexpected behavior for formatting file sizes and datetimes if used before config.nu is loaded (the default config is used instead of your runtime config), this PR adds a note about that to the "Working with strings" page.

The docs do say [here](https://www.nushell.sh/book/how_nushell_code_gets_run.html) that sourcing a *dynamic* path like `source $"($my_path)/common.nu"` isn't possible, but this may be interpreted as meaning that stuff like `const x = "foo"; source $"($x)"` doesn't work too. There are a few different ways (not mutually exclusive) to avoid confusion:
- Adding `let my_path = ...` to show that `my_path` isn't a constant
- Amending "Sourcing a dynamic path" to something like "Sourcing a dynamic path (constants are fine, see \<some section\>)"
- Adding examples (like fdncred's screenshots [here](https://github.com/nushell/nushell/pull/11562#issuecomment-1904254075)) of using string interpolation at parse time for sourcing
- Something else?

### DIFF
--- a/book/how_nushell_code_gets_run.md
+++ b/book/how_nushell_code_gets_run.md
@@ -149,7 +149,7 @@ let my_path = 'foo'
 source $"($my_path)/common.nu"
 ```
 
-Let's break down what would need to happen for this to work (assuming `$my_path` is set somewhere):
+Let's break down what would need to happen for this to work:
 
 1. Parse `let my_path = 'foo'` and `source $"($my_path)/config.nu"`
 2. To evaluate `source $"($my_path)/common.nu"`:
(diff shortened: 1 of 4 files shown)


fdncred:
I agree that I think all the examples you mentioned above should be mentioned in the docs.

ysthakur:
@fdncred Just updated the How Nushell Code Gets Run and Thinking in Nu pages, let me know if that's good

fdncred:
Thanks. Looks good!

PR #1232, opened 2024-01-30
Update custom_completions.md to use new spread operator in carapace example

mmastrocinque:
Noticed I was getting an error after updating nushell warning me that auto spreading lists would be deprecated, updated my configuration and thought this change could be reflected in the docs.
image


fdncred:
Thanks

PR #1235, opened 2024-02-01
Update query-web's documentation

kik4444:
Reflects changes in https://github.com/nushell/nushell/pull/11705


fdncred:
Thanks!

fdncred:
oh, ya. this can't be changed because they're systematically generated by a script. so after the next release this will get update.

PR #1240, opened 2024-02-04
fill contents to handling stderr

WindSoilder:
This is a follow up to https://github.com/nushell/nushell/pull/11708

Which introduces all rules to redirections and pipes

### DIFF
--- a/lang-guide/lang-guide.md
+++ b/lang-guide/lang-guide.md
@@ -456,7 +456,7 @@ It doesn't work because it sees `\P` and `\s` as escapes that are not recognized
 
 ### Interaction with unix pipes
 
-### Handling stderr
+### Handling stdout and stderr
 You can do the following things to stderr:
 1. do nothing, it will be output to stderr directly
 2. pipeline stderr to next commands, using `e>|` or `e+o>|`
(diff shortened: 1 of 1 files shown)


kubouch:
Looks good. We might want to split the document up at some point, but that could be done later. 

The examples with `nu --testbin echo_env_mixed out-err FOO BAR` might not be clear. Even though it's defined in the code block at the top, maybe it would be clearer to wrap it in a custom command, like `def print-to-stdout-stderr [] { nu --testbin ... }`.

One more thing we could do is change the title from `Handling stderr` to `Handling stdout and stderr` because both are being discussed. And it could be mentioned at the top that it works only for external streams because Nushell doesn't have stdout/stderr.

I like the style overall, mostly tables and short examples. Some minor grammar things that we can check once it's ready.

WindSoilder:
> The examples with nu --testbin echo_env_mixed out-err FOO BAR might not be clear. Even though it's defined in the code block at the top, maybe it would be clearer to wrap it in a custom command, like def print-to-stdout-stderr [] { nu --testbin ... }

Sadly I think `print-to-stdout-stderr` won't work, because it's not external command.  I've update the example to use a script called `demo.nu`, then we can interactive with the script.

kubouch:
@WindSoilder Looks good to me. I updated some language and also made the tables more structured. If you're fine with that, feel free to merge. Thanks for clarifying the different redirection styles, it's now much cleaner!

WindSoilder:
Thank you for the change!

PR #1245, opened 2024-02-06
Escape false <int> HTML element

IanManske:
Other whitespace changes are due to one of the git hooks I think.


kubouch:
Nice catch, thanks!

PR #1247, opened 2024-02-07
Fix typo in `into cell-path` in the 0.90.1 release notes

abusch:
The release notes for `0.90.1` mention a new `into cellpath` command, but it is actually called `into cell-path`.


hustcer:
Good catch, Thanks

PR #1251, opened 2024-02-11
docs: start rework data types and operator

vwkd:
Start to rework sections on records, lists and tables in data types, and sections cell path access and spread operator in operators.

The main theme is to clarify the wording and examples used in order to aid comprehension. By being more concise and saying only what's needed, we can remove complexity that's impeding comprehension.

The commits are self-contained and have a description with more detail on the reasoning for each change, such that they can be reviewed individually.

Because fundamental changes to prose like moving and rewording can make reading the line-by-line diff difficult, I'd recommend to glance at the diff only for seeing which sections changed, and then read the rendered new version as well as the rendered old version to feel which one reads better.

### DIFF
--- a/book/operators.md
+++ b/book/operators.md
@@ -123,171 +123,43 @@ Operators are usually case-sensitive when operating on strings. There are a few 
 
 ## Spread operator
 
-Nushell has a spread operator (`...`) for unpacking lists and records. You may be familiar with it
-if you've used JavaScript before. Some languages use `*` for their spread/splat operator. It can
-expand lists or records in places where multiple values or key-value pairs are expected.
+The spread operator (`...`) allows to expand lists and records.
 
-There are three places you can use the spread operator:
+lists in a list literal or records in a record literal, as well as expand a list into multiple arguments in a command call.
 
-- [In list literals](#in-list-literals)
-- [In record literals](#in-record-literals)
-- [In command calls](#in-command-calls)
+### Creating lists and records
 
-### In list literals
+When creating a list literal, the spread operator can expand the values of another list. For example, to insert a list into another list:
 
-Suppose you have multiple lists you want to concatenate together, but you also want to intersperse
-some individual values. This can be done with `append` and `prepend`, but the spread
-operator can let you do it more easily.
-
-```nushell
-> let dogs = [Spot, Teddy, Tommy]
-> let cats = ["Mr. Humphrey Montgomery", Kitten]
-> [
-    ...$dogs
-    Polly
-    ...($cats | each { |it| $"($it) \(cat\)" })
-    ...[Porky Bessie]
-    ...Nemo
-  ]
-╭───┬───────────────────────────────╮
-│ 0 │ Spot                          │
-│ 1 │ Teddy                         │
-│ 2 │ Tommy                         │
-│ 3 │ Polly                         │
-│ 4 │ Mr. Humphrey Montgomery (cat) │
-│ 5 │ Kitten (cat)                  │
-│ 6 │ Porky                         │
-│ 7 │ Bessie                        │
-│ 8 │ ...Nemo                       │
-╰───┴───────────────────────────────╯
+```nu
+> [sam, ...[fred, george], susan]
+╭───┬────────╮
+│ 0 │ sam    │
+│ 1 │ fred   │
+│ 2 │ george │
+│ 3 │ susan  │
+╰───┴────────╯
 ```
 
-The below code is an equivalent version using `append`:
-```nushell
-> $dogs |
-    append Polly |
-    append ($cats | each { |it| $"($it) \(cat\)" }) |
-    append [Porky Bessie] |
-    append ...Nemo
+This is shorter than to chain multiple `append` or `prepend` commans for each list part. Also, since only one new list is created, it might be slightly more performant.
+
+When creating a record literal, the spread operator can expand the values of another record. For example, to insert a record into another record:
+
+```nu
+> { ...{ name: sam, rank: 10 }, age: 21 }
+╭──────┬─────╮
+│ name │ sam │
+│ rank │ 10  │
+│ age  │ 21  │
+╰──────┴─────╯
 ```
 
-Note that each call to `append` results in the creation of a new list, meaning that in this second
-example, 3 unnecessary intermediate lists are created. This is not the case with the spread operator,
-so there may be (very minor) performance benefits to using `...` if you're joining lots of large
-lists together, over and over.
+### Multiple command arguments
 
-You may have noticed that the last item of the resulting list above is `"...Nemo"`. This is because
-inside list literals, it can only be used to spread lists, not strings. As such, inside list literals, it can
-only be used before variables (`...$foo`), subexpressions (`...(foo)`), and list literals (`...[foo]`).
-
-The `...` also won't be recognized as the spread operator if there's any whitespace between it and
-the next expression:
+When calling a command with a rest parameter or an external command, the spread operator can expand the values of a list of arguments.
 
 ```nushell
-> [ ... [] ]
-╭───┬────────────────╮
-│ 0 │ ...            │
-│ 1 │ [list 0 items] │
-╰───┴────────────────╯
-```
-
-This is mainly so that `...` won't be confused for the spread operator in commands such as `mv ... $dir`.
-
-### In record literals
-
-Let's say you have a record with some configuration information and you want to add more fields to
-this record:
-
-```nushell
-> let config = { path: /tmp, limit: 5 }
-```
-
-You can make a new record with all the fields of `$config` and some new additions using the spread
-operator. You can use the spread multiple records inside a single record literal.
-
-```nushell
-> {
-    ...$config,
-    users: [alice bob],
-    ...{ url: example.com },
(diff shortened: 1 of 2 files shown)


book/types_of_data.md
```diff
@@ -336,24 +336,22 @@ To append one or more lists together, optionally with values interspersed in bet
...
+│ 1 │ bob  │    7 │
+╰───┴──────┴──────╯
 ```
 
-You can also create a table as a list of records, JSON-style:
+A row can also be written as list of values. Then, the first element in the list is a list of column names and separated by a semicolon.
 
 ```nu
-> [{name: sam, rank: 10}, {name: bob, rank: 7}]
+> [[name, rank]; [sam, 10] [bob, 7]]
```
fdncred, line 335: This is called `table literal` syntax and probably should be mentioned.

book/types_of_data.md
```diff
@@ -336,24 +336,22 @@ To append one or more lists together, optionally with values interspersed in bet
...
 
 ```nu
-> [[column1, column2]; [Value1, Value2] [Value3, Value4]]
-╭───┬─────────┬─────────╮
-│ # │ column1 │ column2 │
-├───┼─────────┼─────────┤
-│ 0 │ Value1  │ Value2  │
-│ 1 │ Value3  │ Value4  │
-╰───┴─────────┴─────────╯
+> [{name: sam, rank: 10}, {name: bob, rank: 7}]
```
fdncred, line 323: This is called `table literal` syntax with records.

book/types_of_data.md
```diff
@@ -362,29 +360,83 @@ You can also create a table as a list of records, JSON-style:
 ╰───┴──────┴──────╯
 ```
 
-:::tip
-Internally, tables are simply **lists of records**. This means that any command which extracts or isolates a specific row of a table will produce a record. For example, `get 0`, when used on a list, extracts the first value. But when used on a table (a list of records), it extracts a record:
+To access a row or a column, you can use the [`get`](/commands/docs/get.md) command or cell path access:
```
fdncred, line 344: You're missing the primary explanation of what `get` does here versus `select`. `get` gets data out of its shape and transforms the data. `select` maintains the shape of the data. Say that however you think make sense, but it's a fundamental difference between `get` and `select`.

book/types_of_data.md
```diff
@@ -362,29 +360,83 @@ You can also create a table as a list of records, JSON-style:
...
+╰───┴─────╯
+> [{name: sam, rank: 10}, {name: bob, rank: 7}].name
+╭───┬─────╮
+│ 0 │ sam │
+│ 1 │ bob │
+╰───┴─────╯
 ```
 
-This is true regardless of which table syntax you use:
+Note, the records in a table don't necessarily need to have matching keys. If a record lacks a key, the table will lack a value for that column. You can think of the table having a hole.
```
fdncred, line 369: This is true but it may be a good place to talk about the difference between these items in tables.
- absence of a value
- empty string
- null value

book/types_of_data.md
fdncred, line 344: You may also want to mention that `get <nth>` gets a row whereas `get <name>` gets a column. Technically, you can have columns named <nth> as well but when you `get` them you have to enclose them in quotes. e.g. `get "5"` would get the column named `"5"` instead of the row named `5`. Columns are always strings; rows are always numbers.

book/types_of_data.md
```diff
@@ -277,41 +277,42 @@ To make a copy of a record with new fields, you can use the [spread operator](/b
 
 ## Lists
 
-Lists are ordered sequences of data values. List syntax is very similar to arrays in JSON. However, commas are _not_ required to separate values if Nushell can easily distinguish them!
+A list is an ordered sequence of values. Unlike most languages, comma separators are optional.
 
 ```nu
-> [sam fred george]
+> [sam, fred, george]
```
fdncred, line 263: Commas are optional, yet you show them in the example. I'd rather not show them here.

book/types_of_data.md
fdncred, line 323: Not related to this specific line but one thing that may be missing is that data in tables is currently not restricted to a type per column. You could theoretically have every nushell value in one column. Some systems like databases and dataframes require columnar data to be a single type.

book/types_of_data.md
fdncred, line 263: This is called `list literal` syntax.

book/types_of_data.md
```diff
@@ -209,35 +209,45 @@ Structured data builds from the simple data. For example, instead of a single in
 
 ## Records
 
-Records hold key-value pairs, which associate string keys with various data values. Record syntax is very similar to objects in JSON. However, commas are _not_ required to separate values if Nushell can easily distinguish them!
+A record is a collection of key-value pairs. Unlike most languages, comma separators are optional.
 
 ```nu
-> {name: sam rank: 10}
+> {name: sam, rank: 10}
```
fdncred, line 211: This is called `record literal` syntax.

fdncred, line 211: I prefer no commas in examples. Showing them tries to take away one of the superpowers of nushell, IMO.

book/types_of_data.md
fdncred, line 323: I'll mention this in other commits too, but my personal preference is to not have commas in examples. It's really a superpower of nushell to not require commas and also not require quotes (this is a bare word).

fdncred:
I think most of this is fine but brief. It almost sounds more like it belongs in the language reference guide. The language reference guide is meant to be short and succinct. https://www.nushell.sh/lang-guide/lang-guide.html

book/types_of_data.md
```diff
@@ -203,10 +203,6 @@ You can write binary as a literal using any of the `0x[...]`, `0b[...]`, or `0o[
 
 Incomplete bytes will be left-padded with zeros.
 
-## Structured data
```
fdncred, line 206: I'd rather see this left in or rewritten. A founding principle of nushell is working with structured data. If it's not defined anywhere it makes it more difficult to understand. We refer to structured data in discord all the time.

book/operators.md
```diff
@@ -289,3 +289,53 @@ as being omitted:
...
+```nu
+> [sam, fred, george].1
+fred
+> {name: sam, rank: 10}.name
+sam
+```
+
+:::tip
+
+If the key isn't a bare string, it needs quotes:
```
fdncred, line 178: we use the term "bare word" and it probably should be defined before using it.

book/operators.md
```diff
@@ -289,3 +289,53 @@ as being omitted:
 > foo "bar" ...[1 2] "not opt" # The null means no argument was given for opt
 [false, bar, null, [1, 2, "not opt"]]
 ```
+
+## Cell path access
+
+To access a value in a list or record in the output of a command, we can pipe it into `get` and specify the index or key. However, if we have a literal or variable this would often require a separate subexpression. Instead, we can use cell path access like this:
+
+```nu
+> [sam, fred, george].1
```
fdncred, line 170: Would rather have examples without commas.

book/operators.md
```diff
@@ -289,3 +289,53 @@ as being omitted:
 > foo "bar" ...[1 2] "not opt" # The null means no argument was given for opt
 [false, bar, null, [1, 2, "not opt"]]
 ```
+
+## Cell path access
```
fdncred, line 165: I didn't see any mention of `cell-path literals`. We should probably mention it with an example.

```nu
let cp = $.1.name
[{name: sam rank: 10} {name: bob rank: 7}] | get $cp
```

book/operators.md
```diff
@@ -123,171 +123,43 @@ Operators are usually case-sensitive when operating on strings. There are a few
 
 ## Spread operator
 
-Nushell has a spread operator (`...`) for unpacking lists and records. You may be familiar with it
-if you've used JavaScript before. Some languages use `*` for their spread/splat operator. It can
-expand lists or records in places where multiple values or key-value pairs are expected.
+The spread operator (`...`) allows to expand lists and records.
```
fdncred, line 126: "allows to expand" doesn't sound proper

book/operators.md
```diff
@@ -123,171 +123,43 @@ Operators are usually case-sensitive when operating on strings. There are a few
...
 
-- [In list literals](#in-list-literals)
-- [In record literals](#in-record-literals)
-- [In command calls](#in-command-calls)
+### Creating lists and records
 
-### In list literals
+When creating a list literal, the spread operator can expand the values of another list. For example, to insert a list into another list:
 
-Suppose you have multiple lists you want to concatenate together, but you also want to intersperse
```
fdncred, line 138: I'm not a fan of removing this huge section here. I'd rather have more examples than less. It's nice to be brief and concise but if it doesn't explain the nuances of what it does and how it works, it's not necessarily better.

fdncred (changes requested):
Most of this sounds good. Thank you for the effort to make our docs clearer and easier to understand. However, in some places I have requested changes and some changes I just didn't like.

I did not review this as you suggested so I could've missed something too. I just reviewed all the commits, one at a time and made comments on each commit by looking at the diff and rendered content.

fdncred:
@vwkd not interested in making changes?

fdncred:
no response

PR #1252, opened 2024-02-13
Fix regex section

J0J0:
Currently, the docs say

> Regular expressions in Nushell's commands are handled by the rust-lang/regex crate.

-- which is not correct as of https://github.com/nushell/nushell/commit/cdeb8de75d69b76c9120a53aa8e1814d388d3f0e .

In particular, `help parse` suggests that one can use "fancy-regex look behind pattern", which got me terribly confused (until visiting the source code ...).

Regarding the commands that use (only) `regex`: In my opinion, the proper fix would be to upgrade them to `fancy-regex` as well, but unfortunately i don't know much Rust, so updating the docs is the best, i can offer. :/


fdncred:
We need to standardize on fancy-regex since it offers more I think. I tried to do this and it's either been reversed or I didn't catch all the places.

PR #1263, opened 2024-02-27
Run typos on blog

IanManske:
Fixes typos on the blog and other parts of the repository. It will also make the CI run `typos` on the blog and most of the repository in the future. Since there are some remaining false positives in the blog, these are ignored in the typos config through the glob `"blog/202[0-3]*"` instead of adding each false typo to the list of ignored words. The other whitespace changes in the PR are due to `prettier` and the git hook.


fdncred:
Thanks

PR #1265, opened 2024-02-27
Fix issues with spans in the JSON examples in the plugin protocol reference

devyn:
Noticed a few things that I didn't catch before after reading it again. `"internal_span"` was renamed to `"span"` in the serialized format but that isn't reflected in some of the JSON examples

And there was also one JSON example that was invalid because it didn't include the `"start"` and `"end"` keys


sholderbach:
Thanks for all the work on the plugins!

PR #1268, opened 2024-02-29
Add an example of adding multiple paths

jordanst3wart:
Hi,

I just wasted a few hours looking for a nice way to add multiple paths to `$env.path`. I came up with this, which I quiet like:

```nu
let extra_paths = ['/some/path', '/another/path']

$env.PATH = (($env.PATH | split row (char esep)) ++ $extra_paths)
```

The double `((` might not be the best so feel free to edit.

I think the standard use case of adding values to the `$env.PATH` is "I want to add a list of values to `$env.PATH`", and not "I want to add a single path to my `$env.PATH`". So I think this example is ideal for new users like myself.

I also think having duplicate paths in `$env.PATH`, is not the right thing to be worried about. So I removed that example, to keep things simple.

### DIFF
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -157,7 +157,7 @@ $env.PATH = ($env.PATH |
 $env.PATH = ($env.PATH | uniq)
 ```
 
-This will add `/some/path`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.
+This will add `/usr/local/bin`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.
 
 ### Homebrew
 
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -144,13 +144,20 @@ This will append `/some/path` to the end of PATH; you can also use [`prepend`](/
 
 Note the `split row (char esep)` step. We need to add it because in `env.nu`, the environment variables inherited from the host process are still strings. The conversion step of environment variables to Nushell values happens after reading the config files (see also the [Environment](environment.html#environment-variable-conversions) section). After that, for example in the Nushell REPL when `PATH`/`Path` is a list , you can use [`append`](/commands/docs/append.md)/[`prepend`](/commands/docs/prepend.md) directly.
 
-To prepend a new path only if not already listed, one can add to `env.nu`:
+To add multiple paths only if not already listed, one can add to `env.nu`:
+
 ```nu
-# create a new string holding the desired path
-let my_path = ( $nu.home-path | path join "bin" )
-# return $env.PATH if $my_path is already listed, return $env.PATH with $my_path prepended otherwise
-$env.PATH = ( if $my_path in $env.PATH { $env.PATH } else { $env.PATH | prepend $my_path } )
+$env.PATH = ($env.PATH | 
+    split row (char esep) |
+    append /usr/local/bin |
+    append ($env.CARGO_HOME | path join "bin") |
+    append ($env.HOME | path join ".local" "bin")
+)
+# filter so the paths are unique
+$env.PATH = ($env.PATH | uniq)
 ```
+
+This will add `/some/path`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.
 
 ### Homebrew
 


fdncred:
This is how I modify the path
```nushell
$env.PATH = ($env.PATH | 
    split row (char esep) |
    prepend /opt/homebrew/opt/sqlite/bin |
    # prepend /opt/homebrew/opt/coreutils/libexec/gnubin |
    prepend /opt/homebrew/opt/uutils-coreutils/libexec/uubin |
    prepend /opt/homebrew/bin |
    prepend /opt/homebrew/sbin |
    prepend /Users/fdncred/.cargo/bin |
    prepend /opt/homebrew/opt/python/libexec/bin |
    prepend /usr/local/bin |
    prepend /Users/fdncred/.local/bin |
    prepend /Users/fdncred/src/zig
    # append /Applications/CrystalFetch.app/Contents/MacOS
)
```
This is how amtoine sets the path
https://github.com/amtoine/dotfiles/blob/c893248d7e59ad297ed441d91a0f949fb6d74235/.config/nushell/env.nu#L128-L154

jordanst3wart:
Changed the documentation to include:

To add multiple paths only if not already listed, one can add to `env.nu`:

```nu
use std "path add"
$env.PATH = ($env.PATH | split row (char esep))
path add /some/path
path add ($env.CARGO_HOME | path join "bin")
path add ($env.HOME | path join ".local" "bin")
$env.PATH = ($env.PATH | uniq)
```

This will add `/some/path`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.

fdncred:
I'm not sure the purpose of removing what was already there. I believe it still works. We can probably add yours as an alternative way though.

jordanst3wart:
I think amtoine's way is better than the sample in the documents. I just added the standard `split row (char esep)`.

With this example in the document:

```nu
# create a new string holding the desired path
let my_path = ( $nu.home-path | path join "bin" )
# return $env.PATH if $my_path is already listed, return $env.PATH with $my_path prepended otherwise
$env.PATH = ( if $my_path in $env.PATH { $env.PATH } else { $env.PATH | prepend $my_path } )
```

It doesn't show an absolute path example, or that multiple values can be joined (ie `path join ".local" "bin"`). The`if` statement checking for unique paths is a lot longer than `$env.PATH = ($env.PATH | uniq)`.


As an unaware new user of nushell, just copying and paste the example, I could create my config as:

```nu
let local_bin = ( $nu.home-path | path join "/.local/bin" )
let usr_bin = ( $nu.home-path | path join "../../../usr/bin" )

$env.PATH = ( if $local_bin in $env.PATH { $env.PATH } else { $env.PATH | prepend $local_bin } )
$env.PATH = ( if $usr_bin in $env.PATH { $env.PATH } else { $env.PATH | prepend $usr_bin } )
```

That has a lot of boiler plate, and just isn't nice.

fdncred:
I'm fine with both but if I were to pick one, I'd want one that doesn't require another script like `path add` does.

jordanst3wart:
yeah, great point

jordanst3wart:
I changed it to:

```nu
$env.PATH = ($env.PATH | 
    split row (char esep) |
    append /usr/local/bin |
    append ($env.CARGO_HOME | path join "bin") |
    append ($env.HOME | path join ".local" "bin")
)
# filter so the paths are unique
$env.PATH = ($env.PATH | uniq)
```

So it's got an example of an absolute path, a relative path, joining directories on a relative path, and filtering out duplicate paths. I changed `prepend` to `append` as well

book/configuration.md
```diff
@@ -144,14 +144,21 @@ This will append `/some/path` to the end of PATH; you can also use [`prepend`](/
...
+    split row (char esep) |
+    append /usr/local/bin |
+    append ($env.CARGO_HOME | path join "bin") |
+    append ($env.HOME | path join ".local" "bin")
+)
+# filter so the paths are unique
+$env.PATH = ($env.PATH | uniq)
 ```
 
+This will add `/some/path`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.
```
fdncred, line 160:
```suggestion
This will add `/usr/local/bin`, the `bin` directory of CARGO_HOME, the `.local/bin` of HOME to PATH. It will also remove duplicates from PATH.
```

fdncred:
Thanks.

PR #1260, opened 2024-02-24
Add pyenv path setup to configuration.md

JackTheMico:
Hi there, I added the pyenv path setup to the configuration.md.

Here's a asciinema recording to show how it work.
[asciicast](https://asciinema.org/a/636181)

PR #1269, opened 2024-02-29
Remove pyenv setup instructions

jordanst3wart:
Hi,

I was a little confused why setup instructions for `pyenv`, a python version manager was included here.

So I removed it and created a PR for `pyenv` to include it in there setup here:
https://github.com/pyenv/pyenv/pull/2916


fdncred:
Maybe you should work together with the author https://github.com/nushell/nushell.github.io/pull/1260 and figure out what the best place to add it is.

PR #1275, opened 2024-03-03
sample of string interpolation and evaluation inside a string for echo command

jordanst3wart:
I thought I'll add an example of string interpolation on the echo command page, as it's a common use case coming from other shells, ie. bash/zsh.

### DIFF
--- a/commands/docs/echo.md
+++ b/commands/docs/echo.md
@@ -49,6 +49,16 @@ Returns the piped-in value, by using the special $in variable to obtain it.
 
 ```
 
+To do string interpolation, and evaluation inside a string:
+```nu
+> let name = "Alice"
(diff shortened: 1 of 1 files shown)


ysthakur:
The text "string interpolation" could maybe link to the [docs](https://www.nushell.sh/book/working_with_strings.html#string-interpolation) on string interpolation in case people want more in-depth information on how it works

jordanst3wart:
Yeah, that a great idea ysthakur

jordanst3wart:
reference: https://github.com/nushell/nushell.github.io/pull/1275

fdncred:
Command help text isn't allowed to be changed since it's systematically generated with each release.

PR #1284, opened 2024-03-06
update text to replace `std testing` with `testing.nu`

maxim-uvarov:
The chapter about unit testing is incorrect, as `std testing` is deprecated. I've made small changes that look logical to me. 

Probably, this chapter needs a foreword about the depreciation of the previous version of std and the current status. However, I don't have such details in mind.

Please, feel free to edit or let me know what I should correct anything. Thanks.


fdncred:
This is probably good for now but could change in the future.

PR #1288, opened 2024-03-07
Fix a bug in zoxide completer

shenjiangqiu:
should using ...$in in extern command. other wise will cause an error and no result will be returned.


fdncred:
yup, that's right. it recently changed. thanks.

PR #1292, opened 2024-03-09
Add cookbook chapter on how to read env vars from foreign shell scripts

remmycat:
This originated from a conversation in the Nushell Discord, where I showed a less sophisticated version of this script in the #cool-scripts channel, and was asked to contribute it to the Cookbook.

Next to reviewing the code, please give your honest opinions on the text, and if you want me to change anything, I appreciate the feedback 🙂 

You may of course also just edit anything yourself if that is preferred.

I hope this is helpful to people!

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -137,6 +137,7 @@ export const sidebarEn: SidebarConfig = {
         'help',
         'system',
         'parsing',
+        'foreign_shell_scripts',
         'pattern_matching',
         'external_completers',
         'files',
--- a/cookbook/foreign_shell_scripts.md
+++ b/cookbook/foreign_shell_scripts.md
@@ -1,0 +1,241 @@
+---
+title: Foreign Shell Scripts
+---
+
+# Working With Foreign Shell Scripts
+
+A common issue with nu is, that other applications export environment variables
+or functionality as shell scripts, that are expected to then be evaluated by
+your shell.
+
+But many applications only consider the most commonly used shells like `bash` or
+`zsh`. Unfortunately, nu has entirely incompatible syntax with these shells, so
+it cannot run or `source` these scripts directly.
+
+Generally nothing stops you from running a `zsh` script by invoking `zsh` itself
+(given it is installed). But unfortunately this will not allow nu to access
+exported environment variables:
+
+```nu
+# This works, using zsh to print "Hello"
+'echo Hello' | zsh -c $in
+
+# This exits with an error because $env.VAR is not defined
+'export VAR="Hello"' | zsh -c $in
+print $env.VAR
+```
+
+This chapter presents two workarounds for getting around this issue, and the
+involved drawbacks.
+
+---
+
+## Parsing a Script as a String
+
+A naive workaround to extract environment variable declarations is to read the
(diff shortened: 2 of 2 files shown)


.vuepress/configs/sidebar/en.ts
```diff
@@ -137,6 +137,7 @@ export const sidebarEn: SidebarConfig = {
         'help',
         'system',
         'parsing',
+        'foreign_shell_scripts',
```
remmycat, line 140: Note: I placed this here because it seemed like a good follower to the `system` and `parsing` documents, but I'm totally fine with it being moved to the bottom or somewhere else.

cookbook/foreign_shell_scripts.md
```diff
@@ -0,0 +1,241 @@
...
+after running the script. This is done by parsing output of the `env` command
+available on unix-like systems. The shell to execute can be specified and
+configured using the `--shell` and `--arguments` parameters, the command has
+been tested using sh (bash), bash, zsh, fish, ksh, and dash.
+
+As an example, we can use this set up Homebrew paths using `brew shellenv`. Note
+that the path to the `brew` binary depends on the OS and installation.
+
+```
+load-env (^/opt/homebrew/bin/brew shellenv | capture-foreign-env)
```
kubouch, line 118: This is perhaps the most important part of the page: A quick example of how to use the capture command. Ideally, it should be something more widely understandable. People not using Homebrew don't know what `brew shellenv` is. I think even `load-env ('script.sh' | capture-foreign-env)` would do. It is not immediately clear the `$in` should be a script filename, some people might try to do `open script.sh | capture-foreign-env` or similar to pipe in the script contents.

kubouch: (By "people" I mean "me" as these things came to mind when reading it.)

fdncred:
oooh, nice work!

cookbook/foreign_shell_scripts.md
remmycat: Yeah, that makes a lot of sense! I'll create a simpler, highlighted example tomorrow and fix a few leftover typos.

remmycat: @kubouch Actually, after rereading your comment: I didn't plan for you to be able to specify a script name, and I don't think it works, unless I missed something about `eval`.

So, `open` or `open --raw` would be the way to go, when having an existing script file. I liked having only one input method and thought it's important to allow the piping-in of the contents because tools generating env-var-scripts often pipe the script to stdout. Like brew in this case.

But I'm up for suggestions!
For now I'd add the example with `open`

kubouch: OK, then I misunderstood. It's not clear whether `$in` should be a script name or a source string, I didn't read the script source that thoroughly. Then the example could be `load-env (open script.nu | capture foreign-env)` or similar.

Side note: You can check [bass](https://github.com/edc/bass) if you didn't know it. A similar idea but for running bash scripts in fish, written in Python.

remmycat: Oh I see, no worries!
I updated the PR and tried to be a little more clear about this.

kubouch:
Perfect, thanks!

PR #1294, opened 2024-03-10
Remove feature `extra` detection from make_docs.nu

sholderbach:
If we sunset the `--features extra` for building Nushell we don't need
to warn in upcoming releases anymore.

Depends on https://github.com/nushell/nushell/pull/12140

cc @hustcer: Does this simplify the command generation enough so that we
don't need a parallel clone of the main repo anymore (you run it
locally, right?)


hustcer:
Yeah, it's easier than it was. Thanks. 👍🏻

PR #1295, opened 2024-03-10
Add "Contrib" link to the navbar

devyn:
This links the contributor book on the navbar, which seems to have been previously linked but removed, probably because it was long and took up a lot of space. As an alternative, this just calls it "Contrib" instead and puts it between "Commands" and "Cookbook" (the alphabetical order was appealing to me)


fdncred:
Thanks. It was originally removed because it was so out of date.

PR #1296, opened 2024-03-11
Add info on XDG_CONFIG_HOME

ysthakur:
Adds information on using `XDG_CONFIG_HOME` to control where Nushell looks for config files. Also added a warning about setting `XDG_CONFIG_HOME` to the directory above the Nushell-specific config directory, rather than the `nushell` directory itself.


fdncred:
We might want to have another warning that says, "Do not set your XDG_CONFIG_HOME in env.nu". I know you emphasized **before** but I think a warning and a more verbose statement of "do not do this" should be there too.

fdncred:
Thanks!

PR #1297, opened 2024-03-12
Update plugins.md (#12146)

jweckman:
More comprehensive plugin installation instructions that should be beneficial especially for newcomers.


jweckman:
I just noticed i used a copy/pasted #1297 at first. Added correct #12146.

Feel free to make any edits you like. I wonder if the warning was a bit excessive but in terms of reducing maintenance load and communicating clearly it is probably a good idea to indicate that a user should not expect plugins to work flawlessly at this stage in development.

Despite having a few years of dev experience this is actually my first ever open source commit so please go easy on me :)

fdncred:
Thanks

PR #1298, opened 2024-03-14
Document dynamic command arguments

zerfix:
I had been wondering how to add arguments to commands directly without a new instance of the shell, as its not documented anywhere as far as i can tell.

Now that i have figured it out a way to do it, i thought id add it to the doc.

I'm not sure if it should be mentioned that `"-a arg2"` does not work, as they become one argument with space instead of the normal two arguments split by space as one would normally expect. That also means that one does not need to double wrap like `'"long arg"'`.

I guess it could also be mentioned that you can also have static arguments instead of, or in addition, like `^$"command" --static-arg ...$arguments`.

### DIFF
--- a/lang-guide/lang-guide.md
+++ b/lang-guide/lang-guide.md
@@ -377,7 +377,7 @@ There are a couple things to be aware of in the above example.
 
 ### Executing String Interpolated strings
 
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #1302, opened 2024-03-14
Env var plugin engine calls docs

devyn:
Documentation for these PRs:

- nushell/nushell#12166
- nushell/nushell#12204


fdncred:
I realize this isn't released yet but I'd rather get it in and reference it.

PR #1306, opened 2024-03-16
Release notes for 0.92.0 for @devyn so far

devyn:
Mostly plugin stuff, except `ignore`.

I might have more to add before release, but figured it might be good to get these in now.


fdncred:
Good adds. Thanks!

PR #1307, opened 2024-03-16
fix(files.md): override Cargo.toml in example

qdm12:
- override existing `Cargo.toml` file with `-f` flag for `save` in example
- Point out `inc` is available through the `nu_plugin_inc`


fdncred:
Thanks

PR #1315, opened 2024-03-26
Update redirection examples in "coming from bash" for 0.92.0

IanManske:


sholderbach (approved):
Thanks!

sholderbach:
Any blockers keeping this draft?

IanManske:
> Any blockers keeping this draft?

I was waiting for 0.92.0 release, since I thought the examples may not work on previous versions. But it looks like this isn't the case.

PR #1320, opened 2024-03-31
Update workflow actions

IanManske:
Bumps the versions of actions used in workflows. This is because the old actions versions use `node` 16 which is at end of life, and support will [soon be deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) on Github. The relevant warnings can be seen, for example, [here](https://github.com/nushell/nushell.github.io/actions/runs/8493326000). As such, the `node` version was also bumped to 20.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


hustcer:
Thanks 👍

PR #1323, opened 2024-04-02
More little changes to the 0.92 release notes to cover things I did

devyn:
- Be a little more clear about what's changed with plugins from the beginning
- `into string` changes
- advice for upgrading (found some issues that can happen with `register` commands saved in `plugin.nu`)
- custom values examples


fdncred:
thanks

PR #1340, opened 2024-04-06
Fix make_docs.nu, adding generate docs for plugin commands support

hustcer:
We need to register the plugins first and then generate the docs for related plugin commands.

BTW:
It's weird that if you run `nu make_docs.nu` there will be an error output:
```console
Error: nu::shell::plugin_failed_to_load

  × Plugin failed to load: Plugin file not found
```
However, all the command docs have been generated successfully.


fdncred:
Thanks for all the great work you do on our docs!

PR #1347, opened 2024-04-10
ssh agent workaround for keychain

intellild:


cookbook/ssh_agent.md
```diff
@@ -52,6 +52,12 @@ do --env {
 }
 ```
 
+### [Keychain](https://www.funtoo.org/Funtoo:Keychain)
+
+```nushell
+load-env (keychain --eval --quiet <your ssh keys, eg. id_ed25519> | lines | split column ";" | get column1 | first 2 | split column "=" | rename name value | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value })
```
fdncred, line 58: I think it would look better if this pipeline was formatted as the other commands on this page in multiple lines. Other than that, it looks good.

intellild: done

fdncred:
Thanks

PR #1350, opened 2024-04-11
Update http_post.md fix example

ulvido:
http post never works without a body provided


fdncred:
updating anything in the commands folder is not permitted by the ci because they're systematically generated from the commands themselves. If this text is wrong, please file a pr to change the command help itself. thank you!

PR #1352, opened 2024-04-14
Update docs for 0.92.0 (file, pipes, and io)

IanManske:
Just realized I never made a PR for a branch I already prepared for 0.92. This edits the documentation / book to be up to date with the IO changes introduced in 0.92.


IanManske:
Translations will need to be updated, not sure what the process for that is.

PR #1353, opened 2024-04-15
Document all variants of `Value` and fix error types

devyn:
This is one of the really important Rust types that wasn't documented in full in the plugin protocol reference, and multiple non-Rust plugin developers said they couldn't quite grasp what was required from the Rust docs, so this provides examples for all of the variants we currently have.

Unfortunately this will need to be kept in sync manually.


fdncred:
looks good, thanks.

PR #1354, opened 2024-04-17
Link docs address in source.md

c4lliope:


fdncred:
Command md files cannot be modified here since they are auto generated with a nushell script after each release. We have to change the rust code for the command, and then it shows up here. Thanks for trying to make nushell better.

PR #1355, opened 2024-04-18
The `not` operator was missing in the list of the order of operations.

maxim-uvarov:
I tried to find its position experimentally (with the help of GPT-4), and it suggested the next test. So I placed the `not` before `and`.

```nu
> not true and false
false
> not (true and false)
true
```

I would be glad if anybody could check my conclusions.


fdncred:
Thanks

PR #1356, opened 2024-04-18
fix(lang-guide): IEE-754 -> IEEE-754

SCKelemen:
Spelling issues in lang-guide/lang-guide.md +252.
IEE-754 should be IEEE-754


sholderbach:
Thanks!

PR #1360, opened 2024-04-21
Edit 0.93.0 release notes

IanManske:


sholderbach:
Thanks for starting this. (Ian taking away my job :D)

In attempting to place my tweaks on it (the suggestion comment broke with the nushell code fences) accidentally merged when pushing a commit on top as the branch name was the same and I head already set my upstream to the main repo.

PR #1369, opened 2024-04-24
Update docs for commands from `nu-std/std/dirs.nu`

isti115:
Create an entry for `dexit`, which was missing, and add backlinks to the category page to the rest of the commands as well. (Closes https://github.com/nushell/nushell/issues/12642.)

### DIFF
--- a/commands/docs/dexit.md
+++ b/commands/docs/dexit.md
@@ -1,0 +1,18 @@
+---
+title: dexit
+categories: |
+  shells
+version: 0.92.0
+shells: |
+  Enters a new shell at the given path.
+usage: |
+  Enters a new shell at the given path.
+---
+
+# `dexit` for [shells](/commands/categories/shells.md)
+
+<div class='command-title'>Leaves the currently active working directory if there are multiple shell paths open.</div>
+
+## Signature
+
+```> dexit```
--- a/commands/docs/enter.md
+++ b/commands/docs/enter.md
@@ -2,16 +2,16 @@
 title: enter
 categories: |
   shells
-version: 0.79.0
+version: 0.92.0
 shells: |
   Enters a new shell at the given path.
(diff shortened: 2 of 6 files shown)


fdncred:
We don't allow editing of anything in the commands folder because those files are systematically generated with each new release. If we need changes to those commands the rust code or the std-lib code needs to change. Closing because of this. Thanks for trying to make our docs better. Please submit a PR on the nushell repo.

isti115:
Thanks for the feedback, @fdncred!
As noted by my comment on the issue at the `nushell` repo (https://github.com/nushell/nushell/issues/12642#issuecomment-2075440609) as far as I could tell, these specific files are not part of the automatic generation process. :thinking: 

They lack the `<!-- This file is automatically generated. [...] -->` notice, they have been left untouched for much longer (7 months) than everything else (3 weeks), their versions are stuck at `0.79.0` and (as opposed to the other commands, which are clearly documented by implementing the `Command` trait) I wasn't able to find their contents anywhere in the `nushell` repo, where they could potentially be generated from.

I think that even if the goal is to automatically generate the documentation of the standard library as well in the future, these changes are still valuable as they are right now, because the currently published version of the docs for these commands are clearly outdated. At least the entry for `dexit` should be added in my opinion.

fdncred:
@hustcer any ideas about these? Should we land these changes or are they systematically generated too?

hustcer:
Yes，I think we can land this PR. It's true that we can't generate docs for std libs

isti115:
Thanks for reopening and merging @fdncred and @hustcer!
In typical fashion, I noticed an error as soon as the PR got accepted, I've left the wrong frontmatter for `dexit.md`... :man_facepalming: 
Sorry for the mistake, I've sent a fixup in #1370!

PR #1372, opened 2024-04-28
Add reference to Nu Plugin Tracer in the Nushell Book.

tesujimath:


fdncred:
I've been on the fence about this PR, here's why. We allow other 3rd party software to be mentioned in our docs, but this is not just an external tool but something that is tied to the deep "guts" of nushell plugins. So, it makes me feel responsible that it always works, even though it's not nushell owned software. I'm not trying to make a case to "own" your software. Just sharing my feelings to have a conversation. I think we can probably land this. 🤔 

The last thing that bothers me about this tool is that there's no on/off switch. Correct me if I'm wrong, but once you "turn it on" with `plugin add --shell`, it will keep appending to the files in your home dir until "turn it off" with `plugin rm`. Is that correct? I was playing around with it and thought it was cool but then I was like, oh these files are getting larger and larger, how do I make it stop. 😆 

If it's not already in our awesome_nu repo, it could definitely go there. I don't have the same feelings about that repo.

tesujimath:
I did think about adding it to awesome_nu, but I decided against it.  Here's why.

The tracer is not a plugin per se.  It's a tool for plugin developers.  So it ought to be referenced from the place that plugin developers look to learn how to create plugins.

In contrast, the list of plugins on awesome_nu is a reference for end users.

Unless you envisage a new section, say Tools for Plugin Developers?  But that would be a pretty small section so far! :grin:

fdncred:
> Unless you envisage a new section, say Tools for Plugin Developers? 

I have no objections to another section.

Also, I ran into another problem. Since it's named in the same style as a plugin, it's easily mistaken for a plugin and fails to be registered with `plugin add` as in `toolkit add plugins`.

So, for me, having it named like a plugin isn't great. Also, as mentioned previously, not having some type of on/off switch makes it a little tricky to use too. There should at least be a warning that you'll fill up your file system if you don't disable it.

tesujimath:
Yes, I've also had reservations about the name. How about `trace_nu_plugin`?

I'll add a warning to the README about the ongoing accumulation of trace output. (I don't think it's worth adding an actual feature to mitigate that though, would make it unreasonably complicated.)

fdncred:
> trace_nu_plugin

Yes, I think that's even more descriptive.

> I'll add a warning to the README about the ongoing accumulation of trace output.

Sounds good. I'll land this PR with these changes.

tesujimath:
That's done. :sunglasses:

fdncred:
I thought you were going to put a warning here too?

tesujimath:
Ah sorry, I can also do that if you like.

tesujimath:
How's that?

book/plugins.md
```diff
@@ -157,6 +157,12 @@ Nu's main repo contains example plugins that are useful for learning how the plu
 
 The simplest way to debug a plugin is to print to stderr; plugins' standard error streams are redirected through Nu and displayed to the user.
 
+### Tracing
+
+The Nu plugin protocol message stream may be captured for diagnostic purposes using [trace_nu_plugin](https://crates.io/crates/trace_nu_plugin/).
+
+**WARNING: trace output will accumulate for as long as the plugin is installed with the trace wrapper.  Large files are possible.  Be sure to remove the plugin when finished tracing, and reinstall without the trace wrapper.**
```
fdncred, line 164:
```suggestion
**WARNING: trace output will accumulate for as long as the plugin is installed with the trace wrapper.  Large files are possible.  Be sure to remove the plugin with `plugin rm` when finished tracing, and reinstall without the trace wrapper.**
```

fdncred:
Looks good. Just one small change to help people know how to remove a plugin since it's new.

tesujimath:
Thanks!

PR #1376, opened 2024-04-30
Explain that $env is case-insensitive

YizhePKU:
[Related change](https://github.com/nushell/nushell/pull/12701).


fdncred:
Thanks

PR #1380, opened 2024-05-01
Make clear that `plugin use` is not needed in the config file

devyn:
At least one user on Discord found this surprising, as the `use` command for modules generally does need to be in the config file. Be more clear that it isn't.


fdncred:
Thanks

PR #1386, opened 2024-05-04
Fix user name in sys_get_nested_example.sh

### LINKED ISSUE #1385: Quick Tour example not related to the text
Hello,

While reading the Quick Tour chapter in the book I found this error :

Code example :
```sh
> sys | get host.sessions.name
╭───┬────╮
│ 0 │ st │
╰───┴────╯
```

Text :
```
Right now, there's just one user on the system named "sophiajt".
You'll notice that we can pass a column path (the host.sessions.name part) and not just the name of the column.
Nu will take the column path and go to the corresponding bit of data in the table.

You might have noticed something else that's different.
Rather than having a table of data, we have just a single element: the string "sophiajt".
Nu works with both tables of data as well as strings.
Strings are an important part of working with commands outside of Nu.
```

The code example should be :
```sh
> sys | get host.sessions.name
╭───┬──────────╮
│ 0 │ sophiajt │
╰───┴──────────╯
```

UmeSiyah:
If I well understood the line `sys | get host.sessions.name` should return the string `sophiajt` instead of `st`

I create an issue https://github.com/nushell/nushell.github.io/issues/1385

Close #1385


fdncred:
Thanks

PR #1388, opened 2024-05-04
Fix Keychain instructions in ssh_agent.md

justbispo:
Currently, the Keychain instructions in the cookbook don't work if you're trying to use GPG and SSH. The instructions assumes you're only using SSH.

With this change, it now removes the empty lines and uses the remaining ones, instead of only using the first 2.

Imo, ideally the page should also mention GPG. But my knowledge is superficial in this matter so I'll leave it to others.

### DIFF
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -58,9 +58,9 @@ do --env {
 load-env (
     keychain --eval --quiet <your ssh keys, eg. id_ed25519>
     | lines
+    | where not ($it | is-empty)
     | split column ";"
     | get column1
-    | first 2
     | split column "="
     | rename name value
     | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value }
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -55,16 +55,14 @@ do --env {
 ### [Keychain](https://www.funtoo.org/Funtoo:Keychain)
 
 ```nushell
-load-env (
-    keychain --eval --quiet <your ssh keys, eg. id_ed25519>
+keychain --eval --quiet <your ssh keys, eg. id_ed25519>
     | lines
     | where not ($it | is-empty)
-    | split column ";"
-    | get column1
-    | split column "="
-    | rename name value
-    | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value }
-    )
+    | parse "{k}={v}; export {k2};"
+    | select k v
+    | transpose --header-row
+    | into record
+    | load-env
 ```
 
 ## Non-nushell workarounds


fdncred:
@amtoine Seems like you made the most recent changes to this file, iirc. Could you look over this PR and provide feedback or land it please?

amtoine:
mm actually, i don't know nothing about `keychain`, i only wrote the SSH part above i think :confused: 
i'm using a slightly modified version of the _Workaround_ section :yum:

amtoine:
maybe if @justbispo can give a sample output of a `keychain` call, that'd be possible to see if the change is legit or not, otherwise i'm afraid i can't help much

justbispo:
Sure, the command I have on my config is `keychain --eval --quiet --noask -Q --agents gpg,ssh id_rsa id_ed25519` and the output is:

```
GPG_AGENT_INFO=/run/user/1000/gnupg/S.gpg-agent:1968:1; export GPG_AGENT_INFO;

SSH_AUTH_SOCK=/tmp/ssh-XXXXXXyd2dk5/agent.2138; export SSH_AUTH_SOCK;
SSH_AGENT_PID=2139; export SSH_AGENT_PID;


```

amtoine (commented):
thanks @justbispo :pray: 

with
```txt
GPG_AGENT_INFO=/run/user/1000/gnupg/S.gpg-agent:1968:1; export GPG_AGENT_INFO;

SSH_AUTH_SOCK=/tmp/ssh-XXXXXXyd2dk5/agent.2138; export SSH_AUTH_SOCK;
SSH_AGENT_PID=2139; export SSH_AGENT_PID;


```
and without considering the `load-env` call, this PR goes from
```
Error: nu::shell::column_not_found

  × Cannot find column
   ╭─[entry #4:9:54]
 8 │     | rename name value
 9 │     | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value }
   ·                                                      ─┬─ ──┬──
   ·                                                       │    ╰── cannot find column 'value'
   ·                                                       ╰── value originates here
   ╰────
```
to
```
──────────────┬───────────────────────────────────────
GPG_AGENT_INFO│/run/user/1000/gnupg/S.gpg-agent:1968:1
SSH_AUTH_SOCK │/tmp/ssh-XXXXXXyd2dk5/agent.2138
SSH_AGENT_PID │2139
──────────────┴───────────────────────────────────────
```

what about the following though?
```nushell
keychain ...
    | lines
    | where not ($it | is-empty)
    | parse "{k}={v}; export {k2};"
    | select k v
    | transpose --header-row
    | into record
```
looks much simpler ot me :)

justbispo:
That works too and it's much easier to understand than use the reduce.
Also, can the call
```nushell
load-env (
	...
	)
```
be replaced with
```nushell
	...
	| load-env
```
to not use a different syntax of the first example of the page?
It seems it can after testing it, but I'm still new to nushell so I'm just trying to be sure.

amtoine:
@justbispo 
yeah,
```nushell
{ FOO: "this is foo" } | load-env
```
and
```nushell
load-env { FOO: "this is foo" }
```
yield the same results :ok_hand:

justbispo:
@amtoine alright, thanks for your suggestion, just implemented it. Should be good to merge.

amtoine (approved):
looks pretty good to me, let's see how it goes :pray:

PR #1390, opened 2024-05-05
fix: wrong result in an example from 'lang-guide'

woosaaahh:
[Lang Ref Guide > Basic Types > Any > Example 1](https://www.nushell.sh/lang-guide/lang-guide.html#example-1)

The result in the example is `'true'` but the right value is `'false'`


IanManske:
Thanks!

PR #1391, opened 2024-05-07
Dataframe chapter update

maxim-uvarov:
`dfr` is going to be deprecated in favor of the `polars` plugin.
I've written installation instructions and updated all the examples with `polars` using `numd`.

I'm preparing a draft PR to wait before all the decisions on `dfr` deprecation are made. I guess I will submit it closer to 0.94.


fdncred:
Thanks

PR #1393, opened 2024-05-09
Corrections to `Operators precedence` chapter

maxim-uvarov:
There was a mistake with the order of the `not` operator. It has been fixed, and information about the command `help operators | sort-by precedence -r` has been added.

P.S. I'm not a native English speaker. Even though I have consulted with ChatGPT several times, any corrections from native English speakers are welcome!


fdncred:
Thanks

PR #1398, opened 2024-05-14
Initial `$in` language reference

NotTheDr01ds:
Here's a first pass at `$in` documentation for the Language Reference Guide.  Let me know if I missed any nuances or use-cases.

### DIFF
--- a/lang-guide/lang-guide.md
+++ b/lang-guide/lang-guide.md
@@ -1490,6 +1490,183 @@ It doesn't work because it sees `\P` and `\s` as escapes that are not recognized
 
 ### The pipeline special variable `$in`
 
+Much of Nu's composability comes from the special `$in` variable, which holds the current pipeline input.
+
+`$in` is particular useful when used in:
+
+- Command or external parameters
+- Filters
+- Custom command definitions or scripts that accept pipeline input
+
+#### `$in` as a command argument or as part of an expression
+
+Imagine you are creating a directory that includes tomorrow's date as part of the name. The following two commands are equivalent:
+
+```nushell
+mkdir $'((date now) + 1day | format date '%F') Report'
+```
+
+or
+
+```nushell
+date now                    # 1: today
+| $in + 1day                # 2: tomorrow
+| format date '%F'          # 3: Format as YYYY-MM-DD
+| $'($in) Report'           # 4: Format the directory name
+| mkdir $in                 # 5: Create the directory
+```
+
+While the second form may be overly verbose for this contribed example, you'll notice several advantages:
+
+- It can be composed step-by-step with a simple <kbd>↑</kbd> (up arrow) to repeat the previous command and add the next stage of the pipeline.
+- It's arguably more readable.
+- Each step can, if needed, be commented.
+- Each step in the pipeline can be [`inspect`ed for debugging](/commands/docs/inspect.html).
+
+Let's examine the contents of `$in` on each line of the above example:
+
+- On line 2, `$in` refers to the results of line 1's `date now` (a datetime value).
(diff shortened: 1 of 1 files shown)


kubouch:
I think this is a great addition, but I would split it a bit. The language guide is meant as a concise reference for the language. There should be only a short mention of what `$in` is and where it appears. The rest of your text is more of a tutorial style which is better fit for the book, possibly under https://www.nushell.sh/book/variables.html.

lang-guide/lang-guide.md
```diff
@@ -1490,6 +1490,183 @@ It doesn't work because it sees `\P` and `\s` as escapes that are not recognized
...
+
+With most filters, `$in` would refer to the entire `file` record (with `name`, `type`, `size`, and `modified` columns). However, with `update`, it refers specifically to the contents of the _column_ being updated, in this case `name`.
+
+#### Implicit `$in`
+
+The preceding example can be simplified even further because `$in |` is implicit at the beginning of a closure. So we can just use the more concise:
+
+```nushell
+ls | update name {str upcase}
+```
```
sholderbach, line 1563: I think this section is misleading because there is no hidden `$in` variable in our internal semantics. 

Instead part of the custom command semantics is that the first pipeline element of the first pipeline expression will receive the pipeline input to the custom command. (We should make sure that this is both documented in the spec and the general user documentation)

This has the distinct difference that it follows the pipelining semantics instead of the value/variable semantics. Thus it is lazy and carries the pipeline metadata (latter can be considered somewhat of an implementation detail for now)

We may implement optimizations for `$in` in the "first element of custom command"-position but from an educational perspective with the current semantics of Nushell it is much easier to teach `$in` is a special variable that you can use if needed but the pipelining behavior is expressed with `|` and the beginning of a custom command. (Teaching it that way also makes you aware of potential pitfalls, that you can not assume to receive no input in the first pipeline element of a custom command)

NotTheDr01ds: Fair enough - I'll work on fixing that.  Agree that the value/variable semantics are different enough (in streaming vs. collection) to make that a bad analogy.

NotTheDr01ds:
> I think this is a great addition, but I would split it a bit. The language guide is meant as a concise reference for the language. There should be only a short mention of what `$in` is and where it appears. The rest of your text is more of a tutorial style which is better fit for the book, possibly under https://www.nushell.sh/book/variables.html.

Thanks! And certainly open to seeing how we can split it out.  Regardless, I think The Book needs more discussion of `$in` than is currently there.

Keep in mind, though, that the README for the Language Guide says the goal is:

> to provide a little bit of documentation text for context and then many examples of usage

I know it "looks" like a tutorial at first glance, but it's really just 10 examples of usage of `$in`, along with the explanation of those examples.  In that light, doesn't it pretty much fit the criteria for the "many examples" of the Language Guide?

I also think it's a bit overkill for The Book, which it seems should be the intro/tutorial/digest version, right?

kubouch:
The goal of the reference is to have a concise overview of Nushell's syntax, kind of like https://doc.rust-lang.org/reference. The readme probably should be reworded to reflect that better. The goal of the book is to teach users how to use the language which can include examples. If we don't keep this distinction strict, we'll end up with two books.

I believe you can take most of your text, put it under a new section in https://www.nushell.sh/book/variables.html and in the language reference just add a short enumeration of where `$in` can appear. It can include examples, but just for showcasing the syntax, it doesn't need to showcase all the possible use cases for `$in` (like the `let day = $in` example, doesn't need to be in the language guide). Formulations like "Imagine that ..." are better suited for the tutorial-style book as well.

PR #1406, opened 2024-05-17
Change usages of `$nu.scope` to the `scope` command.

suimong:


fdncred:
Good catch. We changed from $nu.scope a long time ago. Just shows how docs get out of sync pretty easily.

Issue #1365, opened 2024-04-22
Broken links


IanManske:
The [latest deployment](https://github.com/nushell/nushell.github.io/actions/runs/8786044376/job/24107954260#step:5:16) had the following warnings regarding broken links.

```
warning Broken links found in cookbook/tables.md: ../book/docs/reduce.md
warning Broken links found in de/book/advanced.md: standard_library.md, dataframes.md, creating_errors.md, parallelism.md, explore.md
warning Broken links found in de/book/coloring_and_theming.md: commands/ls.md
warning Broken links found in de/book/coming_to_nu.md: coming_from_cmd.md, nushell_map.md, nushell_map_imperative.md, nushell_map_functional.md, nushell_operator_map.md
warning Broken links found in de/book/custom_completions.md: externs.md, externs.md, ../cookbook/external_completers.md
warning Broken links found in de/book/design_notes.md: how_nushell_code_gets_run.md
warning Broken links found in de/book/environment.md: modules.md, modules.md, modules.md, /commands/docs/env.md, /commands/docs/env.md
warning Broken links found in de/book/getting_started.md: quick_tour.md, thinking_in_nu.md
warning Broken links found in de/book/hooks.md: /Befehle/docs/if.md
warning Broken links found in de/book/loading_data.md: /commands/docs/size.md, /commands/docs/fetch.md
warning Broken links found in de/book/nu_as_a_shell.md: stdout_stderr_exit_codes.md, shells_in_shells.md, line_editor.md, externs.md
warning Broken links found in de/book/nu_fundamentals.md: pipelines.md
warning Broken links found in de/book/overlays.md: modules.md, modules.md#environment-variables, /commands/docs/overlay_remove.md, /commands/docs/overlay_remove.md
warning Broken links found in de/book/programming_in_nu.md: modules.md, testing.md
warning Broken links found in de/book/working_with_strings.md: /commands/docs/str_rpad.md, /commands/docs/build-string.md
warning Broken links found in ja/book/README.md: nu_fundamentals.md, programming_in_nu.md, nu_as_a_shell.md, coming_to_nu.md, design_notes.md, advanced.md
warning Broken links found in ja/book/getting_started.md: thinking_in_nu.md
warning Broken links found in ja/book/installation.md: book/dataframes.md
[...]

PR #1408, opened 2024-05-21
fix: use correct link to point to the command reduce

jaudiger:
Part of https://github.com/nushell/nushell.github.io/issues/1365


fdncred:
Thanks

Issue #1407, opened 2024-05-21
[Overlays] Basic example is not usable with latest Nushell


jaudiger:
From the example of chapter Basics: https://www.nushell.sh/book/overlays.html#basics

```nu
> module spam {
    export def foo [] {
        "foo"
    }

    export alias bar = "bar"

    export-env {
        load-env { BAZ: "baz" }
    }
}
```

I got the following error using Nushell 0.93.0:

```nu
Error: nu::parser::cant_alias_expression

  × Can't create alias to expression.
   ╭─[entry #1:6:24]
 5 │
 6 │     export alias bar = "bar"
   ·                        ──┬──
   ·                          ╰── aliasing String is not supported
 7 │
   ╰────
  help: Only command calls can be aliased.
```

jaudiger:
Works like a charm with the modification:

```nu
module spam {
    export def foo [] {
        "foo"
    }

    export alias bar = echo "bar"

    export-env {
        load-env { BAZ: "baz" }
    }
}

use spam

spam bar
> bar
```

jaudiger:
I could open a PR if you want

fdncred:
sure, go for it. thx!

PR #1409, opened 2024-05-21
fix: update export alias in overlays.md

jaudiger:
Resolve https://github.com/nushell/nushell.github.io/issues/1407


fdncred:
Thanks

PR #1413, opened 2024-05-27
Delete wrong vi motion command `d`

veeven:
There is no Line End motion implemented for `d` in https://github.com/nushell/reedline/blob/main/src/edit_mode/vi/motion.rs

I looked to see if three are other characters defined for Line End. `$` is already listed. No other keybindings defined for `d`. Hence deleting from the list in the book.


fdncred:
Thanks

PR #1422, opened 2024-05-29
update testing doc

amtoine:
`std testing` has been _deprecated_ -hidden to the users to be more precise- for quite some time now...

this PR updates the _testing_ documentation page to not nudge people to have a look at that hiddent module, especially because downloading `testing.nu` manually and putting it anywhere is not a very good workflow :confused: 

i removed all mentions to `testing.nu` and wrote a new section to run tests, which is split into two parts
- running tests from a Nupm package
- running tests for standalon modules or scripts


fdncred:
Thanks

PR #1424, opened 2024-05-29
Fix configuration cell-path link reference

Kissaki:
There is also `/lang-guide/chapters/types/basic_types/cellpath.html#cellpath`, but given that we are within `/book/` linking to that chapter seems more appropriate than linking to the language reference.

Partial #1365


IanManske:
Thanks!

PR #1425, opened 2024-05-29
Update cheat_sheet.md

### LINKED ISSUE #1262: Explanations in the cheat sheet should come before the code that's explained
I could open a PR making this change.

I think it'd be much more readable if the explanations weren't  `> quotes` and if they came before the code they explain. I've gotten confused many times when reading it, thinking one explanation belongs to another piece of code.

Also I think many other parts of the book also show the explanation before the example and I think this change would go in line with that.

Kissaki:
Fixes #1262

This change fixes a few issues the cheat sheet had, and makes some additional improvements:

* Drop non-printable characters
* Use description text instead of formatting them as quotes
* Description before code (block/case titles/headlines) (like suggested in #1262, which I agree with)
* Use imperative mood (for the most part)
* Some wording improvements
* Some conciseness improvements (a cheat sheet should be concise, specific, and generalized at the same time)
* Consistently use `>` for commands and add missing returns (some code lines were missing, mixed, or wrong)

### DIFF
--- a/book/cheat_sheet.md
+++ b/book/cheat_sheet.md
@@ -2,548 +2,469 @@
 
 ## Data types
 
-```nu
-    > "12" | into int
-```
+convert string to integer: `"12" | into int`
 
-> **converts string to integer**
+convert present date to provided time zone: `date now | date to-timezone "Europe/London"`
 
----
+update a record's language and if none is specified insert provided value: `{'name': 'nu', 'stars': 5, 'language': 'Python'} | upsert language 'Rust'`
 
-```nu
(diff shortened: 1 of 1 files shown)


fdncred:
ok, thanks.

fdncred:
Now that I look at the rendering, the top part doesn't look as good without the nushell syntax highlighting.
image

Kissaki:
I guess I'll change those back to code blocks then

Kissaki:
I agree it looks better with syntax highlighting

PR #1432, opened 2024-06-02
fix: Fixed broken link pointing to localhost

ZennoZenith:
Title
Fix: Correct broken link pointing to localhost

Description
This pull request addresses an issue with a broken link that was incorrectly pointing to localhost. The link has been updated to point to the correct relative URL within the documentation.

### DIFF
--- a/book/pipelines.md
+++ b/book/pipelines.md
@@ -118,7 +118,7 @@ With most filters, the second version would refer to the entire `file` record (w
 
 ### Pipeline input in custom command definitions and scripts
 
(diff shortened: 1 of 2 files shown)


book/pipelines.md
```diff
@@ -118,7 +118,7 @@ With most filters, the second version would refer to the entire `file` record (w
 
 ### Pipeline input in custom command definitions and scripts
 
-See: [Custom Commands -> Pipeline Input](http://localhost:8080/book/custom_commands.html#pipeline-input)
+See: [Custom Commands -> Pipeline Input](custom_commands.html#pipeline-input)
```
hustcer, line 121: It's better use `/book/custom_commands.html#pipeline-input` instead

PR #1434, opened 2024-06-02
Add reference to bash-env to cookbook

tesujimath:
Using bash-env is a lot simpler than using the convoluted work-arounds currently described there.

### DIFF
--- a/cookbook/foreign_shell_scripts.md
+++ b/cookbook/foreign_shell_scripts.md
@@ -63,6 +63,13 @@ This quickly gets tricky though, for example when the script is declaring a
 There are ways to implement some form of expansion too, but at some point it
 might make more sense to leave the parsing to the shell it was meant for.
 
+## Bash Env Plugin
+
+There is a third-party Nu plugin [bash-env](https://github.com/tesujimath/nu_plugin_bash_env)
+for importing environment variables from Bash format files and pipes.
+This plugin uses Bash itself to parse the environment definitions,
+and can therefore cope with arbitrarily complex Bash sources.
+
 ## Capturing the environment from a foreign shell script
 
 A more complex approach is to run the script in the shell it is written for and
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -21,6 +21,12 @@ Adding this to your `env.nu` will however start a new ssh-agent process every ti
 See the workarounds.
 :::
 
+Alternatively, use the third-party Nu plugin [bash-env](https://github.com/tesujimath/nu_plugin_bash_env) as follows.
+
+```nushell
+^ssh-agent | bash-env | load-env
+```
+
 ## Workarounds
 
 You can work around this behavior by checking if a ssh-agent is already running on your user, and start one if none is:


fdncred:
I'm not sure how I feel pushing other tools here in the cookbook. The real intent of the cookbook is to teach people how to use nushell, not how to work-around nushell.

tesujimath:
The problem here is that the cookbook makes it look quite awkward to use Bash environment, but that's not the case with the plugin.

I noticed also on a discussion that people have been wasting time solving this very same problem.

I encourage you to embrace the wider community contributions, so that all will benefit. 

(I was careful to describe the plugin as third-party BTW.)

Alternatively, you could simply remove the description of importing Bash environment and integrating with ssh-agent, as those descriptions are no longer best practice, and are in fact misleading IMO.

fdncred:
My comments aren't an indictment on your plugin but more of a comment about our documentation and the support we provide. If it's in our documentation people will ask us to support it, so we need to be careful about which external tools we allow in our documentation since we already have a hard time supporting everything nushell does. This may be able to be addressed by having some warning statement saying that questions regarding your plugin should be addressed in your repo.

Your plugin also uses jq which could be seen as a competitor to nushell in some respects. I wonder if it could be created without external dependencies like this.

One thing that I do love about your plugin is that it's written in a scripting language other than nushell. I'm not sure if you've seen it but devyn recently created a plugin written in nushell script. I'd personally love to see plugins written in several languages.

tesujimath:
Totally understand your concern about implying it is supported by the core Nushell team.  Have pushed up a very clear statement about that.

Regarding the use of `jq`, I don't think there's a reasonable alternative.  Trying to parse JSON with raw bash would be error prone and quite unworkable I think.  Using Nu itself would be really heavyweight for something for which `jq` is absolutely the right tool.  In Nu land, `jq` is obsolete.  But this plugin doesn't run in Nu land, it runs outside.

The very nature of the `bash-env` plugin is that it has to be written in Bash.  Everything else follows on from that.

cookbook/foreign_shell_scripts.md
```diff
@@ -63,6 +63,17 @@ This quickly gets tricky though, for example when the script is declaring a
...
+## Bash Env Plugin
+
+There is a third-party Nu plugin [bash-env](https://github.com/tesujimath/nu_plugin_bash_env)
+for importing environment variables from Bash format files and pipes.
+This plugin uses Bash itself to parse the environment definitions,
+and can therefore cope with arbitrarily complex Bash sources.
+
+**Please note that the `bash-env` plugin is not supported by the core Nushell team.
+All issues and requests for support should be directed to its own
+[issue tracker](https://github.com/tesujimath/nu_plugin_bash_env/issues).**
```
fdncred, line 75: Can you make these sections warnings?
```
::: warning blah blah blah :::
```

tesujimath: Done!

fdncred: Thanks. Sorry for making you jump through all these hoops.

tesujimath: It's all good. Nushell is totally worth the investment. You've managed to build something awesome. Thanks!

PR #1439, opened 2024-06-07
Link from 'TiN -> Immutable Variables' to Variables chapter to reduce redundancy

NotTheDr01ds:
I noticed that the "Thinking in Nu" chapter still had a section discussing the fact that *"Variables are immutable"* even though we've had `mut` for some time now.

There's still some useful information in here, as the *default* should be to use immutables, but it's no longer a *"Common surprise"* that we don't have mutables.

I've reworded and moved this content to the *"Variables"* chapter in the Book.  I've also consolidated some other info I wrote up in the Language Ref that better fits here.

### DIFF
--- a/book/thinking_in_nu.md
+++ b/book/thinking_in_nu.md
@@ -63,6 +63,14 @@ would work, since the string can be evaluated at compile-time:
 
 For more in-depth explanation, check [How Nushell Code Gets Run](how_nushell_code_gets_run.md).
 
+## Variables are immutable by default
+
+Another common surprise for folks coming from other languages is that Nushell variables are immutable by default. Coming to Nushell, you'll want to spend some time becoming familiar with working in a more functional style, as this tends to help write code that works best with immutable variables.
+
+**Thinking in Nushell:** If you're used to using mutable variables for different tasks, it will take some time to learn how to do each task in a more functional style. Nushell has a set of built-in capabilities to help with many of these patterns, and learning them will help you write code in a more Nushell-style. The added benefit of speeding up your scripts by running parts of your code in parallel is a nice bonus.
+
+See [Immutable Variables](variables.html#immutable-variables) and [Choosing between mutable and immutable variables](variables.html#choosing-between-mutable-and-immutable-variables) for more information.
+
 ## Nushell's environment is scoped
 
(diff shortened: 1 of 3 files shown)


fdncred:
It looks to me like all the info in thinking_in_nu.md is still valid but could be clarified by saying, "... from other languages is that Nushell `let variables` are immutable" and somewhere else we could say `mut variables` are mutable.

NotTheDr01ds:
Fair - I was reading it with my "historical" glasses on and the realization that, at the time it was written, we didn't have `mut` at all (or `for`/`while` and others that need mutable variables).  It just reads "oddly" to me a bit since it seems to imply that we still don't have mutable variables at all.

It also *seemed* to me that "Thinking in Nu"* was focused on things that might completely trip up users who were used to doing things a different way.  While it's still "best practice" to use immutability in Nushell, it's no longer *required*.

That said, perhaps it is such an *important* best practice that it still belongs in the "front matter" (Getting Started -> Thinking in Nu)?

I'll move it back and work on a few tweaks to make it "cleaner".

Any objection to me shortening up the "Thinking in Nu -> Variables are immutable by default" section by making it an "overview" with the *detail* still in the Variables chapter?

fdncred:
> at the time it was written, we didn't have mut at all

yup, 100% right. it hasn't aged terribly well.

One thing that still fits but may not be talked about much are the things like how it's related more closely to a compiled language like why we have env.nu and config.nu although some of that has changed too because we now have const and we have a growing set of commands that are const. Maybe this is what I'm thinking about too https://www.nushell.sh/book/how_nushell_code_gets_run.html.

> Any objection to me shortening up the "Thinking in Nu -> Variables are immutable by default" section by making it an "overview" with the detail still in the Variables chapter?

probably not

NotTheDr01ds:
> Maybe this is what I'm thinking about too

Yup - Given that there are links in both directions (from TiN <-> *"How Nushell code gets run"*), I think it's up-to-date.  And I'm still grokking it a year after first reading it ;-)

As for the Immutable Variables, I refactored it so that the "core concept" is still in *"Thinking in Nu"* but consolidated the rest of the info and examples into the *"Variables"* chapter.  Let me know if that revision looks okay to you.

fdncred:
let's move forward with this one and the cookbook one. thanks

PR #1440, opened 2024-06-08
remove feature from frontmatter and commands page

amtoine:
because Nushell doesn't have features anymore apart from the default pool of commands, i propose to remove the `$.feature` column from the "commands" page and remove it from the `make_docs.nu` script as well for it not to appear in the frontmatter of the commands anymore.

> **Note**  
> i wanted to run `nu make_docs.nu` and generate the new commands frontmatter sections but the script does not appear to work on 0.94.1 anymore :confused:


hustcer:
Thanks

PR #1444, opened 2024-06-11
Replace `exit` with `dexit` for popd analog

gar1t:
Please see https://github.com/nushell/nushell/issues/12642#issuecomment-2074738988


hustcer:
Thanks

PR #1449, opened 2024-06-17
Extensive Nu Map Cleanup

NotTheDr01ds:
Pretty extensive cleanup on aisle 6, err, [Nu map from other shells and domain specific languages](https://www.nushell.sh/book/nushell_map.html):

* Formatted all commands in Markdown
* Linked to online help for all Nushell commands
* Replaced/fixed deprecated/removed commands (`calc`, `nth`, `sum`)
* Disambiguated and fixed `range` (command) and `range` (type)
* Bash `printf` added as a string formatter
* Keybindings added for `exit` and `clear`
* PowerShell `NewItem -ItemType Directory` added for `mkdir`
* PowerShell `Get-Command` added for `which`
* `$env.FOO = "bar"` example added for Nushell
* Fixed incorrect Bash `export` statement
* `ls --du` added for `du`
* `$nu.default-config-dir` added for config
* And more


fdncred:
Nice cleanup. I'm not sure it's really possible to have this list complete. Like `uname` is now a thing where it wasn't when I first created this. There are a lot of `str` commands that aren't mentioned, which is fine. It's just kind of a hard thing to keep this evergreen.

`size` was renamed `str stats` - so that should probably be in there still.

NotTheDr01ds:
> I'm not sure it's really possible to have this list complete.

Agreed, *especially* as I start looking at the Python (and other languages) mappings.

> Like `uname` is now a thing where it wasn't when I first created this.

Ah, right - I did know that.  I'll fix that when I get to a stopping point on the Language mapping.

> `size` was renamed `str stats` - so that should probably be in there still.

Thanks - I'll update that one as well.

> There are a lot of `str` commands that aren't mentioned, which is fine.

On the bright side, I was able to link to the main `str` command, so at least they are represented there.

NotTheDr01ds:
Oh, and I came back here to add some "formatting notes" as well ...

* I tried to match Markdown column widths, but if one got *too* wide for a particular row, I just fixed/aligned it in the *next* column.
* Because of wrapping ugliness, I've added some hardcoded `<br />`'s and even (since `&nbsp;` can't be used inside backticks), some hardcoded nonbreaking-spaces (character code 0160).  Hopefully this doesn't cause issues ...

fdncred:
Ping me once you're ready on this one too.

NotTheDr01ds:
> Ping me once you're ready on this one too.

Should be good to go - Thanks!

fdncred:
Thanks

PR #1451, opened 2024-06-17
Imperative Language Map Cleanup

NotTheDr01ds:
Similar cleanup as in #1449:

* Linked to Nushell command help
* Formatted all keywords with backticks - Note that there may be a couple of minor errors in here, as I couldn't quite tell if some items were a "keyword" or a "concept" (or even a struct/type).  For instance, I did not format C++ "Regex" as a keyword, since it seems to me a concept.  On the other hand, I'm not sure about some uses of `set`.  Please review and tweak as needed.

* Fixed `size` as noted in #1449


fdncred:
I think it's fine to make incremental improvements. So, we can land this and if there are other tweaks, we can keep adding them.

PR #1460, opened 2024-06-27
Update installation instructions with `cargo`

Garbaz:
Hi,

This is only a small change.

I have removed the paragraph about including the `dataframes` feature, as it no longer exists (I presume it's just enabled by default?). At least the command `cargo install nu --locked --features=dataframe` given in the book fails with:

```
error: failed to compile `nu v0.95.0`, intermediate artifacts can be found at `/tmp/cargo-install0x8JXF`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  none of the selected packages contains these features: dataframe
```

I also added `--locked` to the remaining `cargo install` command.

### DIFF
--- a/book/installation.md
+++ b/book/installation.md
@@ -82,10 +82,6 @@ Nu releases are published as source to the popular Rust package registry [crates
 
 That's it! The `cargo` tool will do the work of downloading Nu and its source dependencies, building it, and installing it into the cargo bin path.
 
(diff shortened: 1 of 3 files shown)


fdncred:
We traditionally don't recommend using `--locked`. We use that as a fallback mechanism if an upstream crate is misbehaving. Please remove that and we can land this.

fdncred:
Thanks

PR #1461, opened 2024-06-28
Dataframes chapter update

maxim-uvarov:
I removed the obsolete warning and added a conventional `Output:` line before the results of block executions (so `numd` can update them out of the box now).

Also, I tested examples with `polars 0.95.1` and they worked with no problems, but I didn't update timings in the markdown as they were without significant changes.


fdncred:
Thanks!

PR #1462, opened 2024-06-29
Document std path add breaking change

t-mart:
As instructed in #1457, this is a suggestion for the release notes related to <https://github.com/nushell/nushell/pull/13258>, a breaking change for `std path add`.

To retain the univocality these blog posts seem to have, I offer full editorial control for things like structure, style, grammar, list ordering, etc.


fdncred:
thanks

PR #1463, opened 2024-07-01
Fix str contains case sensitive option

f3wenbo:
Change the case sensitive option from '--insensitive' to '--ignore-case'.


f3wenbo:
When I was learning to use nushell, I found that the case-sensitive option for the command `str contains` still keeps the old one in the document. So I fixed it and created this pull request.

sholderbach (approved):
Thanks, good catch!

PR #1464, opened 2024-07-04
Fix broken `$env.PATH` example

opeik:
I lost several hours to this typo tonight and I don't want anyone else to.


IanManske:
Thanks!

PR #1470, opened 2024-07-16
Replace `group` with `chunks` in docs

IanManske:
Following from https://github.com/nushell/nushell/pull/13377 which deprecates the `group` command in favor of a new `chunks` command.

### DIFF
--- a/cookbook/tables.md
+++ b/cookbook/tables.md
@@ -26,13 +26,13 @@ Output:
 ───┴───┴───┴───┴───
 ```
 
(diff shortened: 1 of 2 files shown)


NotTheDr01ds:
I can get these separately, but *may* also used in:

* `cookbook/tables.md`
* `commands/docs/query_web.md` (need to check for sure) (but regardless, in core, not doc)
* `cookbook/jq_v_nushell.md` (maybe, looks like it maybe commented out)
* `book/nushell_map_functional.md`

Will update after I review

devyn:
You can undo the typos change, I just corrected the example to not trigger it

PR #1471, opened 2024-07-20
add --quiet flag to `watch` command

Zoybean:
see nushell/nushell#13415


sholderbach:
The command help pages are generated from the internal help after the release, so you don't have to do anything here manually.
Thanks for being proactive about it.

PR #1477, opened 2024-07-23
Fixes related to units in types_of_data.md

TheTrueWhiteOwl:
Changed the file size units from lowercase to their actual metric and ISO 80000 units. This includes changing the lowercase `b` into `B` which avoids the confusion with the `bit` unit which is sometimes written as `b`.

Changed microsecond unit from `us` to `μs`. Kept a footnote stating that `us` can be used as an alternative if necessary.


book/types_of_data.md
```diff
@@ -97,16 +97,18 @@ Dates are in three forms, based on the RFC 3339 standard:
...
-| `1ms`    | one millisecond |
-| `1sec`   | one second      |
-| `1min`   | one minute      |
-| `1hr`    | one hour        |
-| `1day`   | one day         |
-| `1wk`    | one week        |
+| Duration  | Length          |
+| --------- | --------------- |
+| `1ns`     | one nanosecond  |
+| `1μs`[^1] | one microsecond |
```
sholderbach, line 103: While I understand the appeal to follow the general conventions, I think in practical use we should point users to `us` directly as that is what most people can use more easily with international keyboards.

sholderbach (commented):
While I think its okay to follow the best practices in our documentation, it is worth pointing out that we support `b`/`B` interchangeably, as there are a bunch of things out there that are not as tight on those conventions.

book/types_of_data.md
```diff
@@ -22,7 +22,7 @@ The [`describe`](/commands/docs/describe.md) command returns the type of a data
 | Booleans          | `true`                                                                |
 | Dates             | `2000-01-01`                                                          |
 | Durations         | `2min + 12sec`                                                        |
-| File sizes        | `64mb`                                                                |
+| File sizes        | `64MB`                                                                |
```
fdncred, line 25: Maybe this should be `64mb` or `64MB` to drive home the fact that nushell doesn't care about case in these instances.

book/types_of_data.md
```diff
@@ -128,28 +130,26 @@ Nushell also has a special type for file sizes. Examples include `100b`, `15kb`,
...
-- `tb`: terabytes
-- `pb`: petabytes
-- `eb`: exabytes
-- `kib`: kibibytes (aka 1024 bytes)
-- `mib`: mebibytes
-- `gib`: gibibytes
-- `tib`: tebibytes
-- `pib`: pebibytes
-- `eib`: exbibytes
+| Decimal (factor of 1000) | Binary (factor of 1024) |
```
fdncred, line 133: I like the table other than `kB` looks weird below but I'd also lead with unit case doesn't matter. kib = KiB = kIB = Kib, etc.

book/types_of_data.md
```diff
@@ -97,16 +97,18 @@ Dates are in three forms, based on the RFC 3339 standard:
...
+| `1ns`     | one nanosecond  |
+| `1μs`[^1] | one microsecond |
+| `1ms`     | one millisecond |
+| `1sec`    | one second      |
+| `1min`    | one minute      |
+| `1hr`     | one hour        |
+| `1day`    | one day         |
+| `1wk`     | one week        |
+
+[^1]: Can be simplified to `us` if unicode or the greek letter mu (μ) is unavailable
```
NotTheDr01ds, line 111:
```
[^1]: Can be simplified to `us` if Unicode or the Greek letter Mu (μ) is unavailable
```

NotTheDr01ds, line 111:
```suggestion
[^1]: Can be simplified to `us` if Unicode or the Greek letter Mu (μ) is unavailable
```

book/types_of_data.md
TheTrueWhiteOwl:
```suggestion
| File sizes        | `64MB`, `64Mb`, `64mb`                                                |
```

Too many? I put a couple that are equivalent

fdncred: I'm fine with it

book/types_of_data.md
TheTrueWhiteOwl:
```suggestion
| `1μs` or `1us`[^1] | one microsecond |
```
Are you suggesting something like this, where we maintain the footnote, or should I just remove the footnote outright?

fdncred: I'd just keep them both like you have in your suggestion above and remove the footnote.

NotTheDr01ds:
So if I'm understanding the review comments correctly, I think this would be ready to land if the two current review suggestions are accepted, right?

NotTheDr01ds:
If there are no objections, I'd suggest holding off on merging this.  I'm in the process of the larger "Types" rewrite, and it looks to me like most of the changes here would get incorporated into the Language Ref anyway.  I'm pulling them in manually.

fdncred:
I'm fine with that @NotTheDr01ds. Let's just make this draft for now to see how your Types PR comes out.

NotTheDr01ds:
@TheTrueWhiteOwl I pulled most of your (very helpful - Thank you) changes from this PR into my (fairly massive) update on the Types across both the Book and the Language Reference.  Can you review and potentially close this one?  If there are any changes that you feel are important that I missed, of course, let's still look at getting them in.

PR #1478, opened 2024-07-23
Updates and fixes for Cookbook Tables

NotTheDr01ds:
* Main purpose is to update `group` to `chunks`
* While I was there, I found that the entire page was essentially broken. Not a single example worked as intended. I sadly don't think anyone has tried those examples in the nearly 2 years they've been on the site.

With apologies to the author, even after updating, I think these are pretty bad examples.  I'd like another set of eyes on them, but I really think this page needs to be removed or reworked entirely.


cookbook/tables.md
```diff
@@ -52,18 +55,18 @@ Output:
...
 
-We could join all three tables like this:
+We can merge all three tables like this:
 
 ```nu
-$first
-| chunks ($second | length)
-| each { merge $second }
+$first_table
+| chunks ( $second_table | length )
```
IanManske, line 65: All the others don't have spaces around parens.
```suggestion
| chunks ($second_table | length)
```

cookbook/tables.md
```diff
@@ -79,7 +82,7 @@ Output:
 ───┴───┴───┴───┴───┴───┴───
 ```
 
-Or just like last time we could use the [`reduce`](../commands/docs/reduce.md) command to merge tables together recursively:
+Or as mentioned in [The Cookbook](https://www.nushell.sh/book/working_with_tables.html#merging-tables) we can use the [`reduce`](../commands/docs/reduce.md) command to merge tables together recursively:
```
IanManske, line 85: Don't need to capitalize "the".
```suggestion
Or as mentioned in the [Cookbook](https://www.nushell.sh/book/working_with_tables.html#merging-tables) we can use the [`reduce`](../commands/docs/reduce.md) command to merge tables together recursively:
```

IanManske (commented):
Thanks for the fixes! Just two small things:

NotTheDr01ds:
> Thanks for the fixes! Just two small things:

Thanks!  Good catch - I tried to resolve a lot of those spacing issues (they were mentioned in your original review 2 years ago as well), but still missed a few.

fdncred:
Thanks y'all

PR #1482, opened 2024-07-25
Update 00_flow_control_overview.md

sinisa-szabo:
There is missing "s" in the example command:
help commands | where category == filter
should be
help commands | where category == filters


sholderbach (approved):
Thanks! Well spotted.

PR #1483, opened 2024-07-27
Update `CustomValue` usage example

devyn:
The example given in `contributor_book/plugins` was out of date. `value_string()` was renamed to `type_name()` quite a while ago to reflect what it is actually used for, and `as_mut_any()` was added to support mutating the custom value without having to clone it first.


fdncred:
Thanks

PR #1484, opened 2024-07-27
Add example for the new CallDecl engine call

devyn:
@fdncred - hopefully this helps fill that gap!

### DIFF
--- a/contributor-book/plugins.md
+++ b/contributor-book/plugins.md
@@ -730,6 +730,47 @@ Note that opting out of garbage collection does not stop users from explicitly s
 
 If your plugin takes a particularly long time to launch, you can recommend to your users that they change their [garbage collection settings](/book/plugins.html#plugin-garbage-collector) to either increase the `stop_after` duration, or disable garbage collection entirely for your plugin.
 
+## Making calls to other Nushell commands
+
+Plugins can look up and make calls to other Nushell commands in the scope of the original plugin call. This includes internal commands, custom commands written in Nushell, as well as commands provided by plugins. The relevant calls are [`FindDecl`](plugin_protocol_reference.html#finddecl-engine-call) and [`CallDecl`](plugin_protocol_reference.html#calldecl-engine-call).
+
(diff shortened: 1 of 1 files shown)


fdncred:
This is great! Will definitely help. 

I wasn't able to express my real problem yesterday so let me try to say it here. In order to figure out how to use `EvaluatedCall` was to look at `view ir` and play around with it. 

My problem was that I got stuck trying to figure out if I should use `.with_flag()`, `.with_positional()`, `.with_named()` for `into datetime --timezone 'LOCAL'`. I'm guessing they map back to functions used in `Signature`? My problem is that I've never thought the functions for `Signature` was very intuitive.

devyn:
Right. We basically have two major types of arguments: positional, and named. For named, we have kind of two variants: named arguments that take values, and named arguments that don't / can be set to boolean only with `=`. We call that latter type a few different things and haven't seemed to settle on one name for it: sometimes it's called flag, other times switch, and I think maybe I've seen some other names too.

A named (`--option`) argument that always takes a value is a `named` argument though, for sure. It's not wrong to use that for the boolean switch type arguments as well and just set it to true, but to match exactly what `--switch` does, it's just present with no value. Otherwise it's `--switch=true` basically, which works, it's just not the same.

fdncred:
Eventually it would be nice to have the function names cleaned up or at least some documentation that describes the nuances. 
Maybe I'm missing the point but we have this:

- `Signature` - builder functions `required`, `optional`, and `rest`, then later on we have regular functions `named`, `required_named`, and `switch`
- `CallExt` - `has_flag`/`get_flag`, `rest`, `opt`, `opt_cost`, `req`, `req_parser_info`. 
- Then we have the plugin equivalents. 
- Then there's the other builder in `EvaluatedCall` that I mentioned above.
- I'm sure I'm leaving someone out here 😆 

I'm just saying it would be nice to be verbose in the naming for what these are for and to be consistent. Not blaming anyone for anything, it's just not super intuitive to me.

Thanks for your input and this PR to clean things up a bit. ❤️

devyn:
Yeah, I think it's just objectively not intuitive so I'll spend some time on trying to define it

PR #1490, opened 2024-08-02
More improvements to 'Moving around' chapter

NotTheDr01ds:
Continuing the general clean-up of this chapter that I started when adding the "multi-dot" info.

* General grammar and wording tweaks - In general, tried to remove prepositions, especially "we" (but not all), and focus more on what *Nushell* does. This involves the use of more passive voice, but I hope it reads more professionally in most cases.
* Linked to other sections to reduce redundancy
* Removed the wording about Nushell passing globs "as is" to commands, since this only applies to internals, not externals.  Also it was quite out-of-date, since all applicable filesystem commands now accept `...rest`
* Added raw strings as a valid glob-escaping method
* Clarified the explanation/example of why backticks don't escape globs
* Likewise, the "string to glob" conversion was outdated due to the `...rest` changes
* Removed the examples from the `cp`, `mv`, etc. commands since they are already linked to the command help itself, and these already include better and more complete examples.
* Other improvements

### DIFF
--- a/book/moving_around.md
+++ b/book/moving_around.md
@@ -1,41 +1,104 @@
-# Moving around your system
+# Moving around the system
 
-Early shells allow you to move around your filesystem and run commands, and modern shells like Nu allow you to do the same. Let's take a look at some of the common commands you might use when interacting with your system.
+A defining characteristic of a shell is the ability to navigate and interact with the filesystem. Nushell is, of course, no exception. Here are some common commands you might use when interacting with the filesystem:
 
 ## Viewing directory contents
 
 @[code](@snippets/moving_around/ls_example.sh)
 
-As we've seen in other chapters, [`ls`](/commands/docs/ls.md) is a command for viewing the contents of a path. Nu will return the contents as a table that we can use.
+As seen in other chapters, the [`ls`](/commands/docs/ls.md) command returns the contents of a directory. Nushell's `ls` will return the contents as a [table](types_of_data.html#tables).
 
-The [`ls`](/commands/docs/ls.md) command also takes an optional argument, to change what you'd like to view. For example, we can list the files that end in ".md"
+The [`ls`](/commands/docs/ls.md) command also takes an optional argument to change what you'd like to view. For example, we can list the files that end in ".md"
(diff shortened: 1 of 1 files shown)


book/moving_around.md
```diff
@@ -1,41 +1,104 @@
...
+However, _backtick_ quoted strings do not escape globs. For example:
+
+```nu
+# Removes all files in the current directory that contain
+# "myfile" as part of the filename
+> rm *myfile*
+
+# Quoted asterisk - Removes only a single file with the
+# name *myfile* (including the asterisks)
+> rm "*"
```
fdncred, line 58: I thought double quotes meant this is a literal and don't expand it. At least that's what I thought it was supposed to be. 

Example:
```nushell
❯ touch "$"
❯ ls
╭─#─┬─name─┬─type─┬─size─┬─modified─╮
│ 0 │ $    │ file │  0 B │ now      │
╰───┴──────┴──────┴──────┴──────────╯
❯ rm $
Error: nu::parser::variable_not_found

  × Variable not found.
   ╭─[entry #17:1:4]
 1 │ rm $
   ·    ┬
   ·    ╰── variable not found. Did you mean '$r'?
   ╰────

❯ rm "$"
```

NotTheDr01ds:
@fdncred You're right - That's what I say in the comment, but then the code example isn't what I thought I put there.  Let me see what went wrong, but I definitely will push a fix for that in a moment.

NotTheDr01ds:
Thanks for catching that!

It looks like I was mid-edit on changing the `"*"` example to `"*myfile"`* and forgot to finish it.

I've fixed the issue, but I've also expanded the text a bit.  Hope that looks better.

book/moving_around.md
NotTheDr01ds: Thanks for catching that!

It looks like I was mid-edit on changing the `"*"` example to `"*myfile"`* and forgot to finish it. :-/

I've fixed the issue, but I've also expanded the text a bit.  Hope that looks better.

fdncred:
Thanks

PR #1492, opened 2024-08-02
Updates to Book - Types chapter

### LINKED ISSUE #1073: Documentation for `null` should mention `default`
I was looking for null-related commands in the book and (without knowing in advance what to search for) did not find anything about `default` until after searching for "null" on the nushell github and finding a [stray comment](https://github.com/nushell/nushell/issues/7655#issuecomment-1368959540) mentioning the `default` command as the equivalent for rust's `unwrap_or`.
It would be great to mention the available commands that manipulate possibly-null values in the [section about null](https://www.nushell.sh/book/types_of_data.html#null). I'm still not sure if I'm missing other relevant commands.

NotTheDr01ds:
Started with the goal of implementing the request in #1073 and, well ... it grew from there. ;-)

Multiple changes:

* First and foremost, adds information on `default` to the section on Nulls/`nothing` - This should close #1073 

* The type chapter never actually listed the *type* names themselves.  I started to add these to the heading, but that makes anchor-links really unwieldy.  Instead, I've put in a "callout" in the "warning" formatting (just yellow) with the type name for each type. This could be changed to a "tip" (cyan) callout if the current is too jarring, but I think the yellow is nice in that it is distinct from the other "tips"/"notes".

* Fixed many nitpicks, like making names in examples proper case (e.g., "Sam" instead of "sam")
* Removed the String examples and instead linked to the String chapter which has much better and more complete examples.
* Added information on delimiter options for record and list literals
* Many, many grammatical fixes.  Probably created a few grammatical issues as well ;-).
* Made many of the examples more clear and complete
* Removed note regarding Nushell release 0.71
* Many others


fdncred:
I'm not saying what you've done is wrong, I just have a question. I wonder if some of this type information should be in the reference guide.

NotTheDr01ds:
Right - I'm not sure either, especially after [this discussion](https://github.com/nushell/nushell.github.io/pull/1398#issuecomment-2113701690).  My first thought when I was looking at the Language Reference was that *most* of the Type information should be moved to the Reference, with only a "brief overview" (couple of paragraphs each) in the Guide, but after that feedback from @kubouch I wasn't as sure.  This is specifically what I was thinking of [when I wrote](https://github.com/nushell/nushell.github.io/pull/1398#issuecomment-2133633073):

> I still have some questions on what belongs in each doc, though, but I'll take those up on a separate issue. I plan to do a lot more work around this, and I'd like to make sure it's clear where things go.

But then I never followed up on that.  But, since you bring it up ... ;-)

@kubouch With some time passed since we originally discussed this in context of `$in`, what are your thoughts on putting Types *detail* in the Language Ref?  The "Getting Started" version would remain in the Book, along with links between the two, of course.

---

Keep in mind that we could also make this *look* a bit more concise in the Book by putting some of it behind a:

<details><summary>Details</summary>
<p>
<br />VuePress supports this as a built-in
</p>
</details>

kubouch:
Looking at our Ref, it seems like much (not all) of this PR, and the Types of Data chapter in the book could be in the Ref. There seem to be information duplication between the two.

Instead of details of each type, the book could focus more on integration, e.g., how type annotations works, type inference, gradual type system (type errors caught at parse- vs eval-time), etc.

We still need to flesh out the scope of the Book and the Ref, but to me, the Book should have just about enough info to get you started with the most important stuff and help you understand Nushell's language design, and if you need more detail on the syntax, you'd go to the Ref. If you need more practical examples, you'd go to the Cookbook.

NotTheDr01ds:
Thanks - I think that's a sensible approach, and I can add that guidance to the Contributing guide as well.

I'll take a stab at sorting out the existing information between the (mostly two) docs and reducing/removing duplication.

NotTheDr01ds:
Still WIP/draft, but I pushed an update at lunch today that takes a first pass at refactoring the Book and Language Guide types info. Would appreciate any feedback at this stage.

I've made it through the Basic Types so far.

* In the Book, standardized the format for each type, with a list of:
  - Description
  - Annotation
  - Literal Syntax (when it's basic enough - For some I just punted it to the Language Guide)
  - "See also" link to the Language Guide
  - Simple examples
  - Basic tips

The result is that each Type in the Book is pretty basic - As @kubouch said:

> just about enough info to get you started with the most important stuff and help you understand Nushell's language design

I haven't progressed as far in the Language Guide, but I've most of the basics (other than String, I think).

The *Language Guide* duplicates only the most basic info (Description, Annotation) but also goes into more complex literal syntax - Such as the info in #1477 which was originally in the Book.

It also includes:

* Internal information when appropriate
* Cast command
* More complete examples when appropriate (e.g., `range`).
* Common commands that can utilize the type (from @fdncred's material originally)
* Common operators that can operate on the type (ditto)

lang-guide/chapters/types/basic_types/date.md
```diff
@@ -1,28 +1,18 @@
 # Date
 
-What it is: A value representing a specific point in time using international standard date time descriptors.
+- **_Description:_** Represents a specific point in time using international standard date time descriptors
+- **_Annotation:_** `date`
+- **_Literal syntax:_** RFC 3339
+  - Date-only: `2022-02-02`
+  - Date and time (GMT): `2022-02-02T14:30:00`
+  - Date and time including timezone: `2022-02-02T14:30:00+05:00`
```
fdncred, line 8: correct me if I'm wrong, but the +05:00 is called a timezone offset I believe
```suggestion
  - Date and time including timezone offset: `2022-02-02T14:30:00+05:00`
```

lang-guide/chapters/types/basic_types/duration.md
```diff
@@ -1,72 +1,47 @@
...
-| `1us`    | one microsecond |
-| `1ms`    | one millisecond |
-| `1sec`   | one second      |
-| `1min`   | one minute      |
-| `1hr`    | one hour        |
-| `1day`   | one day         |
-| `1wk`    | one week        |
+| Duration  | Length          |
+| --------- | --------------- |
+| `ns`      | one nanosecond  |
```
fdncred, line 14: The headers need to be changed if you plan on removing all the 1's. They appear to no longer be Duration and Length but Abbreviation and Definition.

fdncred:
So far, so good. I just want to be sure not to lose any data. If you're deleting stuff from one place, I want to make sure we're not losing it altogether, but it's put back in the more detailed parts of the book.

I made a couple minor comments.

NotTheDr01ds:
Thanks - Both good catches, of course. Committed the first one via Github suggestion and I have the other already changed locally. I'll mark it resolved when I push next.

NotTheDr01ds:
Looking through the diff - The most "wholesale" deletion was from `range` in the Book.  But that's information that I had added in the first round of this PR myself anyway, and it was better represented by (i.e., a duplicate of) your examples in the Language Guide.

I also nixed the *code* example from #1477 that showed that capitalization doesn't matter - It was already represented in text form (also from #1477) and I highlighted it even more by putting it in a "tip" callout.

Everything else I see was moved from one location to the other.

lang-guide/chapters/types/basic_types/duration.md
NotTheDr01ds: Resolved in latest commit

lang-guide/chapters/types/basic_types/date.md
```diff
@@ -1,28 +1,18 @@
 # Date
 
-What it is: A value representing a specific point in time using international standard date time descriptors.
+| **_Description:_** | Represents a specific point in time using international standard date time descriptors
```
fdncred, line 3: I think this table is missing the top part

fdncred:
@NotTheDr01ds nice work! you continue to do an excellent job with these docs. I really like the standardization of the table header for types. That type of consistency really helps users understand things.

lang-guide/chapters/types/basic_types/date.md
NotTheDr01ds: Thanks! Fixed locally and will resolve in next commit.

NotTheDr01ds:
So I don't forget to mention it later, I'm looking at replacing the `realpath` example in the `path` document because it doesn't really do a `realpath`.  Linux `realpath` is primarily used to resolve symlink targets, which that example doesn't do.

If I'm misunderstanding something there, do let me know!

NotTheDr01ds:
Also removing this example:

> This code snippet shows how to detect the error data type from within Nushell:
>
> ```nu
> do { error make {msg: 'can we see this?'} } | describe
> error
> ```

This no longer works - The pipeline is terminated and never reaches `| describe` (nor `| inspect`, if you (like me) had that thought).

book/types_of_data.md
```diff
@@ -1,23 +1,24 @@
...
+int
 ```
 
 ## Types at a glance
 
 | Type              | Example                                                               |
 | ----------------- | --------------------------------------------------------------------- |
 | Integers          | `-65535`                                                              |
-| Decimals (floats) | `9.9999`, `Infinity`                                                  |
+| Floats (decimals) | `9.9999`, `Infinity`                                                  |
```
sholderbach, line 21: I noticed recently: we lack cell path in this overview.

NotTheDr01ds: True - I'll add it in.  It's in the Language Guide as well as the "Navigating and Accessing Structured Data", but there should be a section the "Types of Data" chapter to point to those.

NotTheDr01ds:
Latest commit finishes up the remaining types and tries to standardize the contents of each type as much as possible.  Should be ready for review.  I'm sure there's a few changes needed, but hopefully we're close.

book/types_of_data.md
NotTheDr01ds: Oh wait - It is there.  It's just in "other types".  I've moved it and `closure` from there to "Basic Types".

NotTheDr01ds: Oh ... sorry, the *table* at the top.

fdncred:
Let's go ahead and land this. It's so much work. Nice job @NotTheDr01ds!!!! As with all our docs, if we missed something or something needs to be tweaked we can still do it. It's just software.

PR #1495, opened 2024-08-04
Fix line numbers having different line height in code blocks

danyx23:
The index.scss style sheet override the line height for codeblocks to 1 but it does not do the same for the line numbers at the left gutter. This commit fixes this.

grafik


fdncred:
let's try it. thanks!

PR #1498, opened 2024-08-07
Fix missing links due to changed command names

NotTheDr01ds:
The Imperative Languages map had two missing links.  Updated/fixed these:

* `count` command is now `length`
* `split-row` should be `split_row`


fdncred:
Good catch. Thanks!

PR #1499, opened 2024-08-07
Remove dead snippets from translated Installation chapters

NotTheDr01ds:
Vuepress was giving several warnings on startup of the dev server related to a missing snippet:

```
/snippets/installation/cargo_install_nu_more_features.sh
```

Running the translations through ChatGPT, this was in the (removed from the English version) section on the `--features` flag which isn't needed currently.  I've removed the relevant sections on building with `--features=all` and `--features=dataframe`.

> Wenn mehr Funktionalitäten installiert werden sollen, kann der folgende Befehl verwendet werden:

That is:

> If you want to install more functionalities, you can use the following command:

Removed the offending sections from German, Japanese, and Chinese translations.

The guides are probably desperately out of date in other areas, of course, and probably need some automated "warning" inserted, but at least we don't get any more "missing snippet" related warnings on the server-side.


fdncred:
sounds good. thanks.

Issue #84, opened 2021-01-07
Make `each` easier to find.


Ryan-McLemon:
When trying to find an analog for a Bash `for`, I searched for `for`, `loop`, and `iteration` in the Nushell docs, but came up short ... can we add something to make it a bit easier to find `each`?

NotTheDr01ds:
This has been implemented and should be closeable - *"for"*, *"loop'*, and *"iterate"* are all search terms for `each` now:

```nu
help -f iterate
```

Returns - `each`, `each while`, and `items`.

I'll also add a mapping in the "Coming from Bash" chapter of the book in a few minutes, but even without that, I'd call this one "done", right?

PR #1500, opened 2024-08-07
Bash `for` command mapping

NotTheDr01ds:
Adds `each` and `for` Nushell examples for the Bash `for` command

Closes out #84 for sure ;-)


sholderbach:
Good addition to have some language features in there!

Issue #1072, opened 2023-09-21
Mention `!...` history shortcuts in "coming from..." pages


BatmanAoD:
The `!!`, `!$`, `!<number>`, `!-<number>`, and possibly other history shortcuts work in Reedline roughly the way they do in Bash (though Bash's behavior is complicated and it seems some features are not supported, such as substitutions with `s^...^...`). This is probably worth mentioning in the "coming from ... " pages for different shells.

PowerShell's equivalent for `!$` is `$$`; I'm not sure if it has an equivalent for the others. I don't think cmd.exe supports this feature at all.

fdncred:
Good idea. They should be mentioned.

BatmanAoD:
One major difference is that by default, `Enter` performs the substitution rather than running the command as usual. In Bash, you can configure `Space` to do this (see [this blog post](https://blog.sanctum.geek.nz/bash-history-expansion/#comment-4106)).

fdncred:
ya, that would be cool to have space expand it. i really wish we could combine this stuff with an fish abbrev-like functionality. they're very similar, one gets that last commands and one just expands aliases.

orr721:
Is it possible to map !$ to some other key combination? Unfortunately on my keyboard layout to execute `!$<Enter>` requires 5 key presses (shift to get !, alt to get $). I am longing for bash `<Esc>.` all the time (2 key presses). A user defined combination would be great..

sholderbach:
> I am longing for bash `<Esc>.` all the time (2 key presses). A user defined combination would be great..

Worth checking the https://github.com/nushell/reedline/issues and adding a feature request there, maybe someone is interested in figuring this out :)

fdncred:
@orr721 I believe this works
```
  {
    name: insert_last_token
    modifier: alt
    keycode: char_.
    mode: [emacs vi_normal vi_insert]
    event: [
      { edit: InsertString, value: " !$" }
      { send: Enter }
    ]
  }
```

orr721:
yes! I'm on a mac though, so I had to change the modifier to control and keycode to other key. but finally it works, thank you!👍

fdncred:
> yes! I'm on a mac though, so I had to change the modifier to control and keycode to other key. but finally it works, thank you!👍

been there. done that. i use a mac too during the evenings and weekends. keybindings are sometimes hard to get recognized. it's why i wrote `keybindings listen` and someone else wrote `input listen`.

[...]

PR #1502, opened 2024-08-08
History and keybindings in Coming from Bash

NotTheDr01ds:
Implemented suggestion in #1072:

* Added `!!`, `!$`, `!<n>`, and `!<-n>` history substitutions
* Also added a few Bashism keybindings and pointed to the Reedline chapter for the rest.

### DIFF
--- a/book/coming_from_bash.md
+++ b/book/coming_from_bash.md
@@ -78,7 +78,7 @@ $env.Path = ($env.Path | prepend 'C:\Program Files\Git\usr\bin')
 | Bash                                 | Nu                                                            | Task                                                              |
 | ------------------------------------ | ------------------------------------------------------------- | ----------------------------------------------------------------- |
 | `!!`                                 | `!!`                                                          | Repeat last commandline from history                              |
-| `!$`                                 | `!$`                                                          | Repeat last commandline from history                              |
+| `!$`                                 | `!$`                                                          | Repeat last spatially separated token from history                |
 | `!<n>` (e.g., `!5`)                  | `!<n>`                                                        | Repeat \<n\>th command from the beginning of the history          |
 |                                      |                                                               | Tip: `history \| enumerate \| last 10` to show recent positions   |
 | `!<-n>` (e.g., `!-5`)                | `!<-n>`                                                       | Repeat \<n\>th command from the end of the history                |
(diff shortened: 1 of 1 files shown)


fdncred:
You can reference this link https://www.nushell.sh/blog/2024-05-15-bashisms.html and I think there may be one missing.

NotTheDr01ds:
Thanks - Definitely missed one.  I forgot about the blog post and was just going from what I could grok from the source ;-)

Are you saying that I should put a link near the table in the Book to the Blog post?  Or just that I could use it for my own reference?

fdncred:
I'm fine with putting the blog information in the docs, no need to link to it.

NotTheDr01ds:
Sounds good - Should be ready to merge if the update for `!<string>` looks good.

book/coming_from_bash.md
```diff
@@ -69,3 +73,18 @@ $env.Path = ($env.Path | prepend 'C:\Program Files\Git\usr\bin')
 | `pwd`                                | `$env.PWD`                                                    | Display the current directory                                     |
 | `read var`                           | `let var = input`                                             | Get input from the user                                           |
 
+## History Substitutions and Default Keybindings:
+
+| Bash                                 | Nu                                                            | Task                                                              |
+| ------------------------------------ | ------------------------------------------------------------- | ----------------------------------------------------------------- |
+| `!!`                                 | `!!`                                                          | Repeat last commandline from history                              |
+| `!$`                                 | `!$`                                                          | Repeat last commandline from history                              |
```
fdncred, line 81: This description is wrong. `Get the last spatially separated token from the last command in your history.` from my blog post.

NotTheDr01ds:
Thanks - Copy/paste error

book/coming_from_bash.md
fdncred: Now that I look at these again, they don't repeat anything. They just get the info and put it on the command line. Bash repeats. Nushell just gets but doesn't execute.

fdncred:
Now that I look at these again, they don't repeat anything. They just get the info and put it on the command line. Bash repeats. Nushell just gets but doesn't execute.

NotTheDr01ds:
How about "substitute" in place of "repeat"?

fdncred:
I think the key point that we need to make is that it does not work the same as bash. In bash you type !!<enter> and it runs the last command. I nushell you do !!<enter> and it puts the last command on the commandline but does not execute it. I'm not sure substitute or repeat convey what it does. Get is closer but it's not quite that either. Just pick one and put some blurb that describes the functionality, and I'll land it. Thanks for helping!

NotTheDr01ds:
*"Insert"* - I think that's the word we're looking for.  Updated.

Issue #1044, opened 2023-09-04
There seems not to be any document about path as a type.


otakutyrant:
Altough it is confirmed that [path is a type and a command](https://github.com/nushell/nushell/discussions/10203) simutaneously. [Types of Data](https://www.nushell.sh/book/types_of_data.html) does not include path, and [path](https://www.nushell.sh/commands/categories/path.html) only regards itself as a command.

fdncred:
Technically `path` is not a type. You can see the nushell types here:

https://github.com/fdncred/nushell/blob/main/crates/nu-protocol/src/ty.rs#L9-L37

It is a `SyntaxShape` as you can see here:

https://github.com/fdncred/nushell/blob/main/crates/nu-protocol/src/syntax_shape.rs#L7-L119

and here:

https://github.com/fdncred/nushell/blob/main/crates/nu-parser/src/parser.rs#L2685-L2765

I'm not sure all of these can be used as "types" in custom commands. I thought there was another list for custom commands but I can't seem to find it now. Maybe someone else has an idea.

otakutyrant:
The real X question is that I want to define the shape of an argument of an extern command `stow` as directories only so that I can complete the command with directories names only. So I want to check the document of the "type" `path` to see whether I can find the appropriate type.

PR #1505, opened 2024-08-12
Rework Completions and Externs chapters

### LINKED ISSUE #788: Document arguments to autocomplete commands
The documentation on [completions](https://www.nushell.sh/book/custom_completions.html) has no documentations or examples on completions that the arguments passed to autocomplete commands for providing textual completion, or even suggest that they exist. In order to write a completer that makes use of them, I first had to experiment to find out what they were.  An example that uses the arguments to produce the completion list would be helpful. Here's a possible example. I'm happy to submit a PR for it if it's useful.

```
let places = {
	usa: [ut, ca, co],
	canada: [bc, on, qc],
}

def country-autocomplete [args arglength] {
	$places | columns
}

def region-autocomplete [args arglength] {
	let parts = ($args | split words | skip 1)
	try {
		let country = $parts.0
		let result = ($places | get $country)
		$result
	} catch {
		let result = ($places | columns)
		$result
	}
}

# get the weather for a given country and region
export def "weather" [
	country: string@"country-autocomplete" # country name
	region: string@"region-autocomplete" # region code
] {
	print $"getting weather for ($region), ($country)"
}

```

NotTheDr01ds:
Started out implementing #1044 and #788.

* Moved Completions chapter ahead of Externs. This reduces duplication since Externs can take advantage of Completions.
* Extensive wording cleanup
* Document that completers must return a `list`
* Document that the default completer is for files/directories
* Document that the default completer can be suppressed using a command that returns an empty list
* Mention the menu-indicator and how to change it
* Clarify that changes to completers require reparsing the command-to-be-completed
* Added documentation on the `$position` argument (should close #788)
* Updated example to use cleaner `match` syntax
* Use <kbd>Tab</kbd> in place of `tab` (at some point I'd like to update the style on this)
* The External chapter used `git push` as the example, and said it was in the default config (which it isn't, at least not any longer) -- Changed example to be a simple `ssh` example that demonstrates just the key points, much more succinctly.
* Changed warning for `# description` to just a tip
* Added limitation note regarding caret sigil
* Added limitation note regarding `-longoption` syntax not being supported


book/custom_completions.md
```diff
@@ -83,7 +98,16 @@ Lulu                Enzo
...
 ```
 
-On the second line, once we press the `<tab>` key, the argument `"my-command dog"` is passed to the `animal-names` command as context.
+On the second line, after pressing the <kbd>tab</kbd> key, the argument `"my-command dog"` is passed to the `animal-names` completer as context.
+
+::: tip
+Completers can also obtain the current cursor position on the command-line using:
+
+```nu
+def completer[context:string, position:int] {}
```
fdncred, line 107: needs a space
```suggestion
def completer [context:string, position:int] {}
```

fdncred:
looks good, thanks! one small typo above.

NotTheDr01ds:
Thanks! Fixed.

fdncred:
Thanks

PR #1506, opened 2024-08-15
Remove outdated example for `touch -d`

weirdan:
Fixes nushell/nushell#8723


sholderbach:
We leave those files alone as they are auto generated and only modify the command source in nushell/nushell to tweak the documentation.

PR #1507, opened 2024-08-17
Typese - A few link corrections and added table-links

NotTheDr01ds:
Mainly was adding Cell-path to the overview table per @sholderbach suggestion in review, but also:

* Added links to each of the types in the table to their detail further down the page
* Fixed a few links
* Fixed the double-table header, but @fdncred had already gotten to that as well ;-)


fdncred:
Thanks for the cleanup

PR #1511, opened 2024-08-19
Fix 13541 description

NotTheDr01ds:
Slight correction on my PR 13541 - Integers passed to `into datetime` *without* a format string have always been interpreted as nanoseconds.  With this change, they are able to be interpreted based on the format string as well.

### DIFF
--- a/blog/2024-08-20-nushell_0_97_0.md
+++ b/blog/2024-08-20-nushell_0_97_0.md
@@ -141,7 +141,12 @@ After [#13612](https://github.com/nushell/nushell/pull/13612), `glob` now also t
 
 ### `into datetime` [[toc](#table-of-content)]
 
-Thanks to [@NotTheDr01ds](https://github.com/NotTheDr01ds) in [#13541](https://github.com/nushell/nushell/pull/13541), integer values can now be piped into `into datetime` in combination with a format string. The integers will be interpreted based on the format string.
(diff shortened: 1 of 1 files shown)


blog/2024-08-20-nushell_0_97_0.md
```diff
@@ -141,7 +141,7 @@ After [#13612](https://github.com/nushell/nushell/pull/13612), `glob` now also t
 
 ### `into datetime` [[toc](#table-of-content)]
 
-Thanks to [@NotTheDr01ds](https://github.com/NotTheDr01ds) in [#13541](https://github.com/nushell/nushell/pull/13541), integer values can now be piped into `into datetime` in combination with a format string. The integers will be interpreted as nanosecond-precision UNIX timestamps.
+Thanks to [@NotTheDr01ds](https://github.com/NotTheDr01ds) in [#13541](https://github.com/nushell/nushell/pull/13541), integer values can now be piped into `into datetime` in combination with a format string. The integers will be interpreted based on the format string.
```
sholderbach, line 144: Maybe worth restricting that this requires a format string accepting a single value, right?

sholderbach (commented):
If you have a small code example feel free to add it.

blog/2024-08-20-nushell_0_97_0.md
NotTheDr01ds: Yes, AFAIK it's only useful with one format string (`%s`) (but a common enough one that I felt it was worthwhile).  I'll add the example.

sholderbach: Thanks!

PR #1518, opened 2024-08-20
make-docs: Run in different Nu with plugins

### LINKED ISSUE #1516: make_docs doesn't work with cargo-installed Nushell in path
Just trying out `make_docs` for the first time and coming across a few issues.

I use `cargo install nu` on Linux after a new release to get the latest.  That means the Nushell in my path is:

```
~/.local/share/rust/cargo/bin/nu
```

This also means that there are no plugins in the same directory, unless I `cargo install` them as well.

So when running `make_docs`, I instead download a release package and run *that* Nushell, e.g. from:

```
~/nu-0.96.1-x86_64-unknown-linux-gnu/nu
```

This directory *does*, of course, have the plugins.

But `make_docs` is using the `nu` that is on the path rather than the *currently running* version. That results in `make_docs` simply erroring out since `command-names` is attempting to pass an empty plugins list to `nu --plugins`.

The solution seems simple here - Instead of using the current `which nu`, it should use `$nu.current-exe`.

I'm going to submit a PR in a bit, but I wanted to explain the issue (issues, really - There's another, I believe) first.

### LINKED ISSUE #1517: make_docs is not returning plugin commands unless they are installed
It looks to me like `make_docs` attempts (via `command-names`) to retrieve the core plugin commands regardless of whether they are installed or not.  However, this is not working, since the result essentially gets thrown away in the next line in `main` when it gets joined with the `scope commands` currently in scope.

This means, as far as I can tell, that help is only generated for plugins which are currently installed (even though there's code that seems to attempt otherwise).

NotTheDr01ds:
* Uses `$nu.current-exe` to determine the path of plugins rather than `which nu` - This fixes #1516 to allow users with Nu install by `cargo install` to still use `make_docs` with a different `nu`
* Moves plugin retrieval logic to a new function since it's now used in two places
* Slight refactor on plugin retrieval to use `each` rather than a mut-with-for
* Adds an *optional* helper command that will run `make_docs` in a new, temporary subshell with all core plugins installed, even if the user doesn't have them installed already. This implements/fixes #1517.

  ```nu
  source make_docs.nu
  make_docs
  # or
  make docs <path_to_nu>
  ```

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -1,21 +1,30 @@
-# get all command names from a clean scope
-def command-names [] {
-    const PLUGINS = [
-        nu_plugin_inc,
-        nu_plugin_gstat,
-        nu_plugin_query,
-        nu_plugin_polars,
-        nu_plugin_formats,
-    ]
-    let nu_dir = (which nu) | get path.0 | path dirname
-    mut plugins = []
-    for plugin in $PLUGINS {
-        if (sys host | get name) == 'Windows' {
-            $plugins ++= $'($nu_dir | path join $plugin).exe'
-        } else {
-            $plugins ++= $'($nu_dir | path join $plugin)'
+const PLUGINS = [
+    nu_plugin_inc,
+    nu_plugin_gstat,
+    nu_plugin_query,
+    nu_plugin_polars,
+    nu_plugin_formats,
+]
+
+def plugin-paths [ nu_path?: path ] {
+    # If no custom path is provided, default to the
+    # directory of the currently running nu
+    let nu_dir = match $nu_path {
+        null => ($nu.current-exe | path dirname)
+        _ => ($nu_path | path dirname)
+    }
+
+    $PLUGINS | each {|plugin|
+        match (sys host | get name) {
+            'Windows' => $'($nu_dir | path join $plugin).exe'
+            _ => $'($nu_dir | path join $plugin)'
         }
     }
+}
+
+# get all command names from a clean scope
+def command-names [] {
+    let plugins = (plugin-paths)
     nu --no-config-file --plugins ($plugins | to nuon) --commands $'scope commands | select name | to json'
         | from json
 }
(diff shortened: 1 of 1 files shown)


make_docs.nu
```diff
@@ -1,21 +1,31 @@
-# get all command names from a clean scope
-def command-names [] {
+
+def plugin-paths [ nu_path?: path ] {
     const PLUGINS = [
```
fdncred, line 3: These are still hard coded here. I thought we were going to use `plugin list` for these names?

fdncred:
Good work. Just had one question. I figure we'll wait on this since the release is today and we'll be running it soon. But I would defer to @hustcer since he runs these scripts after each release (and does a great job keeping up with it ❤️)

make_docs.nu
NotTheDr01ds: Right - We still do have to hardcode the "core" plugin names that we're going to document.  Otherwise we'd also load things like `nu_plugin_stress_internals`.  On the bright side, if any commands are added to the core plugins, they'll get added, documented, and "bannered" automatically.

But yes, if a new "core" plugin is added, removed, or renamed, we'll have to change these hardcoded values.

Your comment does make me think that perhaps `CORE_PLUGINS` would be a better moniker, though.

NotTheDr01ds:
Right - I was nervous about touching this at first since it felt like it was so "specific" to the @hustcer's command updates, but then I saw that quite a few people have contributed to it, so I went ahead.  Still, I definitely realize that this one is gating on @hustcer review.

> I figure we'll wait on this since the release is today and we'll be running it soon.

Understood. On the flip side, the banner update in #1519 is dependent on this work, so if we want to get the banners fixed on things like `inc` soon, it will need to be merged and then the command doc regenerated regardless.  Just up to Justin if it's better to run it once (with these changes) or twice.

hustcer:
Thanks, it works perfectly

PR #1519, opened 2024-08-20
Generate plugin banner for all plugin commands

### LINKED ISSUE #1515: Commands provided by a plugin should appear in the Commands doc differently
Currently [`inc`](https://www.nushell.sh/commands/docs/inc.html) appears without any indication that they require a plugin in order to work.  Probably need to modify the `make_docs.nu` to add something explaining this to the online-doc.

Edit: Looks like `make_docs` is already supposed to handle this, but it's not working for `inc` for some reason.

NotTheDr01ds:
Fixes #1515 

Previously, a few commands were hard-coded as being plugin commands, and a banner was generated for the command doc accordingly. This was out-of-date, as not even Polars was showing up properly.  There's still code to generate a warning for `dfr` commands.  (Oops - Forgot to take this out - Brb.) (Okay - Done)

After this change, `plugin list` is used to return a list of plugin commands, and if a command is in that list, the banner is added.

Example:


NotTheDr01ds:
Note: This is branched from #1518 - Best to merge that first if possible.

make_docs.nu
```diff
@@ -250,18 +273,23 @@ $"($example.description)
...
-        $"::: warning(char nl)Command `($command.name)` resides in [plugin]\(/book/plugins.html) [`nu_plugin_formats`]\(https://crates.io/crates/nu_plugin_formats). To use this command, you must install/compile and register nu_plugin_formats(char nl):::(char nl)(char nl)"
-    } else if $command.name in ['query', 'query xml', 'query json', 'query web'] {
-        $"::: warning(char nl)Command `($command.name)` resides in [plugin]\(/book/plugins.html) [`nu_plugin_query`]\(https://crates.io/crates/nu_plugin_query). To use this command, you must install/compile and register nu_plugin_query(char nl):::(char nl)(char nl)"
-    } else { '' }
+    let plugin_commands = (plugin list | flatten)
+    let plugin_warning = if ($command.name in $plugin_commands.commands) {
+        let plugin = ($plugin_commands | where commands == $command.name | first)
+        [ $"::: warning This command requires a plugin"
+          $"The `($command.name)` command resides in the `($plugin.name)` plugin."
+          $"To use this command, you must install and register `($plugin.filename | path basename).`"
```
hustcer, line 281: Should be `($plugin.filename | path basename)`. NOT `($plugin.filename | path basename).`

NotTheDr01ds: Great catch. Thank you. Just pushed the fix.

hustcer:
Thanks for the fix, the plugin docs have already been updated: https://www.nushell.sh/commands/docs/polars.html

PR #1524, opened 2024-08-27
Replace `it` with `elt` / `row` / `line` etc

weirdan:
`$it` is a special variable used in `where`, so using it as a closure parameter name may lead to confusion.

For example:

* nushell/nushell#9443
* nushell/nushell#9271

To reduce possible confusion I suggest to _not_ use `$it` as a closure parameter name in documentation examples.

Refs: https://discord.com/channels/601130461678272522/729071784321876040/1278075102633791520


fdncred:
Thanks

PR #1525, opened 2024-08-27
Replace `it` with `elt` / `row` / `line` - batch 2

weirdan:
I missed several instances in the previous PR


fdncred:
Thanks for the cleanup

PR #1528, opened 2024-08-28
Promotes Installation to its own top-level Book chapter

NotTheDr01ds:
Related to #1508 - This is in preparation for changing the Navbar so that we can link directly to "Install" and "Getting Started".  It just makes *"Installation"* it's own top-level chapter instead of being a sub-section of *"Getting Started"*.

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -9,12 +9,16 @@ export const sidebarEn: SidebarConfig = {
       collapsible: false,
     },
     {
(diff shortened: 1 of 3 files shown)


NotTheDr01ds:
Sorry, just realized that some wording needs to be changed in the Getting Started since it starts off with "Let's install".  Modifying and will remove Draft after I fix that.

NotTheDr01ds:
Okay, it should be ready.

fdncred:
ok, thanks

PR #1529, opened 2024-08-29
fix: use consistent headline capitalization

Hofer-Julian:
I used  ¨Chicago Manual of Style" in https://headlinecapitalization.com/


fdncred:
consistency is good. thanks!

PR #1531, opened 2024-08-29
fix: capitalization for headings level 2 and 3

Hofer-Julian:
This PR fixes the capitalization also for level 2 and level 3 headings


NotTheDr01ds:
FWIW, I vote "yes".  I'd been trying to keep them lowercase when editing/adding, just because most already were (consistency), but it's more readable and proper for them to maintain "title case".  So yes, I'd like this "across the board" fix to make things consistent.

If committed, this should also be a part of a Style Guide.

fdncred:
Thanks. Looks good.

PR #1535, opened 2024-08-30
Fix collapsed table example

abhbh:
The example output of expanded and collapsed table is same in https://www.nushell.sh/commands/docs/table.html


NotTheDr01ds:
Thanks - That looks like it's my fault.

Unfortunately, we can't modify *command* help here in this repo.  The [Commands doc](https://www.nushell.sh/commands/) is automatically generated using the built-in help, so this means that:

```
help table
```

is also wrong, and we'll have to fix this [in the source](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/viewers/table.rs).  I think this is probably due to a bad `table -e` that I put in, but more research is needed.

Since the `help table` result is going to overwrite your change, I'm going to close this PR but open a new issue on the main repo -[nushell #13735](https://github.com/nushell/nushell/issues/13735)

NotTheDr01ds:
Well, slight correction on the part above - `help table` is correct *because* of some code I added a few months ago, but the doc generation script uses `scope`, which wasn't fixed.

abhbh:
Thanks.

I should have realised that the docs were auto-generated and open an issue instead. 😅

NotTheDr01ds:
No worries at all! There's supposed to be a PR action that should have let you know automatically based on the file being changed, but I'm not exactly sure how it works or why it didn't in this case.

Regardless, thanks for the heads-up as well as the attempted fix!

PR #1536, opened 2024-08-30
Fix binary data type table content

yo-goto:
I corrected the table content because the binary data type's content was mistakenly replaced with the range data type's content.
The table content is pulled from `/lang-guide/chapters/types/basic_types/binary.md`.


fdncred:
Thanks
/cc @NotTheDr01ds

PR #1538, opened 2024-08-31
Fix link parentheses in Float type page

yo-goto:
I've noticed a broken link in the float type page, so corrected it.


fdncred:
Good catch. Thanks.

PR #1539, opened 2024-08-31
Fix nothing type literal: none → null

yo-goto:
I fixed wrong nothing type literal: `none` → `null`
This page says nothing type literal syntax is `null`.
https://www.nushell.sh/book/types_of_data.html#nothing-null


fdncred:
That's correct. Thanks for spotting it and fixing it.

PR #1540, opened 2024-09-02
Fixes broken links

lfrancke:
The breaking changes were split up at some point in two sections but the internal link wasn't updated. This is now pointing at the full list of breaking changes again.


sholderbach (approved):
Thanks for the fix.

I personally think that the `All breaking changes` section we have now just containing the commits is not that helpful and we should try to contain the small explanations of the changes under a `breaking changes` section going forward, so things that need manual updating are centralized.

lfrancke:
I can update the links to point to that other link if you prefer?

sholderbach:
That was more a comment for future authors of release notes.

Issue #1514, opened 2024-08-19
Document "default" plugins


NotTheDr01ds:
Things like `nu_plugin_inc` are not documented.  The plugins chapter should include a list of all plugins that are released with Nushell (but not enabled).

NotTheDr01ds:
Closed with the chapter update.

PR #1542, opened 2024-09-04
Plugin Chapter update

NotTheDr01ds:
* Also updates DataFrames chapter to remove erroneous installation information. Points back to Plugins chapter for installation.
* Moves the "implementation" paragraph from the top down to the developer section
* Adds a callout on version matching near the top
* Trying out the automatic `[[toc]]` here (and in DataFrames). This populates a table-of-contents with links to each heading. We can easily (and IMHO should) add this to more chapters in the documentation.
* Adds a section listing the Core plugins that ship with Nushell and how to use them (addresses #1514).
* Changes code blocks to remove `>` prompt to match style guide.
* Reworks heading titles and levels for consistency and flow
* Other miscellaneous improvements

### DIFF
--- a/book/dataframes.md
+++ b/book/dataframes.md
@@ -1,21 +1,10 @@
 # Dataframes
 
-::: warning
+::: warning Important!
+This feature requires the Polars plugin.  See the the 
+[Plugins Chapter](plugins.md) to learn how to install it.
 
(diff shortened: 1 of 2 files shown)


fdncred (approved):

fdncred:
There were some minor things that I was unsure about like "add a plugin" vs "register a plugin" but as a whole it's better than what we had. We can quibble about details some other time.

PR #1545, opened 2024-09-09
Add `wrap` as a cast for table and record

NotTheDr01ds:
Based on a question in Discord today, I realized that `wrap` (while incredibly useful) isn't well promoted.

I'll also look at adding some help keywords to `wrap`, but at the least, it should be listed as a way to cast to `table` and `record`.


fdncred:
I'm not sure I'd call it a cast but it's helpful either way.

PR #1546, opened 2024-09-11
`std` doc page

NotTheDr01ds:
* Adds some much-needed detail to the `std` doc
* Removes the link to the Nushell GitHub repo - The `README` there is out of date anyway, and the doc should be rendered by the SSG (VuePress) rather than in a GitHub Markdown file whenever possible.
* Might be controversial, but I note the performance issues that prevent new commands from being added to `std` at the moment.  I think it's a good explanation give the current state, and it doesn't come across too harsh, hopefully.  I can remove it if needed.

### DIFF
--- a/book/standard_library.md
+++ b/book/standard_library.md
@@ -1,3 +1,45 @@
-# Standard library (preview)
+# Standard Library (Preview)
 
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #1556, opened 2024-09-19
Documented the recommended MIME type for Nushell

cptpiepmatz:
In the [meeting of the 18.09.2024](https://hackmd.io/@nucore/H1bgUtLaC) we talked about adding recommended MIME types for Nushell to the documentation. This is the PR for that.

I added it to the `Language Reference Guide` as this part seems to more about the internals and not so much about using Nushell, so I though it would fit there. If not, let's change that.


fdncred:
LGTM! Thanks!

PR #1557, opened 2024-09-20
Remove outdated mention of $it

nome:
You can't use implicit `$it` with `each` in current versions of nushell any more.


fdncred:
right, $it only works with where these days. thanks

PR #1558, opened 2024-09-20
`coming_from_bash.md`: add references to `explore` and `first`

paulie4:
I think knowing `explore` and `first` are extremely helpful for *nix users that are used to using `less` and `head`, so I think they should be included in the [Coming from Bash](book/coming_from_bash.md) page. I also think there should be more clarification about which things are useful for external versus internal commands.


fdncred:
sounds good. thanks!

PR #1563, opened 2024-09-27
Unify the language identifiers of the code blocks to 'sh'

ghaaj:
This PR unifies the language identifiers of the code blocks for installation commands to 'sh'.
Syntax highlighting will be applied to the modified code blocks, and the extra $ sign at the beginning will be removed when copied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,8 @@ $ nix profile install nixpkgs#nushell
 
 #### Windows:
 
-```shell
-$ winget install nushell
+```powershell
+winget install nushell
 ```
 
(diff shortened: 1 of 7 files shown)


fdncred:
I'm not sure this is a good change because this is a windows only section and probably no one on windows is running `sh` but it is likely they're running `powershell`.

ghaaj:
Thank you for reviewing. Would the preferred choice then be to keep the language identifier of the `powershell` section as is and delete the $ sign?

fdncred:
ya, good idea

fdncred:
Thanks

PR #1564, opened 2024-09-28
One-liner for installing plugins with Cargo

NotTheDr01ds:
Just a quick update:

* Update installation chapter to note that plugins need to be installed separately with Cargo
* Add a one-liner in the Plugins chapter to install all the default plugins at once

### DIFF
--- a/book/plugins.md
+++ b/book/plugins.md
@@ -28,6 +28,18 @@ Core plugins are typically distributed with the Nushell release and should alrea
 
 ::: tip Installing using Cargo
 For example, when installing or upgrading Nushell directly from crates.io using `cargo install nu`, the corresponding core plugins for that version may also be installed or updated using `cargo install nu_plugin_<plugin_name>`.
+
+To install all of the default plugins, from within Nushell run:
+
+```nu
+[ nu_plugin_inc
+  nu_plugin_polars
+  nu_plugin_gstat
+  nu_plugin_formats
+  nu_plugin_query
+] | each { cargo install $in; $in }
+```
+
 :::
 
 ### Third-party Plugins
--- a/book/plugins.md
+++ b/book/plugins.md
@@ -37,7 +37,7 @@ To install all of the default plugins, from within Nushell run:
   nu_plugin_gstat
   nu_plugin_formats
   nu_plugin_query
-] | each { cargo install $in; $in }
+] | each { cargo install $in } | ignore
 ```
 
 :::
--- a/book/installation.md
+++ b/book/installation.md
@@ -6,6 +6,8 @@ The main Nushell binary is named `nu` (or `nu.exe` on Windows). After installati
 
 @[code](@snippets/installation/run_nu.sh)
 
+[[toc]]
+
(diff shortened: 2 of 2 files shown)


book/plugins.md
```diff
@@ -28,6 +28,18 @@ Core plugins are typically distributed with the Nushell release and should alrea
...
+
+To install all of the default plugins, from within Nushell run:
+
+```nu
+[ nu_plugin_inc
+  nu_plugin_polars
+  nu_plugin_gstat
+  nu_plugin_formats
+  nu_plugin_query
+] | each { cargo install $in; $in }
```
fdncred, line 40: what's the point of running $in. shouldn't we be doing `plugin add $in`?

NotTheDr01ds: Just to display/return the results, rather than an empty list.  Could do an `ignore` or `for` instead; it just seemed like this was cleaner.

fdncred: If you want to give feedback then I think print would be better. I think running the plugin gives the wrong idea because we really don't recommend running them as nu_plugin_name except for diagnostic purposes.

NotTheDr01ds: > we really don't recommend running them as nu_plugin_name except for diagnostic purposes.

`$in` doesn't run the plugin ;-).  It simply populates a list with the *name* of the plugin that was passed in.

Take out the `cargo install` statement, and it's like saying:

```nu
[ 1 2 3 ] | each { $in }
```

Are you okay with it in that case?

I have to retract what I said above, too - a `print` won't solve it, since it would still generate the empty list as well.  Here's what it would look like:

```nu
[ nu_plugin_inc
∙   nu_plugin_polars
∙   nu_plugin_gstat
∙   nu_plugin_formats
∙   nu_plugin_query
∙ ] | each { print $'Installing ($in)'; cargo install $in }
Installing nu_plugin_inc
    Updating crates.io index
     Ignored package `nu_plugin_inc v0.98.0` is already installed, use --force to override
Installing nu_plugin_polars
    Updating crates.io index
     Ignored package `nu_plugin_polars v0.98.0` is already installed, use --force to override
Installing nu_plugin_gstat
    Updating crates.io index
     Ignored package `nu_plugin_gstat v0.98.0` is already installed, use --force to override
Installing nu_plugin_formats
    Updating crates.io index
     Ignored package `nu_plugin_formats v0.98.0` is already installed, use --force to override
╭───┬──╮
│ 0 │  │
│ 1 │  │
│ 2 │  │
│ 3 │  │
╰───┴──╯
Installing nu_plugin_query
    Updating crates.io index
     Ignored package `nu_plugin_query v0.98.0` is already installed, use --force to override
╭───┬──╮
│ 4 │  │
╰───┴──╯
```

And a `for` can't be cleanly (IMHO) done as a one-liner.

I'll update to `ignore`, I guess.

NotTheDr01ds:
@fdncred Are you good with the change to `| ignore`?  If so, this is ready to go.

fdncred:
sure. thanks!

Issue #1552, opened 2024-09-17
Add `def --env` to Custom Commands chapter


NotTheDr01ds:
`def --env` isn't mentioned in the [Custom Commands](https://www.nushell.sh/book/custom_commands.html) chapter at all at the moment.  It's in "Nu as a Shell", under [Environment](https://www.nushell.sh/book/environment.html).

TODO:  At least create a reference linking from Custom Commands to that reference in Environment.  Also consider just moving it entirely, as it's more about "Programming" than "Nu as a Shell'.

PR #1565, opened 2024-09-29
Custom Command chapter updates

NotTheDr01ds:
* Implements #1552 by adding information on `def --env` to the Custom Commands chapter.  Moves some of it from the Environment chapter, but links between the two.
* The "Command Signature" section was misplaced below quite a few other topics. Moved it into the Custom Commands chapter.
* Extensive rework/rewording of much of the Custom Commands chapter.
  * Starts out with an explanation of why we call them "custom commands"
  * (Nit-pick) *"Hello, <name>"* examples have proper capitalization and punctuation.
  * Adds information on implicit return.  Before someone asks (because I did) - No, Rust does not have implicit return. It simply enforces *expressions*, which we don't.
  * Adds additional information and examples on return values - `match`, early `return`, `ignore`, and returning `null`.
  * Adds information on *invalid* command names
  * Groups all parameter information into an H2 for ToC purposes.
  * Oh yeah, adds a ToC since this is a long chapter
  * Adds information on multi-parameter commands
  * Adds information on non-type shapes that can be used for annotations (e.g., `directory`)
  * Uses multiple names from many countries as `greet <name>` examples
  * Adds a few *"Cool!"* callouts for features that should "Wow" the user (e.g., automatic `help` for custom commands).
  * Updates table examples to current default style.


book/custom_commands.md
```diff
@@ -219,31 +404,38 @@ def greet [
   name: string
   --age: int
 ] {
-  [$name $age]
-}
+    {
+      name: $name
+      age: $age
+    }
+  }
```
132ikl, line 411: This last line is indented a bit, which seems to be inconsistent with future definitions (also present in the next definition of `greet` below which demonstrates shorthand flags)

book/custom_commands.md
```diff
@@ -252,187 +444,350 @@ def greet [
...
+let guests = [ Dwayne, Shanice, Jerome ]
+vip-greet $vip ...$guests
+# => Hello, Dwayne!
+# => Hello, Shanice!
+# => Hello, Jerome!
+# => And a special welcome to our VIP today, Tanisha!
 ```
 
-## Documenting Your Command
+## Input-Output Signature
```
132ikl, line 587: Would it be possible to name this "Input-Output Pipeline Signature" or "Pipeline Input-Output Signature"? I know I've tried looking for this section before by searching for "pipeline signature", and it could make it a bit clearer what it's referring to

book/custom_commands.md
```diff
@@ -252,187 +444,350 @@ def greet [
...
+  cd ~
+}
+
+cd /
+go-home
+pwd
+# => Your home directory
+```
+
+## Custom Commands and Pipelines
```
132ikl, line 784: I think this section should be moved above the "Input-Output Signature" section, since it's not clear what that section is referring to without the context of this section

132ikl (commented):
Thank you for this, this section has definitely needed some work :smile:

NotTheDr01ds:
@132ikl Thanks! Converting to a draft while I revise based on your suggestions.

book/custom_commands.md
NotTheDr01ds: In upcoming commit

book/custom_commands.md
NotTheDr01ds: Thanks - Thought I'd caught all of them, but missed that one.  Completed in the upcoming commit.

132ikl (approved):

fdncred:
lgtm!

PR #1568, opened 2024-09-30
Documentation for sorting

132ikl:
Companion PR to https://github.com/nushell/nushell/pull/13154. Should not be merged until the nushell PR is merged.

Adds a new page to the book explaining how Nushell's many different sorting behaviors work, including old features and new features.


book/sorting.md
```diff
@@ -0,0 +1,292 @@
...
+╰───┴────────╯
+```
+
+We can see that the numbers are sorted in order, and the strings are sorted to the end of the list, also in order. If you are coming from other programming languages, this may not be quite what you expect. In Nushell, as a general rule, **data can always be sorted without erroring**.
+
+::: tip
+If you _do_ want a sort containing differing types to error, see [strict sort](#strict-sort).
+:::
+
+Nushell's sort is also **stable**, meaning equal values will retain their original ordering relative to each other. This is illustrated here using the [insensitive](#insensitive-sort) sort option:
```
sholderbach, line 45: Do we want to commit to all our sorting implementations being stable or should that be relaxed to `sort` or "stable by default"?

For clarity I would specify the insensitive
```suggestion
Nushell's sort is also **stable**, meaning equal values will retain their original ordering relative to each other. This is illustrated here using the [case insensitive](#insensitive-sort) sort option:
```

book/sorting.md
```diff
@@ -0,0 +1,292 @@
...
+│ 2 │      4 │
+│ 3 │      6 │
+│ 4 │      8 │
+│ 5 │      9 │
+│ 6 │ foobar │
+│ 7 │ hello  │
+╰───┴────────╯
+```
+
+We can see that the numbers are sorted in order, and the strings are sorted to the end of the list, also in order. If you are coming from other programming languages, this may not be quite what you expect. In Nushell, as a general rule, **data can always be sorted without erroring**.
```
sholderbach, line 39: Some part of my inner "language patent lawyer" thinks that promising this leads to some tricky rules to define a total order over all data types with challenges as soon as new types would be introduced and are messy to teach if you run into them. But for the general usability it is definitely a bonus if you don't have to worry about this.

sholderbach (commented):
Thanks for being so thorough about all those semantics.

Maybe some of it is very thorough for end-user documentation but I think it is good if we start out with well documented/specified semantics. We can still move the details to the reference material and have a simpler intro at a later point in time.

book/sorting.md
132ikl: `sort` and `sort-by` both use Rust's `sort_by`, which is stable. I also have a unit test which does something similar to the example here which demonstrates the sort being stable. I'm not sure what other sorting implementation there might be?

I agree, I'll rename the heading to "case insensitive" as well.

book/sorting.md
132ikl: At least currently, there is indeed an ordering between all types. It's effectively the order of the `Value` enum's variants, with some exceptions. As I see it, the tradeoff here is between "sort which errors on mixed types" or "sort that never errors". Personally, if it would be possible for sort to error because of incompatible types, I would rather make it fully strict at that point.

I wasn't planning on specifying what order the types would be, other than that within the same version of Nushell the ordering between types is always the same. I would like to guarantee at least that `null` comes at the end, and the corresponding Nushell PR does do this. Since Nothing is kind of a special type, and you might expect it to appear a lot more than say, an integer and a string, I think it's a useful property for it to always appear at the end.

book/sorting.md
```diff
@@ -0,0 +1,292 @@
...
+```
+
+::: tip
+The parameters are also passed to the custom closure as a two element list, so the following are equivalent:
+
+- `{|a, b| $a < $b }`
+- `{ $in.0 < $in.1 }`
+  :::
+
+Here's an example of a custom sort which couldn't be trivially written as a key sort. In this example, we have a queue of tasks with some amount of work time and a priority. We want to sort by priority (highest first), but if a task has had zero work time, we want to schedule it immediately. Otherwise, we ignore the work time.
```
132ikl, line 231: TODO explain that priority is not considered for work_time = 0 items

devyn:
Is this ready to go? Now that 0.99.0 has released we can merge

book/sorting.md
132ikl: Replaced if-else sort with match case which covers this case explicitly

132ikl:
@devyn Just finished up everything I was planning :smile: 

@sholderbach What are your thoughts on the semantics you brought up? I also added some more info about how the ordering between types works in the "Sorting with mixed types" section, I would appreciate your feedback on that as well!

book/sorting.md
```diff
@@ -0,0 +1,421 @@
...
+To instead sort in order of a record's values, use `sort -v`:
+
+```nu
+> {foo: 123, bar: 456, baz: 89} | sort -v
+╭─────┬─────╮
+│ baz │ 89  │
+│ foo │ 123 │
+│ bar │ 456 │
+╰─────┴─────╯
+```
```
IanManske, line 190: Duplicate section here

book/sorting.md
IanManske: I think it's unlikely that we will add more types (I would probably be against it). For the current set of types, a total order is totally doable. Floats are really the only issue, and could be ordered using [`total_cmp`](https://doc.rust-lang.org/std/primitive.f64.html#method.total_cmp) or sorting NaNs last like [`ordered_float`](https://docs.rs/ordered-float/4.4.0/ordered_float/struct.OrderedFloat.html).

132ikl: Oh, good point. I might look into adding a total order 🤔

book/sorting.md
132ikl: Fixed, thank you!

132ikl:
I'd like to get this landed, @sholderbach if you are still concerned about those sections you commented on I can modify them or outright remove them. Let me know what you think 😄

sholderbach (approved):
Fine with the write up thanks! Sorry for not getting back.

Have some vague thoughts about forward compatibility around any type extensions but those are future problems 😄

fdncred:
I just wanted to drop by and say that this sorting documentation is completely awesome! I'm impressed by the thoroughness and all the details but keeping it very approachable with samples. Great work!

132ikl:
Thanks so much, I appreciate it 😄

PR #1569, opened 2024-10-01
Convert blockquotes to callouts

NotTheDr01ds:
After @132ikl's updates to the Modules chapter, I decided to check if there were any other blockquotes hanging around. Found and fixed a few more in other chapters.


hustcer:
Thanks for the improvement

PR #1570, opened 2024-10-01
Fix broken link from Custom Commands update

NotTheDr01ds:
The "Input-Output Signature" chapter was moved into "Custom Commands", which resulted in a broken-link-warning. Since this is no longer a top-level chapter, I removed it from the "Overview" page.  Also fixed a minor typo (I think it was mine in the first place) while I was there.


sholderbach:
Thanks!

PR #1572, opened 2024-10-01
Document `wrap` behavior with index

AlifianK:


fdncred:
Thanks, but it's the same for this command. It must be changed in the rust code to show up in the docs.

NotTheDr01ds:
Also a good point, at least that consecutive, 1-based numbers are probably a bad example to use for `wrap`, at least.  I wish I'd seen this before I committed the other.

I think for this one, I'll do something a bit different, though:

```nu
[ Pachisi Mahjong Catan Carcassonne ] | wrap games
```

As for the `index` column name being magic - I agree this is confusing.  See my comment in your main repo issue.  @sholderbach has suggested *potentially* removing the "auto-renaming" feaure.

PR #1576, opened 2024-10-04
add Windows command_not_found example

chapmanjacobd:
This PR adds an additional example for Windows to the existing command_not_found hook documentation.

For example, I used this to find the program path of excel after typing `excel` resulted in `command_not_found`


fdncred:
cool! Thanks for the addition!

PR #1578, opened 2024-10-06
std-lib breaking changes

NotTheDr01ds:
Adds breaking change information regarding `std` to 0.99.0 release notes


fdncred:
Thanks

Issue #1573, opened 2024-10-01
Line numbers in code blocks


NotTheDr01ds:
Our current config for Vuepress is set for [`lineNumbers: false` in the config](https://github.com/nushell/nushell.github.io/blob/afc86c2a63ff645a3cf6e7d02e3b970e014b9b71/.vuepress/config.js#L96).

But that's not working, which looks pretty odd in some cases, primarily one-liners:

![image](https://github.com/user-attachments/assets/529d2f56-e1be-42f3-93c2-b79d3645d542)

I think there are two reasons that line numbers aren't currently disabled:

1. First, when we switched to Shiki, the configuration option for turning off line numbers changed, and it wasn't updated in the config.

2. But more importantly, it looks like there's an upstream problem where the line number setting for the Shiki plugin is completely broken.  Attempting to set [`lineNumbers: false` or `lineNumbers: disabled`](https://ecosystem.vuejs.press/plugins/markdown/shiki.html#linenumbers), even in the Shiki plugin options, results in line numbers for everything *except* the last line of a code block.

   I've even tested this in a fresh Vuepress site unrelated to Nushell:
   
   ![image](https://github.com/user-attachments/assets/cec77210-f161-4a24-8bb9-cec6c4550e93)

   Of course, this appears to be an upstream issue, but I'm surprised something so fundamental could be

   There's also an option to only turn on line-numbers when the code-block is more than `x` lines, but that is just as broken.

   I'm wondering if there's a conflict between prismjs and Shiki, since the code-blocks (even on our site) still show both a `div` with a `prismjs` attribute and a separate `pre` block with a `shiki` attribute.

Anyway - If anyone knows what is going on here, feel free to chime in.

In the meantime, I'll look at other options, such as disabling for each block (still works) or doing it with custom CSS rules.

PR #1579, opened 2024-10-06
Turn back on line-numbers

NotTheDr01ds:
While I would certainly love to turn line-numbers off (or at least only use them when there's a certain number of lines in the code-block), as mentioned in #1573 this just doesn't seem to work in Vuepress at the moment when using Shiki.  Attempting to turn them off makes the problem *worse* by only removing the line-number from the last line, while keeping the rest:

image

This PR turns them back on to at least fix that bug.  I don't yet know what the long-term solution should be.

It seems odd to me that this is broken at the Vuepress/Shiki upstream, but as mentioned in #1573, I reproduced it in a clean environment.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -172,7 +172,7 @@ export default defineUserConfig({
     }),
     shikiPlugin({
       theme: 'dark-plus',
-      lineNumbers: false,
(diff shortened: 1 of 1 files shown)


hustcer:
Thanks,let's try it

PR #1583, opened 2024-10-09
Added changes to release notes for version 0.99.0 due to nushell/#14020

JustForFun88:
This PR includes details about Unicode code points in keybindings, as requested in [the comment](https://github.com/nushell/nushell/pull/14020#issuecomment-2399909753). The changes are based on the improvements introduced in [#14020](https://github.com/nushell/nushell/pull/14020).

### DIFF
--- a/blog/2024-10-15-nushell_0_99_0.md
+++ b/blog/2024-10-15-nushell_0_99_0.md
@@ -27,6 +27,7 @@ As part of this release, we also publish a set of optional plugins you can insta
 - [_Highlights and themes of this release_](#highlights-and-themes-of-this-release-toc)
 - [_Changes_](#changes-toc)
   - [_Additions_](#additions-toc)
+    - [_`enhanced keybinding configuration`_](#enhanced-keybinding-configuration-toc)
   - [_Breaking changes_](#breaking-changes-toc)
   - [_Deprecations_](#deprecations-toc)
   - [_Removals_](#removals-toc)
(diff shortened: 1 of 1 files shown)


blog/2024-10-15-nushell_0_99_0.md
```diff
@@ -70,6 +71,32 @@ As part of this release, we also publish a set of optional plugins you can insta
 
 ## Additions [[toc](#table-of-content)]
 
+### `enhanced keybinding configuration` [[toc](#table-of-content)]
+
+**Support for Unicode Code Points in Keycodes.**
+
+Starting from [#14020](https://github.com/nushell/nushell/pull/14020), users can now specify characters in the keycode field using their Unicode code points, such as `char_u003B` for the semicolon (`;`). This provides greater flexibility when configuring keybindings in Nushell.
+
+Previously, it was possible to specify characters by enclosing them in quotes, like keycode: "char_;". However, this approach was inconsistent for all Unicode characters, particularly when distinguishing between uppercase and lowercase Latin letters, although it worked for non-Latin characters. The new feature allows for a more straightforward and universally applicable way to define keybindings using Unicode code points.
```
sholderbach, line 80: This attempt at explaining reasoning makes me even less convinced that this feature should be there.

The example of the `;` is a bad explanation because it dances around the point the only limitation before was a poor understanding of the core tenants of nushells string syntax.
That we accept strings without quotes in our record syntax is a courtesy we provide (and an attempt to transfer the logic for (external) command calls) and depends on being free of otherwise confusable characters, you can't include `:` or `,` in a record bare string either. If our default config would use strings with quoting for the keycode entries it wouldn't have come up. 
We should strive towards making it easier for our users to grasp more of the transferrable base concepts of Nushell they can apply in many other situations than adding more things you have to understand to paper over things that were unclear before. 

Saying that you get a benefit in the capitalization is also questionable, the capitalization of non-ASCII alphabetic characters already mattered as well as the capitalization of the ASCII alphabetics without modifiers (`CTRL`, `ALT` etc.). 
Passing a specific capitalization for the ASCII alphabetics with added modifiers (even through their codepoint id) still doesn't matter because [reedline will coerce the ASCII capitalization for the keys with modifier](https://github.com/nushell/reedline/blob/871075e62b5703cfc584e4e02b14c2473f5b0b63/src/edit_mode/emacs.rs#L133-L135). For all intents and purposes, that we coerce here doesn't really matter if users follow the output of `keybindings listen` in setting up their keybindings. It shows which modifiers will be active and the resulting binding will work. (we could clarify the behavior when improving our documentation there, but being lax there doesn't have much harm if you want to enforce some binding with specific modifiers anyways.)

Adding things to our documentation that can be misunderstood or are a product of misunderstanding doesn't help. I would rather not have those two sentences.

JustForFun88: Thank you for your feedback. I can remove the two sentences as you've suggested. However, would it be acceptable to mention that the new ability to specify Unicode code points in keycodes can be useful for eliminating confusion in configurations where characters from different Unicode blocks look identical but have different code points? For example, several Cyrillic letters appear identical to Latin letters (cyrillic symbols: а, е, о, р, с, у, х, А, В, Е, К, М, Н, О, Р, С, Т, У, Х), but they have different Unicode values. By specifying the Unicode code points directly, users can accurately define keybindings for these characters without ambiguity.

Would including this explanation be appropriate, or would you prefer that I omit this as well?

sholderbach: Yeah the character confusables sounds like the best motivation for using this feature to me.

fdncred: Definitely give an example of specifying Unicode code points. I agree that it's a key value prop here.

JustForFun88: Done

sholderbach: Awesome thanks!

PR #1584, opened 2024-10-09
str_replace: add link to rust doc

quadristan:
Trying to escape a capture group name, I had to discover and read rust doc.
Finding it should be easy for people who wants to know the specification of the regex mode.

Adding a link to it will bring knowledge to users.

:+1: Knowledge!
:thinking: Should we copy rust doc here in order to not be coupled with it? I have the feeling Nushell<>rust coupling is intentional, this is why I am opening the PR this way


fdncred:
unfortunately, command docs cannot be updated this way because they will get overwritten at the next release. For help or examples on commands to be updated, one has to change the rust code.

PR #1585, opened 2024-10-11
Standard Library updates for 0.99

NotTheDr01ds:
Updates for the Standard Library changes that will be a part of the 0.99 release.

### DIFF
--- a/book/standard_library.md
+++ b/book/standard_library.md
@@ -71,7 +71,7 @@ This form requires that you prefix the command using the submodule name. This ca
 
 2. In the `use std/help` example, `help` is both the submodule name and the name of the main command it exports. Using `use std/help` allows you to access:
 
-   ```
+   ```nu
    help
    help aliases
    help modules
--- a/book/standard_library.md
+++ b/book/standard_library.md
@@ -212,7 +212,7 @@ Edit those files to use the recommended syntax below.
 :::
 
 ::: note
-If a Nushell library (e.g., from [the `nu_scripts` repository](https://github.com/nushell/nu_scripts)) is using this syntax, please report it via an issue or PR. These will be updated over time after Nushell 0.99.0 is released.
+If a Nushell library (e.g., from [the `nu_scripts` repository](https://github.com/nushell/nu_scripts)), example, or doc is using this syntax, please report it via an issue or PR. These will be updated over time after Nushell 0.99.0 is released.
 
 If a third-party module is using this syntax, please report it to the author/maintainers to update.
 :::
--- a/book/standard_library.md
+++ b/book/standard_library.md
@@ -40,7 +40,7 @@ While working at the commandline, it can be convenient to load the entire standa
 use std *
 ```
 
-However, this form should be avoided in custom commands and scripts since it has the longest load time. While the load time is only on the order of ~20-30 milliseconds on an average system, that can add up in certain scenarios.
+However, this form should be avoided in custom commands and scripts since it has the longest load time.
 
 ::: important Optimal Startup when Using the Standard Library
 See the [notes below](#optimal-startup) on how to ensure that your configuration isn't loading the entire Standard Library.
@@ -48,7 +48,7 @@ See the [notes below](#optimal-startup) on how to ensure that your configuration
 
 ### Importing Submodules
 
-Each submodule of the standard library can be loaded separately. Again, for best performance, load only the submodule(s) that you need in your code.
+Each submodule of the standard library can be loaded separately. Again, _for best performance, load only the submodule(s) that you need in your code._
 
 There are several forms that can be used:
 
@@ -93,9 +93,9 @@ Submodules that are normally imported with `use std/<submodule>`:
 - `use std/log`: The `log <subcommands>` such as `log warning <msg>`
 - `use std/math`: Mathematical constants such as `$math.E`. These can also be imported without a prefix using Form #2 below.
 
-#### 2. Import all submodule commands without a prefix
+#### 2. Import the _contents_ of the module directly
 
-For certain submodules, you will want the commands from a submodule to be available in the current namespace, so that you _can_ simply access the command by name. For instance:
+For certain submodules, you will want the commands from a submodule to be available in the current scope, so that you _can_ simply access the command by name. For instance:
 
 ```nu
 use std/formats *
@@ -117,7 +117,7 @@ As with most modules, you can choose to import only a subset of the commands. Fo
 use std/iter [ zip-with ]
 ```
 
-#### 4. Not Recommended - `use std <submodule>` and `use std <submodule> *`
+#### 4. `use std <submodule>`
 
 While it is _possible_ to import Standard Library submodules using a space-separated form:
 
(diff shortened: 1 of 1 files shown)


book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
 To see a complete list of the commands available in the standard library, run the following:
 
 ```nu
-use std
-scope commands
-| where name =~ '^std '
-| select name usage extra_usage
-| wrap "std-lib"
+nu -c "
```
fdncred, line 21: why is this using `nu -c` instead of just calling the command?

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
+help commands
+```
+
+This form requires that you prefix the command using the submodule name. This can be useful in two scenarios.
+
+1. In the `use std/iter` example, it's important to prefix the commands with the `iter` namespace. If, on the other hand, you were to `use std/iter *`, you would shadow the built-in `find` command with the `iter` module's `find` command.
+
+2. In the `use std/help` example, `help` is both the submodule name and the name of the main command it exports. Using `use std/help` allows you to access:
+
+   ```
```
fdncred, line 74: i can never remember if these code fences should be `nu` or `nushell`
```suggestion
   ```nu
```

fdncred (commented):
I generally approve of this PR but just had a couple comments. Thanks!

book/standard_library.md
NotTheDr01ds: Either works, but the usage through the docs is inconsistent.  I started using `nushell` for my first few PRs, but found that most existing blocks were `nu`, so I switched over to that.

```nu
ag '```nu$' | lines | length
# => 2798
ag '```nushell$' | lines | length
# => 270
```

book/standard_library.md
NotTheDr01ds: That's explained in the note immediately under it ;-)

> It is also run in a separate Nu subshell simply so that it is not loaded into scope in the shell you are using.

If they load it into the current shell's scope, then some of the examples (e.g., on loading with `use std/log` vs. `use std/formats *`) would be more difficult to follow.  It's not a big deal - Just keeps things a bit "cleaner" rather than messing with the scope unnecessarily.

Personally, I'd rather use a `do {}` here, but `scope` doesn't work inside a closure - It only handles the global scope.  That might be a bug (or at least an enhancement request).

book/standard_library.md
NotTheDr01ds: Pushed the fix.

book/standard_library.md
fdncred: Ah, I missed that. Thanks for explaining it. Maybe I would've paid more attention to it if it were before the usage? Not sure it matters that much though.

NotTheDr01ds:
Thanks!  Fix pushed for the missing language-specified, and also responded on the other one.

book/standard_library.md
kubouch: Ideally, this should be `do { use std; scope commands ... }`, but because of a bug in our scoping logic, `scope xxx` does not see commands inside the `do { ... }` block and the result is empty. That's something the new parser should be able to fix.

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
+
+## Importing the Standard Library
+
+While working at the commandline, it can be convenient to load the entire standard library using:
+
+```nu
+use std *
+```
+
+However, this form should be avoided in custom commands and scripts since it has the longest load time. While the load time is only on the order of ~20-30 milliseconds on an average system, that can add up in certain scenarios.
```
kubouch, line 43: What is an average system? Also, the time might change if we modify add more stuff to the stdlib.

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
+
+::: important Optimal Startup when Using the Standard Library
+See the [notes below](#optimal-startup) on how to ensure that your configuration isn't loading the entire Standard Library.
+:::
+
+### Importing Submodules
+
+Each submodule of the standard library can be loaded separately. Again, for best performance, load only the submodule(s) that you need in your code.
```
kubouch, line 51: The second sentence is one of the main takeaways of this document. It should be highlighted as a rule of thumb.

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
+
+There are several forms that can be used:
```
kubouch, line 53: Some of the following points go too much into how modules work, that's not necessary. We can link to https://www.nushell.sh/book/modules.html where it's explained already (although it also needs improvements).

Here you can just say that `std` is a “virtual” directory and the `iter`, `help` are submodules in it, defined as module directories. Linking to https://github.com/nushell/nushell/tree/main/crates/nu-std/std would be helpful to show the module directory structure.

We don't want to teach users to invoke magic incantations. `std` is a normal module like any other module. We shouldn't give an impression that there is anything special about it.

I think you can formulate the points as something like (very abbreviated version):

> The `std` submodules were designed to be used in different ways
> 1. Import the submodule as subcommands
>   Some modules follow a subcommand structure and meant to be imported as `use std/<submodule>`. 
>   a) E.g., `use std/iter` ... defines `iter xxx`subcommands
>   b) Same, but designed to shadow existing commands, e.g., `use std/help`.
>   ... list all modules in this style
> 2. Import the contents of the module directly
>   These modules were designed to make their content available in the current scope with `use std/<submodule> *`.
>   ... list all modules in this style
> 3. Import specific definitions -- IMO this point is not necessary
> 4. Good
>
> When in doubt, checked the source code of the module (linked above).

The terms “prefix” or “namespace” are not very well-defined in Nushell and can be confusing. You can use “scope”, or talk about defining subcommands.

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
+#### 4. Not Recommended - `use std <submodule>` and `use std <submodule> *`
+
+While it is _possible_ to import Standard Library submodules using a space-separated form:
+
+```nu
+use std log
+use std formats *
+```
+
+This is NOT RECOMMENDED. Using this form has the same performance impact as:
```
kubouch, line 129: I would just mention that “similar to `use std`, calling `use std <submodule>` first loads the ENTIRE `std`, then imports only the `<submodule>` from it. In contrast, `use std/<submodule>` loads ONLY the `<submodule>`”.

book/standard_library.md
```diff
@@ -14,32 +18,218 @@ The standard library currently includes:
...
 ```
 
-In addition, `stdlib-candidate`, found in the [nu_scripts Repository](https://github.com/nushell/nu_scripts/tree/main/stdlib-candidate/std-rfc), serves as a staging ground for new commands before they are added to the standard library.
+:::
+
+## The Standard Library Candidate Module
+
+(Also known as `std-rfc`)
+
+`stdlib-candidate`, found in the [nu_scripts Repository](https://github.com/nushell/nu_scripts/tree/main/stdlib-candidate/std-rfc), serves as a staging ground for new commands before they are added to the Standard Library.
```
kubouch, line 151: Stdlib is not only about commands. Modules can define other stuff, for example constants in the `math` module.

kubouch:
Left some comments, thanks for the write-up! I think the main thing is shortening / simplifying the importing instructions. Now, it seems like the stdlib somehow follows some special rules, but in fact, it only uses standard module syntax rules.

book/standard_library.md
NotTheDr01ds: Fair enough - removed the second sentence.

book/standard_library.md
NotTheDr01ds: Thanks - Done.

book/standard_library.md
NotTheDr01ds: I totally agree with you on both points.  (1) We should be able to point to the Modules chapter, and yet (2) the Modules chapter needs a lot of attention.  One of the things we should (i.e., I plan to do) is split out Modules into "Using Modules" and "Writing Modules".  At the moment, a reader would have to wade through *all* of the "writing" information just to find out how to import the different forms.

I'd like to keep this section mostly intact *for now*, and as soon as I get a chance I'll work on the Modules rewrite.  I can use `std` submodules for the examples in the "Using Modules" chapter (or subchapter) since:

1. As you mention `std` is just a standard module/submodule design - Nothing special.
2. It's already on the user's system with no need to come up with "fake examples"
3. I can just *move* (not duplicate) some of this to that chapter where it belongs.

Does that sound like an acceptable path?

book/standard_library.md
NotTheDr01ds: Updated - Thanks!

book/standard_library.md
NotTheDr01ds: Updated (and rephrased) to use the word "additions" - Thanks!

NotTheDr01ds:
@kubouch Thanks!  I've addressed most of the comments above (or at least will have as soon as I push).

> You were kind of re-explaining how module imports work.

Unfortunately, I'm not sure "re-explaining" is quite accurate here, since the current Modules chapter really *doesn't* explain it.  In fact, only one example uses `use <module>/<submodule>` form and that's only in context of *creating* another module that imports a module.

As I mentioned in the review reply above, I'd like to leave this section "over-explained" for now, and (as soon as I get a chance) work on moving it over into the Modules chapter.

My thought at this point is that we need to split the Modules chapter out into:

* "Using Modules" (Move much of this information from this write-up in there)
* "Writing Modules" (Which still needs, as you said, a lot of attention)

kubouch:
OK, splitting Modules to a) using, b) writing sounds good. Most people will likely use a) the most, so that should be explained first.

The `use <module>/<submodule>` imports from a directory module that's located in the `<module>` directory. It uses an existing language feature. If you present it as a unique form of importing a module, that makes it confusing. I hope you get what I mean. We can over-specify the “how” instructions for now until the Modules chapter is improved. What I'm suggesting is to expand a bit on the “why”.

But honestly, it's good enough. The release is rolling off, let's merge it when it's out, and it can be improved later.

PR #1590, opened 2024-10-16
Updated Coming from Bash for reading a secret

vyadh:
Just because this seems like a super common use-case and it took me a little while to figure out how to do it as you don't get any hits for "secret" or "credential" when searching the website.


fdncred:
Thanks

PR #1594, opened 2024-10-24
Example showing discovery and running of test commands

vyadh:
I couldn't use `nupm` as it was marked as experimental so I thought I'd see how far I could take the "reflection" capabilities of Nushell for basic unit testing. I imagine we could create a somewhat full-featured test framework using something like this technique. I'm somewhat tempted to try 😄.

I can simplify the example by removing the ignored tests and printing of test names if you prefer, but it seemed useful enough to leave in.

### DIFF
--- a/book/testing.md
+++ b/book/testing.md
@@ -177,47 +177,41 @@ use std assert
 source fib.nu
 
 def main [] {
-  print "Running tests..."
+    print "Running tests..."
 
-  let test_commands = (
-    scope commands
-      # Exclude all in-built commands
-      | where ($it.type == "custom")
-      # Include all commands starting with "test_"
-      | where ($it.name | str starts-with "test_")
-      # Exclude commands marked as "ignored" in the description
-      | where not ($it.description | str starts-with "ignore")
-      # Get the name of the test command
-      | get name
-      # Prefix each test command with a command to print the test name
-      | each { |test| [$"print 'Running test: ($test)'", $test] } | flatten
-      # Concatenate into a command list
-      | str join "; "
-  )
+    let test_commands = (
+        scope commands
+            | where ($it.type == "custom")
+                and ($it.name | str starts-with "test ")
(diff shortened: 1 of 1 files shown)


vyadh:
Minor update to comply with #1601 and test compatibility with nupm, which I now realise works in a similar way. Maybe there is some appetite to update nupm to allow the test runner to also work for non-module tests. The use case being what I often do, which is write various test scripts in Nushell, but the implementation might be something else - such as testing infrastructure or even bash scripts.

book/testing.md
```diff
@@ -165,4 +165,63 @@ for t in [
...
+  print "Running tests..."
+
+  let test_commands = (
+    scope commands
+      # Exclude all in-built commands
+      | where ($it.type == "custom")
+      # Include all commands starting with "test_"
+      | where ($it.name | str starts-with "test_")
+      # Exclude commands marked as "ignored" in the description
+      | where not ($it.description | str starts-with "ignore")
```
fdncred, line 189: Instead of doing `where | where | where` it's typically recommended to use and/or.
```nushell
scope commands | where $it.type == custom and ($it.name | str starts-with test_) and not ($it.description | str starts-with ignore)
```
You version allows for better documentation but I'm not sure it's the most efficient way to write it. I won't block this PR because of this though.

vyadh:
That's a fair point. I've got too used to lazily-evaluated sequences perhaps. I've updated based on that. It still seems pretty readable, particularly after removing the line-by-line comments. Plus also added what I intended to push previously (compatibility with nupm tests).

book/testing.md
vyadh: That's a fair point. I've got too used to lazily-evaluated sequences perhaps. I've updated based on that. It still seems pretty readable, particularly after removing the line-by-line comments. Plus also added what I intended to push previously (compatibility with nupm tests).

fdncred:
Thanks

PR #1597, opened 2024-10-27
Module chapter rewrite

NotTheDr01ds:
As discussed with @kubouch in #1585, this is a pretty major rewrite of the Modules Chapter, along with a few tangential updates to other chapters.

### High-level

* Divides Modules into two parts:
  - Using
  - Creating

  This allows newer users who just need to use first-or-third-party modules to skip all of the "Creating" part (which was the major focus in the previous version of the chapter)

* Removes much of the explanation on how to import modules from the Standard Library chapter, since we can now simply point to "Using Modules".  The Standard Library chapter is now much shorter as a result.

---

#### Modules Overview

Briefly covers what they are as well as what they can contain.

#### Using Modules

* Adds a brief section on "Installing" - It's not much, but we can flesh this out later as the story improves (i.e., Nupm). Mostly for now, it punts it to the module's documentation.
* Relies on the Standard Library for many of its examples, since it's available "out-of-the-box" to all users.  No need to force new users to *write* a module in order to demonstrate *using* one.
* There was a brief section in the previous Modules chapter on the "Import Pattern'. This has been moved to the "Using" chapter, of course.  It uses different examples so that it doesn't rely on those written in "Creating"
* Added more detail to the import pattern, including mention of relative paths search from `$env.NU_LIBS_PATH`
* Adds information on how to access constants imported from modules (since I obviously had a hard time figuring that out myself)
* "Hiding" was also moved to the "Using" chapter - There was an erroneous/outdated statement that `hide` could also hide environment variables (that's now `hide-env`) that I nixed entirely.
* "See also" section mentioning adding modules to the configuration and Overlays

#### Creating Modules

* Rewrote all examples to be a bit closer to (but not completely) "real world" scenarios. I hope it's okay if we bid farewell to the venerable `greetings.nu`?  I believe the new examples are a bit more clear and even (mostly) more concise.
* Removed "inline" as an example of how to create a module.  Replaced with a tip:

  > While it's possible to use the `module` command to create a module directly at the commandline, it's far more useful and common to store the module definitions in a file for reusability.

* The discussion of `main` is important and is moved up closer to the beginning of the chapter.  Thanks @132ikl for adding that in the first place.  As promised in the review of that PR, I've cc'd you on this one. You should see, I believe, every concept you mentioned in this rewrite.
* The previous version of the chapter had this tidbit:

  > Submodules are modules inside modules. They are automatically created when you call use on a directory: Any .nu files inside the directory are implicitly added as submodules of the main module. There are two more ways to add a submodule to a module:

  That just seems wrong or outdated.  I've removed it.  Now the section focuses on `export module` and `export use` to create submodules.

* Removed the redundant note about `module` without `export` ... It was already covered above as well as in one of the examples.

* The previous version of the chapter used *"namespace"* and *"prefix"* frequently.  As @kubouch mentioned, these are not well-defined terms in Nushell.  For the most part, I tended to rewrite/reword sections with these terms.

* Moved the last two examples (VirtualEnv/Conda and "Dumping Files in a Directory") to the Cookbook. They seem better aligned there.

#### Standard Library

* Removed the section on "viewing source" for Standard Library as it was (a) incomplete, (b) probably not altogether useful, and (b) my idea in the first place, so I didn't feel bad about it ;-)


132ikl:
Awesome, I'll take a look when I get the chance

fdncred (approved):
LGTM

book/modules/creating_modules.md
```diff
@@ -0,0 +1,473 @@
...
+
+Submodules are modules that are exported from another module. There are two ways to add a submodule to a module:
+
+1. With `export module`: Exports (a) the submodule and (b) its definitions as members of the submodule
+2. With `export use`: Exports (a) the submodule and (b) its definitions as members of the parent module
+
+To demonstrate the difference, let's add to our `increment` example by:
+
+- Adding a new `range-into-list` module and command
+- Creating a new parent module `my-utils` with `increment` and `range-into-list` as its submodules
```
132ikl, line 189: I think it's a bit cleaner to explain the new parent module and then the new module, instead of the other way around:

```suggestion
To demonstrate the difference, let's create a new `my-utils` module, with our `increment` example as a submodule. Additionally, we'll create a new `range-into-list` command in its own submodule.
```

book/modules/creating_modules.md
```diff
@@ -0,0 +1,473 @@
...
+   ```
+
+4. We should now have a `my-utils` directory with the:
+
+   - `increment.nu` module
+   - `range-into-list.nu` module
+
+The following examples show how to create a module with submodules.
+
+### Example: Submodule with `export module`
```
132ikl, line 226: I don't think these examples makes it clear when to use `export module` and when to use `export use`, since the examples have nearly the same behavior. I don't have an example off the top of my head on how to demonstrate this, perhaps it could be demonstrated with constants?

book/modules/creating_modules.md
```diff
@@ -0,0 +1,473 @@
...
+
+### Module files and commands cannot be named after parent module
+
+A `.nu` file cannot have the same name as its module directory (e.g., `spam/spam.nu`) as this would create an ambiguous condition with the name being defined twice. This is similar to the situation described above where a command cannot have the same name as its parent.
+
+## Windows Path Syntax
+
+::: important
+Nushell on Windows supports both forward-slashes and back-slashes as the path separator. However, to ensure that they work on all platforms, using only the forward-slash `/` in your modules is highly recommended.
+:::
```
132ikl, line 432: Not sure if this should have its own header if its only content is a infobox ("importantbox"?)

book/standard_library.md
```diff
@@ -50,83 +52,49 @@ See the [notes below](#optimal-startup) on how to ensure that your configuration
...
-As with most modules, you can choose to import only a subset of the commands. For instance, the following would import the `zip-with` command without requiring that it be called with`iter zip-with`.
-
-```nu
-use std/iter [ zip-with ]
-```
-
-#### 4. `use std <submodule>`
-
-While it is _possible_ to import Standard Library submodules using a space-separated form:
+It is _possible_ to import Standard Library submodules using a space-separated form:
```
132ikl, line 88: This form isn't mentioned in "Using Modules", so this line implies that this form is only possible for the stdlib, which isn't really true. I think it is probably worth mentioning in "Using Modules", even if we explicitly recommend against it.

132ikl (commented):
Finally got around to taking a look at this. Really good stuff! I definitely understand modules better now 😄

book/modules/using_modules.md
```diff
@@ -0,0 +1,227 @@
...
+### Module Path
+
+The path to the module can be:
+
+- An absolute or relative path to a directory containing a `mod.nu` file:
+
+  ::: details Example
+
+  ```nu
+  use ~/nushell/modules/nupm`
```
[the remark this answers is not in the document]
NotTheDr01ds: Thanks - Fixed in next push.

book/modules/using_modules.md
```diff
@@ -0,0 +1,227 @@
...
+  Alternatively, you can import the definitions themselves into the current scope. For example:
+
+  ```nu
+  use std/formats *
+  ls | to jsonl
+  ```
+
+  Notice how the `to jsonl` command is placed directly in the current scope, rather than being a subcommand of `formats`.
+
+  The [Standard Library Chapter](../standard_library.md) covers the recommended imports for each submodule.
```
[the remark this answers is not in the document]
NotTheDr01ds: Done (for both) - Thanks!

book/standard_library.md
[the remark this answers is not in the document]
NotTheDr01ds: Good catch.  I asked about this not too long ago, and I was asked to leave it in for now.  I at least removed the "Alpha" label from this version ;-).

book/modules/using_modules.md
```diff
@@ -0,0 +1,227 @@
...
+Installing a module is simply a matter of placing its files in a directory. This might be done via `git clone` (or other version control system), a package manager such as `nupm`, or manually. The module's documentation should provide recommendations.
+
+## Importing Modules
+
+Anything after the [`use`](/commands/docs/use.md) keyword forms an **import pattern** which controls how the definitions are imported.
+
+Notice above that `use` has two arguments:
+
+- A path to the module
+- (Optional) The definitions to import
```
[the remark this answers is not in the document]
NotTheDr01ds: There's a bit of "technically ..." caveat here ;-)

Technically ... when you `use std/formats *`, you are importing the `std/formats` *module* using a direct path to it.

When you `use std formats *` (which I'm trying to dissuade people from), only *then* are you using a submodule.

So really, the *submodule* goes in the "definitions" argument.

I know kubouch has some ideas on how this might change going forward, but not sure when or if they'll get implemented.

book/modules/using_modules.md
```diff
@@ -0,0 +1,227 @@
...
+
+As seen above with the `std/math` examples, some modules may export constant definitions. The syntax for accessing a constant varies slightly depending on how it was imported. For example:
+
+```nu
+use std/math
+$math.PI
+# or
+use std/math *
+$PI
+```
```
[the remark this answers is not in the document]
NotTheDr01ds: Cool - I like your wording better.  Also added the $math record example and still left the other.

book/modules/creating_modules.md
NotTheDr01ds: Makes sense - Done.

book/modules/creating_modules.md
NotTheDr01ds: Yeah, not my favorite style either, but I want a header so that it shows up in the ToC and sidebar, but I think it's useful as a callout as well given its importance.

book/standard_library.md
NotTheDr01ds: It's covered in the bullet *"Import one or more commands from a module"*, which I've renamed *"Import one or more definitions from a module"*.  Below that, it calls out that a definition can be *"Other modules (submodules)"*.  I've expanded it with a note on that form specifically and also added a pointer here in the Standard Library chapter back to it.

book/modules/creating_modules.md
```diff
@@ -0,0 +1,473 @@
...
+
+## Module Files
+
+As mentioned briefly in the Overview above, modules can be created either as:
+
+* `<module_name>.nu`: Useful for simple modules
+* `<module_name>/mod.nu`: Useful for organizing larger module projects where submodules can easily map to subdirectories of the main module
+
+The `increment.nu` example above is clearly an example of the former. Let's convert it to the directory form:
```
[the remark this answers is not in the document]
NotTheDr01ds: Makes sense - Thanks!  I also added the terms "file form" and "directory form" directly to the bullets, and converted the bullets to an ordered list so that I can refer to it more explicitly.

book/modules/creating_modules.md
```diff
@@ -0,0 +1,473 @@
...
+
+```nu
+export def main []: int -> int {
+    $in + 1
+}
+
+export def by [amount: int]: int -> int {
+  $in + $amount
+}
+```
```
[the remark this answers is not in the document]
NotTheDr01ds: Good catch

book/modules/creating_modules.md
NotTheDr01ds: Good point - And there's really only one difference that I know of, and it's a doozy of a corner-case.  I've added it to the "Additional Examples" with a reference from this section.

PR #1598, opened 2024-10-27
Fix type signatures on custom commands

NotTheDr01ds:
After #1596, the examples for type signatures on custom commands were removed entirely, even though the text above said they were still included.

This adds type signature examples back in, but correctly this time.


fdncred:
Thanks

PR #1599, opened 2024-10-27
wrapped script in "def main [] {...}"

amkhlv:
The example was not running correctly, without `main []`
It resulted in the following error:

```
 1 │ ╭─▶ #!/usr/bin/env -S nu --stdin
 2 │ │   
 3 │ │   echo $"stdin: ($in)"
 4 │ │     
 5 │ ├─▶ 
   · ╰──── block is missing compiled representation
```


132ikl:
oh my, good catch!

fdncred:
Thanks

Issue #1574, opened 2024-10-02
Add information on magic `index` renaming


NotTheDr01ds:
As @fdncred mentioned in https://github.com/nushell/nushell/pull/13985 - We need to at least document the functionality where a field named `index` is relabeled as `#` in tables.

I grepped the book for `index` and didn't find anything.

Probably in [Working with Tables](https://www.nushell.sh/book/working_with_tables.html).

PR #1600, opened 2024-10-28
Updates to "Working with Tables" - Adds info on "index renaming" and the table command

NotTheDr01ds:
Implements #1574 

Also, while there, I noticed that the chapter on Tables doesn't have any info on the `table` command, which seemed a bit odd - Added some detail regarding it, since it sometimes causes confusion.

### DIFF
--- a/book/working_with_tables.md
+++ b/book/working_with_tables.md
@@ -404,7 +404,7 @@ You've noticed that every table, by default, starts with a column with the headi
    - The default mode
    - Nushell provides a 0-based, consecutive index
    - Always corresponds to the cell-path row-number, where `select 0` will return the first item in the list, and `select <n-1>` returns the nth item
-   - Is a display of an internal representation only. In other words, it is not addressible by column name. For example, `get index` will not work, nor `get #`
+   - Is a display of an internal representation only. In other words, it is not accessible by column name. For example, `get index` will not work, nor `get #`
 
 1. "Index"-Renamed #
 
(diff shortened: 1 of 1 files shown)


book/working_with_tables.md
```diff
@@ -390,3 +394,191 @@ You can also [`reject`](/commands/docs/reject.md) columns in a table by passing
...
+# => ╰────┴──────────┴──────┴─────────┴──────────────╯
+```
+
+The `sort-by modified` now _also_ sorts the `index` along with the rest of the columns.
+
+##### Adding a Row Header
+
+```nu
+let table = [
+[additions | deletions | delta ];
```
fdncred, line 515: I don't think we should publicize `|` working as a delimiter here. I think it's a bug and probably shouldn't be allowed.

NotTheDr01ds: You're right, of course.  I've gotten used to it so quickly for table literals that I've almost forgotten it's not supported.  It's definitely a bug, as the parse code specifically *doesn't* include `'|` as a list separator today.  But I'm going to propose that we just fix the bug by changing the parser to actually allow it.  I already have a fix in and tested, so I'll drop a PR soon and see if it flies.

In the meantime, I agree that we shouldn't document it until and unless it's official, so I'll reformat this.

NotTheDr01ds:
@fdncred Thanks! Updated.

fdncred:
Thanks

PR #1605, opened 2024-11-01
Fixup and update cmd.exe shelling out docs

Kissaki:
The documentation about shelling out cmd.exe commands was outdated, incomplete, and confusing.

* It mixed past tense "used to" and the forwarded commands without being clear about their relationship and differences
* There have been multiple undocumented changes across versions
* Neither Nu release notes nor this or the book escaping [the shell] mention these version differences

The docs receive a significant update to structure and information:

* Introduce a headlined section for the forwarded commands
* Introduce version-blocks to describe differences between versions
* Describe forwarded command list changes across versions within a same-behavior block
* Differentiates between CMD.EXE and `cmd.exe` (the shell and the executable/forward call)

These version docs may not be useful in the long run, given the version release ages. Nevertheless, these docs will serve as a good, thorough reference even if the information about older versions will be removed at some point or even soon.

The docs themselves provide enough context and source links to verify their correctness. I will not repeat those here in the commit description.

### DIFF
--- a/book/coming_from_cmd.md
+++ b/book/coming_from_cmd.md
@@ -1,6 +1,6 @@
 # Coming from CMD.EXE
 
-This table was last updated for Nu 0.67.0.
+This table was last updated for Nu 0.99.1.
 
 | CMD.EXE                              | Nu                                                                                  | Task                                                                  |
 | ------------------------------------ | ----------------------------------------------------------------------------------- | --------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


book/coming_from_cmd.md
```diff
@@ -1,6 +1,6 @@
 # Coming from CMD.EXE
 
-This table was last updated for Nu 0.67.0.
+This table was last updated for Nu 0.99.1.
```
Kissaki, line 3: I did not go over the command table. I was unsure whether it should remain the previous, be updated to the new, or split across concerns.

I chose to update it for simplicity, thinking issues can still be resolved either way when they come up/get reported.

Kissaki:
CI failure on a file I did not change

```
error: `punctutation` should be `punctuation`
  --> ./book/modules/creating_modules.md:468:7
```

fdncred:
I don't really see the value of mentioning old versions of nushell here.

Kissaki:
Shall I update it only with the most recent info then?

Or is merging this, and then doing a follow-up to reduce it to the recent version viable, which would produce a conclusive doc of the behavior changes in the git log?

fdncred:
ya, i think it would be better to fix it here and only have info on the latest version.

PR #1609, opened 2024-11-01
Fix 'Creating Modules' code examples

Kissaki:
Both errored when executed.

### DIFF
--- a/book/modules/creating_modules.md
+++ b/book/modules/creating_modules.md
@@ -89,7 +89,7 @@ Now it works as expected:
 
 ```nu
 use ./increment.nu
-2024 | inc
+2024 | increment
 # => 2025
 ```
 
@@ -333,7 +333,7 @@ As with any command defined without `--env`, commands and other definitions in t
 ```nu
 export def examine-module [] {
     # Changes the PWD environment variable
-    cd $env.CURRENT_FILE
+    cd ($env.CURRENT_FILE | path dirname)
     ls
 }
 ```


fdncred:
Thanks

NotTheDr01ds:
@Kissaki Thanks!  On the second one, it was actually working "as intended", but what I *think* must have happened is that you (at some point) imported it with:

```nu
use my-utils/mod.nu *
```

This caused `$env.CURRENT_FILE` to return the `mod.nu` rather than the directory name.

Importing with `use my-utils *` as in the doc *should* (at least in my testing, and re-testing, and re-testing yet some more ;-) properly return the directory name and `cd` to it successfully.

With your change, it now fails when properly imported - It doesn't error, but it returns the wrong directory (the *parent* of the module directory itself).

But you've found an interesting issue, but at least one that I can work around with a change to check if the `CURRENT_FILE` ends in `mod.nu` or not.

If you get a chance, can you (a) confirm my suspicions about `use my-utils/mod.nu *`, and (b) that my fix in (TBD) works?  Thanks!

Kissaki:
I retested the code without the fixup in this MR, with various constellations, and I always get the error. `use my-utils`, `use my-utils *`, `use my-utils\mod.nu`, `use my-utils\mod.nu *`, `examine-module`, `my-utils examine-module`. I made sure to use new sessions without previous use module context.

Do you expect `$env.CURRENT_FILE` to not be a file path? Or that `cd`works with a file path?

For me it makes sense that it never works because CURRENT_FILE is a file path, and cd change directory requires a directory path. Navigating into a file makes no sense.

I changed it to ´cd ($env.CURRENT_FILE | path dirname)` with that understanding. Getting the directory path of the file. If it's `mod.nu` then it's the module's directory. If it's a module file, not `mod.nu`, then it would list those (likely unintended behavior). The example is *with* a module folder though, so makes sense there.

With that context, I don't see how #1610 makes sense. If it's not `mod.nu`, then it's not a module *directory* with files in the first place. `cd`-ing into a file path and listing its content makes no sense to me.

NotTheDr01ds:
While I'm still not sure why we are getting different results, I think the core problem goes back to [this Nushell issue](https://github.com/nushell/nushell/issues/13425), but I still want to make sure.  

I thought the fix for that issue had made it in 0.99, but it appears it was reverted, so it's likely I'll need to find an entirely different example for that stage.

> Do you expect `$env.CURRENT_FILE` to not be a file path? Or that `cd`works with a file path?

In my testing, when a module is imported using directory-form, the `CURRENT_FILE` is the name of the directory.  I see that it is the *file* when `use my-utils/mod.nu *` is used, but I haven't (yet) found other cases where it is.  I've also tested on both Linux and Windows with the same results, but I'm still looking.

It's possible there's more than one root issue here.

PR #1610, opened 2024-11-02
Fix Creating Modules code example (a bit further)

NotTheDr01ds:
Fixes the issue (for real, hopefully) in #1609 - See my comment there for details on why this is needed.


fdncred:
Thanks

PR #1612, opened 2024-11-02
Change "local environment" example (yet again!)

NotTheDr01ds:
cc @Kissaki 

Related #1609 and #1610 

Hopefully this change (finally) gets us consistent results on all systems, since it now uses `$nu.default-config-dir`.

@Kissaki Thank you for finding and attempting to fix the issue!  Clearly neither of our previous fixes works for the other (and that's still bugging me, so I may ping you on it again if I have ideas).


fdncred:
hopefully this is the last change with this. thanks for working on it @Kissaki and @NotTheDr01ds.

PR #1614, opened 2024-11-03
Fix prepend to append in coming_from_cmd.md

rodvieirasilva:
# Description
Fix the command prepend to append in the coming from guide.


fdncred:
Thanks

PR #1618, opened 2024-11-05
Fix code-block line numbering

### LINKED ISSUE #1573: Line numbers in code blocks

NotTheDr01ds:
Fixes #1573 on my local system (as well as my test VuePress repo).  Had to look at the VuePress Ecosystem doc repo to see what they were doing.  Turns out we need both PrismJS line-numbering config in addition to Shiki config.

I've set both to start numbering code-blocks when they have 10 or more lines.  This seems to work fairly well.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -172,6 +172,11 @@ export default defineUserConfig({
         sidebar: sidebarZhCN,
       },
     },
+    themePlugins: {
+      prismjs: {
+        lineNumbers: 10,
+      },
+    },
   }),
   plugins: [
     gitPlugin(),
@@ -186,7 +191,7 @@ export default defineUserConfig({
     }),
     shikiPlugin({
       theme: 'dark-plus',
-      lineNumbers: true,
+      lineNumbers: 10,
       langs: [
         'nushell',
         'rust',


fdncred:
I'm glad you figured out what was going on but I'm not sure why we only allow line numbers when there are 10 or more lines? Why not just have them enabled all the time?

NotTheDr01ds:
> I'm not sure why we only allow line numbers when there are 10 or more lines? Why not just have them enabled all the time?

<details><summary>It's pretty ugly on shorter code-blocks, especially when there are a lot of code-blocks in the "vertical scroll" at once.  For example:</summary>
<p>



</p>
</details> 

<details><summary>But this looks much more natural without numbering:</summary>
<p>



</p>
</details> 


I think @hustcer agrees, as he's tried to turn it off in the past unsuccessfully (at least after moving to Shiki).

I set it to 10 because it does come in useful (and looks okay) on longer blocks, but I wouldn't recommend going lower than 8 for that value.  The lower you go, the more chance of having multiple code-blocks with numbers that break up the flow of the reading.

fdncred:
I like it the "ugly" way.

NotTheDr01ds:
> I like it the "ugly" way.

Even on 1 and 2-liners?

fdncred:
yup, but that's me. i'm fine with doing it your way. i just like always showing line numbers. it reminds me of my editor. it always shows line numbers no matter how many lines i have.

NotTheDr01ds:
>  it reminds me of my editor

Ironically, maybe that's part of my problem with the line numbers.  Nushell itself doesn't have line-numbers, and many code-blocks are about typing commands *into the shell* rather than an editor.

It's also kind of odd seeing line-numbers next to the table-borders.

Note that we can still "force" line numbering on for any block we want to using:

````
```nu:line-numbers
# => code block
```
````

And I'll probably do that for some of the shorter ones when we want to point out something on a particular line.  Conversely, we can also turn them *off* for some when they really, really aren't necessary.

> i'm fine with doing it your way.

I did try bumping the value down to 5 to see if it was a good compromise, but I still think 10 is probably the sweet-spot (might could get away with 8).  Like I said, once we turn them off (if you're okay with that), I'll start "overriding" a few of them manually.

PR #1623, opened 2024-11-09
Charles/v0.100.0

CharlesTaylor7:
As you mentioned on [my PR](https://github.com/nushell/nushell/pull/14246#issuecomment-2455243278), I have included notice of the breaking change in the latest release notes. 

Also, I haded a brief documentation example for matching with rest args. Should I split that off into a separate PR to `main`?


IanManske:
Oops just saw this after I already finished writing that section of the notes. Thanks anyways!

PR #1630, opened 2024-11-14
Be more specific about std/log use in modules

NotTheDr01ds:
Importing `std/log` into another module requires some special-care due to its environment variables. Previously, this caveat was discussed as a more general use-case, but `std/log` is so common that it should be explicitly documented.


fdncred:
Thanks

PR #1631, opened 2024-11-14
Update cal.md

nponraj:


NotTheDr01ds:
Thanks for the contribution!  But a couple of problems:

1. The command help is auto-generated from the command output itself.  So what you see on the website is what comes from (essentially) `help cal`.  To modify the example/doc, you need to modify it in the [main Nushell repo](https://github.com/nushell/nushell/blob/636bae24665b815a88f6253cc4b076cf1939a407/crates/nu-command/src/generators/cal.rs#L85).
2. However, this change doesn't seem to me to be accurate.  The `cal --week-start` argument is based on the column names in the calendar table.  At present, this is "mo".  It's not ideal, and I've considered changing this myself, but that's the way it *works* currently (AFAIK), so the doc seems to currently be correct.

PR #1635, opened 2024-11-17
Replace Pipelines "Behind the Scenes" with "Result Display Rendering"

Kissaki:
The previous "Behind the Scenes" docs has some problems:

* The title is not descriptive
* Confusing context reference: The "You may have wondered" references `ls` output despite no applicable example in front of it - and is not stable against other docs section changes either way
* Confusing false equivalence: Claims "one and the same" and then explains how they are not in fact the same

The section is being replaced with "Result Display Rendering", which

* Mentions and explains the existence of a display output hook
* Begins with establishing context: The section is about interactive shell pipeline result display output

I kept the section small so as not to duplicate content from the Hooks documentation page.

---

Adding examples for current, different, or simplified displaying may be reasonable here, but would duplicate the hooks "Changing how Output is Displayed" and "Filtering or Diverting Command Output" sections.


NotTheDr01ds:
This looks like a good, straightforward simplification to me - Thanks!

132ikl:
As noted in https://github.com/nushell/nushell/pull/14361, `display_output` is actually not what runs `table`, but instead `table` is automatically appended to printed pipelines. I think this should be mentioned in the docs if we maintain this behavior, and shouldn't imply that `display_output` does everything.

book/pipelines.md
```diff
@@ -353,34 +353,10 @@ In this case, sleep takes `nothing` and instead expects an argument.
 So, we can supply the output of the `echo` command as an argument to it:
 `echo 1sec | sleep $in` or `sleep (echo 1sec)`
 
-## Behind the Scenes
+## Result Display Rendering
```
NotTheDr01ds, line 356:
```suggestion
## Rendering Display Results
```

book/pipelines.md
```diff
@@ -353,34 +353,10 @@ In this case, sleep takes `nothing` and instead expects an argument.
...
-
-```nushell
-(ls) == (ls | table)
-# => false
-```
-
-`ls | table` is not even structured data!
-:::
+In interactive mode, when a pipeline ends, the [`display_output` hook configuration](https://www.nushell.sh/book/hooks.html#changing-how-output-is-displayed) defines how the result will be displayed.
+The default configuration uses the [`table` command](/commands/docs/table.md) to render structured data as a visual table.
```
NotTheDr01ds, line 359:
```suggestion
In interactive mode, two additional expressions are added to the end of each pipeline to ensure proper rendering of the result:

1. First, if defined, the [`display_output` hook](https://www.nushell.sh/book/hooks.html#changing-how-output-is-displayed) is added. This allows the user to change the rendering.
2. Nushell then internally adds a `table` command to render the final result.
```

NotTheDr01ds (commented):
@132ikl Does that suggestion look better?

NotTheDr01ds:
You're probably right - When I look at it in light of that, it may be an over-simplification.  Does my suggested edit about look more accurate?

book/pipelines.md
Kissaki: I think I prefer this headline either way; no matter if the hook is the final renderer or not.

Although only one of the other 14 headlines uses a verb form like this.

Kissaki:
@132ikl thank you for raising that concern. Although I tested with null or empty setting the hook, I assumed Nushell fell back to `table` in such cases. Both concepts behave practically the same, but of course it'd be better to document it as it is implemented.

Shall this PR wait for the discussion in https://github.com/nushell/nushell/pull/14361 then, or make adjustments like @NotTheDr01ds suggested so the improvements can land? I don't know how long or not the referenced discussion may take?

Kissaki:
It seemed to me like there was nothing open here but nothing happened anymore. Anyway, I applied the suggested headline change, and added examples including behavior with an empty closure.

PTAL

book/pipelines.md
132ikl: not relevant after display_output PR

PR #1636, opened 2024-11-17
Improve pipelines input output types documentation

Kissaki:
Replace "Notes on Errors when Piping Commands" section content with PowerShell references with neutral and more general and extensive "Command Input and Output Types" content.

* Approach the input output types docs from the Basics context rather than PowerShell behavior and situational errors
* Introduce in a more natural and introductory order:
  1. Discover input output types of commands via `help`
  2. Usable case
  3. Problematic case with unintended behavior due to command default behavior
  4. Unsupported case with error
* Add example pipeline expressions for the supported good case
* Describe alternatives like before
* Keep description of why and in what way commands differ in input output support

With the new approach, mentioning PowerShell felt misplaced. Nushell stands well on its own. The relevant error is still there (remains searchable and discoverable). PowerShell specifics would feel more appropriate in a "Coming from PowerShell" documentation.

### DIFF
--- a/book/pipelines.md
+++ b/book/pipelines.md
@@ -300,9 +300,9 @@ Nu works with data piped between two external commands in the same way as other 
 The Basics section above describes how commands can be combined in pipelines as input, filters, or output.
 How you can use commands depends on what they offer in terms of input/output handling.
 
-You can check what a command supports with `help <command name>`, which shows the relevant *Input/output types*.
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
Thanks! Also looks good to me.  I might do some very-minor wordsmithing later, and I'll cc you on it if/when I do, but this is a good improvement over the original, IMHO.

fdncred:
Thanks

PR #1637, opened 2024-11-17
Add missing cmd to nu command

Kissaki:
Add if defined expression via `scope variables`


fdncred:
Thanks

NotTheDr01ds:
How about?

```nushell
if '$<variable>' in (scope variables).name { <command> }
```

Kissaki:
@NotTheDr01ds I see. Not having to define a closure arguably means simpler syntax.

I wasn't aware `in` could be used like that, with a list-collect+path-get.

I was thinking about potential performance differences as well, but it seems like `(scope variables).name` would or could be streaming data in the same way.

Overall, I don't think there's a clear winner. They're alternative query approaches. Makes me wonder if adding both alternatives would make sense.

fdncred:
I do think this is more nushell idiomatic. This example should probably be updated.
```nushell
'$rn' in (scope variables).name
```

Kissaki:
I created #1643 for it.

PR #1638, opened 2024-11-17
Fix and extend CMD.EXE START docs

Kissaki:
The CMD.EXE `START` command does way more than open the file explorer on a directory path. That is only the case when passing a directory path as a parameter.

* Split START reference into three cases:
  1. Open path in default app
  2. Execute internal command
  3. Execute batch file
* Add documentation about Nushell's own `start` command, which overlaps with CMD.exe START

Although Nushell `start` can be used to execute bat files, doing so does not open a new console window like CMD.EXE START does. Hence, no such mapping is added.


fdncred:
Thanks

PR #1639, opened 2024-11-17
Update and simplify Escaping to the System Windows Note

Kissaki:
The history note (historic behavior, in the past) is very old. It has little value in the current docs.

The "shell built-ins" wording and description may not be obvious to readers.

The Coming from CMD.EXE documents the internal command forwarding as well as other relations between the two shells. Here, in 'Escaping the System', we should focus on the aspect of escaping to the system, only give a general description and note, and reference to the more definitive and exhaustive documentation.

### DIFF
--- a/book/escaping.md
+++ b/book/escaping.md
@@ -16,4 +16,6 @@ Escape to external command:
 
 ## Windows Note
 
-When running an external command on Windows, nushell [used to](https://www.nushell.sh/blog/2022-08-16-nushell-0_67.html#windows-cmd-exe-changes-rgwood) use [Cmd.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd) to run the command, as a number of common commands on Windows are actually shell builtins and not available as separate executables. [Coming from CMD.EXE](coming_from_cmd.md) contains a list of these commands and how to map them to nushell native concepts.
+When running an external command on Windows,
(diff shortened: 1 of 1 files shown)


book/escaping.md
```diff
@@ -16,4 +16,6 @@ Escape to external command:
 
 ## Windows Note
 
-When running an external command on Windows, nushell [used to](https://www.nushell.sh/blog/2022-08-16-nushell-0_67.html#windows-cmd-exe-changes-rgwood) use [Cmd.exe](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd) to run the command, as a number of common commands on Windows are actually shell builtins and not available as separate executables. [Coming from CMD.EXE](coming_from_cmd.md) contains a list of these commands and how to map them to nushell native concepts.
+When running an external command on Windows,
```
fdncred, line 19: Seems like these 3 lines should be a paragraph instead of one concept on each line.

NotTheDr01ds: @fdncred Given that it is Markdown, they should ultimately display as a single paragraph anyway, right?

fdncred: ok, sounds good. thanks for confirming. are you ok with these changes?

NotTheDr01ds: I'm good - I don't have much to say from the Windows side, but I think we both agree that the doc shouldn't reference version numbers.

fdncred:
Thanks

PR #1620, opened 2024-11-06
Module Chapter Edits from Review

NotTheDr01ds:
Addresses #1597 feedback from @132ikl.

PR #1640, opened 2024-11-17
Fix absolute path example item text

Kissaki:
The relative example follows afterwards. This block only contains docs on absolute paths.

Split from #1608 so the fix can be merged without further discussion. The other splittable fixup 9eea190c86573d0354ae169bdb9b82bc436f7bb4 has apparently been resolved through another PR #1620.


NotTheDr01ds:
Also looks good to me.

fdncred:
Thanks

PR #1646, opened 2024-11-22
plugins.md - fix typo in python example code

j1mr10rd4n:
This PR fixes a typo in the python plugin example (occurs twice) that prevents the code from working if a reader tries to run the code as given.


sholderbach:
Thanks good catch!

PR #1647, opened 2024-11-23
Update python plugin example for compatibility with v0.100.0

j1mr10rd4n:
Updates the python plugin example code for compatibility with `v0.100.0`.

NB: the rust example in this document is still at `v0.90.2` - I don't have the rust experience to be able to make equivalent updates - it might be best to create an issue for this.

### DIFF
--- a/contributor-book/plugins.md
+++ b/contributor-book/plugins.md
@@ -1020,8 +1020,8 @@ def signature():
     return {
         "sig": {
             "name": "len",
-            "usage": "calculates the length of its input",
-            "extra_usage": "",
+            "description": "calculates the length of its input",
+            "extra_description": "",
             "search_terms": [],
             "required_positional": [],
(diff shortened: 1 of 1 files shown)


contributor-book/plugins.md
```diff
@@ -1081,15 +1081,18 @@ def send_error(id, error_msg, span):
 
 def handle_call(id, call_info):
     try:
-        input = call_info["input"]["Value"]["String"]
+        input = call_info["input"]["Value"][0]["String"]
```
fdncred, line 1084: This is the only part that gives me pause. Seems odd that we'd have to get the 0th element, but maybe that's right?

fdncred:
Yes, the rust and the under-the-hood part need to be updated too. Issues to update them is fine if you don't have that expertise. Thanks!

PR #1648, opened 2024-11-26
Fix broken links to the default config files

WieeRd:
The links to the `default_config.nu` and `default_env.nu` in the configuration section were broken.
Since a lot of useful examples from `default_*.nu` have been moved to the `sample_*.nu` files, I've updated the link to point at the new `default_files/` directory itself rather than each file like it was before.
Readers will now be redirected to `default_files/README.md` which explains each default/sample configuration.


fdncred:
Thanks

PR #1649, opened 2024-11-26
Update the Fish completer to leverage nushell/nushell#14399

WieeRd:
- nushell/nushell#14398
- nushell/nushell#14399

We no longer need to prepend the header row to make sure about the number of columns.
Updated the fish completer to a more clean & performant version.
Also updated the descriptions below accordingly.


fdncred:
Thanks

Issue #1651, opened 2024-11-28
Outdated 'enter.md' is referenced in 'shells_in_shells.md'


GrabbenD:
PR https://github.com/nushell/nushell.github.io/pull/1587 deleted [commands/docs/enter.md](https://github.com/nushell/nushell.github.io/blob/dd091df803ae86c093e46c2afd16b5b1dd20f1b2/commands/docs/enter.md) but it's still referenced in [book/shells_in_shells.md](https://github.com/nushell/nushell.github.io/blob/main/book/shells_in_shells.md).

What's the new equivalent of this command?

> Nu offers the concept of "shells". As the name implies, they're a way of running multiple shells in one, allowing you to quickly jump between working directories and more.

fdncred:
The docs should be updated. Thanks. We've moved this functionality to the nushell stdlib. In order to use it you must do this `use std/dirs shells-aliases *`

NotTheDr01ds:
Thanks for the reminder on this!

Issue #1632, opened 2024-11-15
Update Shells in Shells for 0.99 / 0.100


catwell:
[Shell in Shells](https://github.com/nushell/nushell.github.io/blob/37c713b23ef420430949f32a145d1a7803ba7298/book/shells_in_shells.md) should be updated now that [`std/dirs` is not loaded on startup](https://www.nushell.sh/blog/2024-11-12-nushell_0_100_0.html#removals-toc).

NotTheDr01ds:
Thanks!  I've updated the chapter.  And apologies for the delay on this one!

PR #1653, opened 2024-11-29
Update/Rename Shells-in-Shells Chapter - 0.100 changes

NotTheDr01ds:
Apologies for the delay on this one - This should fix/close:

* https://github.com/nushell/nushell/issues/14419
* #1632 
* #1651

I've also done some "renaming" here, since "shells-in-shells" isn't really an accurate description of the feature.  Given the `dirs` naming of the module, it seems more like Bash (and other's) "Directory Stack" feature.  I've rewritten it to explain it in that context, with a secondary focus on "shells-in-shells" for those who want to use the aliases.


fdncred:
Thanks for the cleanup!

PR #1658, opened 2024-12-03
`make_docs.nu`: fix flags code to only add `short_flag` when needed

paulie4:
NOTE: I ran the script in a separate commit [ffbe438](https://github.com/nushell/nushell.github.io/pull/1658/commits/ffbe438c9d113a7b15f4fa9ad92b2223eac9433a), so anyone who wants to can look at what it output, but I can undo the commit if this PR should only include the change to `make_docs.nu` and not the changes to the generated files.


hustcer:
Thanks for the fix, It looks like `nu_plugin_formats` and `nu_plugin_polars` haven't been added while generating the docs?

paulie4:
I used [brew install nushell](https://formulae.brew.sh/formula/nushell) to install `Nushell` in my Windows WSL Ubuntu, and the formula code [nushell.rb](https://github.com/Homebrew/homebrew-core/blob/17c5563837c2fce8792b62d9e629a8010a5c1ba4/Formula/n/nushell.rb) in GitHub looks like it first does a `cargo install` of `nushell` then does a `glob("crates/nu_plugin_*")` to `cargo install` each of those, so I don't know why the two plugins you mentioned would be missing.

Although, as I mentioned in my PR description, I really only included the results of my running of `make_docs.nu` so people could see what it produces for the files that I was able to generate. Do you guys not have a standard Docker image you use for ensuring consistent results of running `make_docs.nu`?

To make it clearer, I just pushed another commit that undoes that commit, so this PR will just be the changes to `make_docs.nu`, and people can look at the https://github.com/nushell/nushell.github.io/pull/1658/commits/ffbe438c9d113a7b15f4fa9ad92b2223eac9433a commit if they want to see how the changes generated cleaner flags without the confusing `, -`.

hustcer:
Thanks, BTW: you can try the docker images here: https://github.com/nushell/nushell/pkgs/container/nushell

PR #1656, opened 2024-12-03
Blog and Configuration Chapter Preview for upcoming config changes

NotTheDr01ds:
* A blog entry with details on how to upgrade your configuration to take advantage of the recent changes.  This is being made available now for ...

  1. Users who are currently running nightlies or main
  2. Users who want to plan ahead or ask questions
  3. Early review, so that we can make any updates to it before 0.101 releases

  Rather than put all of this information in the 0.101 release notes directly, I would propose that we just have a link from the 0.101 release notes to this blog entry.

* Also the new configuration chapter rewrite, based on the changes.  Again, this gives an opportunity for early review and dogfooding.  When 0.101 releases, I'll just remove the old `configuration.md`, `git mv` the `configuration_preview.md` to `configuration.md`, and we'll have the new chapter in place on Day 0.

* A few miscellaneous changes, like renaming "escapes" to "Running External System Commands" along with fixing the broken links as a result.

PR #1659, opened 2024-12-03
Added backlink from "Default Shell" to Config chapter

NotTheDr01ds:
At @132ikl's recommendation in the #1656 review, added a link from the "Default Shell" section in the Installation back to the Configuration chapter.  In the upcoming update to the chapter, this will also have a link in the other direction as well, but these can be handled separately.


fdncred:
Thanks

Issue #1527, opened 2024-08-28
Document setting `XDG_DATA_HOME`, `XDG_DATA_DIRS` and their uses


NotTheDr01ds:
From [Nushell #13704](https://github.com/nushell/nushell/issues/13704)

This is related to, but probably separate from #1526.

We need to document how to set `XDG_DATA_HOME`, `XDG_DATA_DIRS`, and what they are used for.  Looking at the source, I think:

* `XDG_DATA_HOME` is only used to set `$nu.data-dir` (`$env.XDG_DATA_HOME/nushell`) on startup (all-platforms)
  * Related: `$nu.data-dir` used to:
    - Set the "final" `vendor autoload path (`$nu.data-dir/vendor/autoload` that overrides all others
    - Set a default completions directory (`$nu.data-dir/completions`) directory in `$env.NU_LIB_DIRS` - Side-note: Is this still useful now that we have autoloads?

* On Unix platforms only - `nushell/vendor/autoload` is appended to each directory listed in `XDG_DATA_DIRS` to set `nu.vendor-autoload-dirs`

PR #1663, opened 2024-12-04
XDG_DATA_DIRS are Unix-only

NotTheDr01ds:
As I was closing #1527, I realized that I hadn't noted that `XDG_DATA_DIRS` are only used for autoloads on Unix/Linux.  This is a minor update to call that out.


sholderbach:
Thanks!

PR #1669, opened 2024-12-07
types at a glance: the block should have a semicolon instead of a comma

stormasm:
I was going through the examples and noticed that this one appears to be syntactically not correct

```rust
if true { print "hello!" }, loop { print "press ctrl-c to exit" }
```

it should be a semi-colon instead of a comma

```rust
if true { print "hello!" }; loop { print "press ctrl-c to exit" }
```

Also when I fire up this example in the repl 
the comma does not complete
whereas the semi-colon does

### DIFF
--- a/book/types_of_data.md
+++ b/book/types_of_data.md
@@ -31,7 +31,7 @@ int
 | [Tables](#tables)                     | `[{x:12, y:15}, {x:8, y:9}]`, `[[x, y]; [12, 15], [8, 9]]`            |
 | [Closures](#closures)                 | `{\|e\| $e + 1 \| into string }`, `{ $in.name.0 \| path exists }`     |
 | [Cell-paths](#cell-paths)             | `$.name.0`                                                            |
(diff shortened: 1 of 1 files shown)


stormasm:
https://github.com/nushell/nushell/issues/740#issuecomment-1678539931

This works !

```rust
print hello;
print bye
```

This does not work !
```rust
print hello,
print bye
```

stormasm:
https://www.nushell.sh/book/types_of_data.html#types-at-a-glance

Here is the reference in the book !

stormasm:
In reference to a block...

Note how the semicolon works like a charm 😄 

```rust
if true {print hello}; if true {print bye}
```

The following is broken

```rust
if true {print hello}, if true {print bye}
```

fdncred:
I don't think the entire list of blocks is meant to be run. It's just examples of two blocks separated by a comma. Just like the rest of the examples in this table with commas like floats, strings, ranges, tables, closures.

stormasm:
Ok cool ! 😄 
Not obvious at first glance but makes sense...
Thanks for taking a look...

stormasm:
I am closing this PR.

Issue #1670, opened 2024-12-09
record-return from custom command completers


NotTheDr01ds:
Another day, another feature I learned about that doesn't appear to have ever been documented ;-).

From https://github.com/nushell/nushell/issues/14547 I learned that command-completers can have a return record with additional completion options.

TODO (for someone, perhaps me) to add this to the documentation.

yerlaser:
I created a PR to document this

ysthakur:
> Happy to take this on in about a week if no one else has done it by then

Load-bearing "about" there, but I have a PR open for this now: #1726

yerlaser:
> > Happy to take this on in about a week if no one else has done it by then
> 
> Load-bearing "about" there, but I have a PR open for this now: #1726

Wow, that's an eye-opener. So many hidden gems. I didn't know about style field and fuzzy matching!

PR #1675, opened 2024-12-11
Document how to disable automatic sorting in custom completions

yerlaser:
Resolves: https://github.com/nushell/nushell.github.io/issues/1670

### DIFF
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -34,6 +34,12 @@ The third line is demonstration of the completion. Type the name of the custom c
 When the completion menu is displayed, the prompt changes to include the `|` character by default. This can be changed using `$env.config.menus.marker`.
 :::
 
+Note that, by default, the list of completions will be presented in alphabetical order, but sometimes it makes more sense to sort the list differently, e.g. commit hashes from `git log` output.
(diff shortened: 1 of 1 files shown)


ysthakur:
There's also `case_sensitive`, `positional`, and `completion_algorithm` options that would be good to mention. I'm not a fan of the fact that `positional` exists at all, but if we're documenting the other options, it should probably be documented too. Also not a fan of the fact that the names for the options don't match the names of the settings in `$env.config.completions` but we can update the docs later if we change those names.

NotTheDr01ds:
> not a fan of the fact that it's called `completion_algorithm` here rather than `algorithm`

Lol - I noticed that as well.  I was originally thinking you had added that in your recent PR and was going to suggest changing it before it went public in 0.101, but that's when I realized that it went back *way* further than your submission :-)

NotTheDr01ds:
@yerlaser Unfortunately this is going to need a more systemic change to the chapter.  Keep in mind that we start off near the top of the chapter talking about how the completer must return a `list`.  I'll take the blame for that, as I didn't even know about the `record` return type when I wrote it.

But the differences between the `list` and `record` need to be mentioned through the chapter, rather than just a section on how to sort alphabetically.

PR #1676, opened 2024-12-13
Removed duplicate 'get' in list.md

Squidroot2:
'get' was listed twice as a command that can be used with lists


hustcer:
Thanks

PR #1678, opened 2024-12-14
fix multiple issues with example table

dtenenba:
-  backtick instead of single quote
- "summy" instead of "sunny"
- trailing comma at end of last row


fdncred:
Thanks

PR #1679, opened 2024-12-14
Update creating_errors.md to use $span variable directly in example

prx0:
Use `$span` variable in `my-command` function definition from code example.


fdncred:
Thanks

PR #1683, opened 2024-12-18
Fixed some examples for Command Output in Nushell

RobbingDaHood:
Before: 
```
/tmp/nushell_experiments> do { $env.config.table.mode = none; ls }
Error:
  × External command calls must be explicit in assignments
   ╭─[entry #37:1:31]
 1 │ do { $env.config.table.mode = none; ls }
   ·                               ──┬─
   ·                                 ╰── add a caret (^) before the command name if you intended to run and capture its output
   ╰────
  help: the parsing of assignments was changed in 0.97.0, and this would have previously been treated as a string. Alternatively, quote the string with single or double quotes to avoid it
        being interpreted as a command name. This restriction may be removed in a future release.
``` 

After: 
```
/tmp/nushell_experiments> do { $env.config.table.mode = "none"; ls | table }
 #         name          type    size        modified
 0   2024-12-19 Report   dir    4.0 KiB   28 minutes ago
```

### DIFF
--- a/book/pipelines.md
+++ b/book/pipelines.md
@@ -470,13 +470,13 @@ In fact, running this in the shell will only display `"What?!"` because that is 
 Knowing when data is displayed is important when using configuration variables that affect the display output of commands such as `table`.
 
 ```nu
-> do { $env.config.table.mode = none; ls }
+> do { $env.config.table.mode = "none"; ls }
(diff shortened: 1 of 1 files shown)


hustcer:
Good catch, Thanks

PR #1686, opened 2024-12-20
internal nushell commands use structured data

SOF3:
The "which" in the original sentence sounds like it is referring to the raw bytes.


fdncred:
Thanks

PR #1687, opened 2024-12-20
Refresh Nu command docs for v0.101

hustcer:
Refresh Nu command docs for v0.101


.github/workflows/command-block.yml
```diff
@@ -4,9 +4,16 @@ on:
     paths:
       - 'commands/docs/**'
 
+# Fix Resource not accessible by integration (updatePullRequest)
+permissions:
+  contents: write
```
sholderbach, line 9: Thanks for digging into why that workflow never commented!

Do we need the `contents` permissions, here? I think those are for modifying the repository itself or comments associated with commits for some reason.

.github/workflows/command-block.yml
```diff
@@ -4,9 +4,16 @@ on:
...
+  pull-requests: write
+
 jobs:
   comment:
     runs-on: ubuntu-latest
+    # If the docs was updated by hustcer, skip the commenting step
+    if: ${{!contains(github.event.pull_request.user.login, 'hustcer') }}
```
sholderbach, line 16: Not security critical here but that would match all user names containing yours as a substring. either a direct match or an array of user names in the first arg of `contains` would make sense.

.github/workflows/command-block.yml
hustcer: I have tried many ways, It works in my repo: https://github.com/hustcer/nushell.github.io/pull/3  however, it does not work here: https://github.com/nushell/nushell.github.io/actions/workflows/command-block.yml
I guess some settings in `Actions permissions` need to be changed and I don't have the permission @fdncred

sholderbach: Resolved by #1704 and following

PR #1688, opened 2024-12-20
Fix outdated commands in `coming_from_bash.md`

0x4D5352:
Found two small errors in this file:

1. `stat (which git).path` errors out without the `...` spread operator
2. `random integer` was replaced with `random int` at some point in the commands but not updated here.


fdncred:
Thanks

PR #1705, opened 2024-12-22
Fix command doc (newline stuff)

sholderbach:
Attempt to trigger the workflow


sholderbach:
@hustcer Switching to `pull_request_target` in #1704 seems to allow me to trigger the comment!

PR #1709, opened 2024-12-23
Update dataframes.md

jesper-olsen:
This is in section "Basic Aggregation".

$df_1 is opened in eager mode - but the sum is still lazy - so it is necessary to collect the result...


fdncred:
Thanks

PR #1711, opened 2024-12-27
Update nushell_map_functional.md - Add Nix equivalents

itzwam:


book/nushell_map_functional.md
```diff
@@ -4,44 +4,45 @@ The idea behind this table is to help you understand how Nu builtins and plugins
...
+| into binary  | Integer/toHexString          |                                 | showHex                  |  `toString (toHex value)`      |
+| count        | count                        | length, size                    | length, size             |  `builtins.length`             |
+| date         | java.time.LocalDate/now      |                                 |                          |  `builtins.time`               |
+| each         | map, mapv, iterate           | map, forEach                    | map, mapM                |  `map`                         |
+| exit         | System/exit                  |                                 |                          |  `throw`                       |
+| first        | first                        | head                            | head                     |  `builtins.head`               |
+| format       | format                       |                                 | Text.Printf.printf       |  `builtins.toString`           |
+| group-by     | group-by                     |                                 | group, groupBy           |  Custom function with `map`    |
```
sholderbach, line 17: This table should help folks quickly find the relevant Nushell command or construct if they have a particular language construct or idiom from other languages in mind or understand the meaning at a glance, so it should probably be left blank if you would have to come up with something custom in the other language. Its docs for Nushell not for Nix.

sholderbach (commented):
I don't have a clue about Nix, is this a useful comparison between those general purpose languages and the `builtins` bits in the Nix DSL?

PR #1713, opened 2024-12-28
Upgrade vuepress plugins and shiki, add ansi lang

hustcer:
Upgrade vuepress plugins and shiki, add ansi lang


NotTheDr01ds:
AFAICT, we don't need `"ansi"` in the list - It's just always there in Shiki.

PR #1714, opened 2024-12-28
Disable PrismJS

NotTheDr01ds:
* Disables the PrismJS highlighter entirely (I hope), setting Shiki as the highlighter for both VuePress and the theme.

* CSV was apparently getting parsed through Prism previously, so this also adds the `csv` language to the Shiki config

* And this changes the Shiki code-block theme from `dark-plus` to `vitesse-dark`.  This does not appear to change any `nu`/`nushell` syntax highlighting, but it does enable proper `ansi` coloring.

This might finally explain why line-numbers were being applied by both Shiki and PrismJS previously.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -103,11 +103,15 @@ export default defineUserConfig({
       handleImportPath: (str) =>
         str.replace(/^@snippets/, path.resolve(__dirname, '../snippets')),
     },
+    highlighter: 'shiki',
   },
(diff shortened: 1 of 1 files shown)


hustcer:
Thanks for the quick fix

fdncred:
>  This does not appear to change any nu/nushell syntax highlighting,

definitely does for me. I don't recall green and pumpkin colored text like this
image

PR #1718, opened 2024-12-28
Apply theme change only for ansi highlights

132ikl:
Use the one-dark-pro theme only for `ansi` code blocks. Additionally, override the background color to be the same as `dark-plus`.

I can switch the theme if others would prefer, but I think the `one-dark-pro` theme looks decent along with the other code blocks, and is very readable. Here's an error using `ansi`, highlighted with `one-dark-pro`:

image


fdncred:
Let's try it out and see if we like it.

PR #1719, opened 2024-12-28
Use `nu-highlight` for code blocks

132ikl:
Work in progress. Adds a Shiki transformer which uses the `highlight.nu` script to highlight nushell code blocks using `nu-highlight`.

Also adds the ability to add code before the true code block, for examples which require extra code to work:
```nushell
def "str mycommand" [] {
  "hello"
}
# BEGIN EXAMPLE
str mycommand
```

Only the `str mycommand` line will show up in the final code block, but it will not be error highlighted since the `str mycommand` definition exists.

The output of `nu-highlight` heavily depends on bold fonts for readability, so we should make sure all of the fonts listed in `--code-font-family` support bold. Notably, on my machine the font fell back to Andale Mono, which did not have a bold face. I was able to get bold `ansi` highlighting working install consolas, which is a higher priority font in `--code-font-family` and does have a bold face.

Need to rebase this on #1718 assuming that gets merged.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -1,4 +1,5 @@
 import path from 'node:path';
+import { execFileSync } from 'node:child_process';
 import { defineUserConfig } from '@vuepress/cli';
 import { gitPlugin } from '@vuepress/plugin-git';
 import { feedPlugin } from '@vuepress/plugin-feed';
@@ -10,6 +11,7 @@ import { copyCodePlugin } from '@vuepress/plugin-copy-code';
 import { docsearchPlugin } from '@vuepress/plugin-docsearch';
 import { backToTopPlugin } from '@vuepress/plugin-back-to-top';
 import { mediumZoomPlugin } from '@vuepress/plugin-medium-zoom';
+import { transformerRenderWhitespace } from '@shikijs/transformers';
 
 import {
   navbarDe,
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -7,6 +7,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.2
 
+      - name: Install Nushell
+        run: cargo install nu --locked --no-default-features
+
```
hustcer, line 12: Install Nu by this way is quite slow, try this one: https://github.com/hustcer/setup-nu

.vuepress/config.js
```diff
@@ -192,8 +194,54 @@ export default defineUserConfig({
...
+            }
+
+            this.options.defaultColor = 'onedarkpro';
+            // this doesn't work at the top-level for some reason
+            this.options.colorReplacements = {
+              // make one-dark-pro background color the same as dark-plus
+              '#282c34': '#1e1e1e',
+              // HACK: change color of comments, since nu-highlight can't highlight them
+              '#abb2bf': '#80858f',
+            };
```
hustcer, line 218: It's better to extract these a few lines to a function like adjustTheme, so we can reuse it later

hustcer (commented):
It's a brilliant idea 👍 , just need a bit tweaks

PR #1721, opened 2024-12-29
Reference the correct command

NiceGuyIT:
Replace the reference "insert into" to match the delete command.


ysthakur:
Hey, thanks for the PR, but you'll have to make the edit in the nushell/nushell repository, where the commands are defined: https://github.com/nushell/nushell/blob/4ff4e3f93de89702219d1559fd8cfa9122dd14b2/crates/nu-command/src/stor/delete.rs#L19

PR #1723, opened 2024-12-29
Use dark-plus in code blocks for light mode also

ysthakur:
Currently, in light mode, there's no code highlighting:
image

This is a quick fix so that light mode also uses the dark-plus theme for code blocks.
image

I think this happened because previously, we just had `theme: 'dark-plus'`, whereas now have a `themes` object where the light theme needs to be specified.


fdncred:
Thanks

PR #1726, opened 2024-12-30
Document custom completion options and styling

### LINKED ISSUE #1670: record-return from custom command completers

ysthakur:
Document the fact that custom completers can return a record with an `options` field, as well as the fact that each individual suggestion can include a style, for changing the color of that suggestion.

Closes #1670 

Thank you to rose and NotTheDr01ds for getting ansi highlighting working. Sadly, I only ended up using it for a single code block, but here's to many more ansi blocks in the future!


fdncred (approved):
LGTM. Thanks!

Issue #1722, opened 2024-12-29
Improve the user flow for editing the (autogenerated) docs


NiceGuyIT:
The Nushell website has links at the bottom of the page to "Edit this page on GitHub". Clicking the link takes the user through the process of updating that page using the GitHub web interface to create a pull request. That user flow bypasses the CONTRIBUTING.MD and it's not until after the PR is submitted the user is notified about not updating autogenerated docs.

Some suggestions to improve this for first time contributors.
1. Remove the "Edit this page on GitHub" links on pages that should not be edited.
2. Change or create the GitHub issue template specifying that certain pages, trees roots or paths should not be modified.
3. Add a note about contributing to the README.md. As it is, the user is expected to know about the CONTRIBUTING.md document.

NiceGuyIT:
The `nushell` repo has a good example for the template:
```text
# Description
<!--
Thank you for improving Nushell. Please, check our [contributing guide](../CONTRIBUTING.md) and talk to the core team before making major changes.

Description of your pull request goes here. **Provide examples and/or screenshots** if your changes affect the user experience.
-->
```

(Of course you probably already knew that.)

PR #1728, opened 2025-01-01
Add PR template

NotTheDr01ds:
Adds a PR template as recommended in #1722.  This probably doesn't save much time, since the user will have already pushed the changes by the time they see this.

I'll look at suppressing the GitHub plugin on the `/commands` route next.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,24 @@
+<!--
+# Important
+
+Nushell command documentation (under https://nushell.sh/commands) is automatically generated from the
+built-in Help system (e.g., `help <command>`). Command doc cannot be updated in this repository, but
+instead should be changed in the command source itself in https://github.com/nushell/nushell
+-->
+
+<!--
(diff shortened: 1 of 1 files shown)


fdncred:
I'm not sure this is helpful. Seems like it would be more helpful to do at that issue suggests and remove the "Edit this page in Github" if possible. We already have CI that says you can't edit commands help.

NotTheDr01ds:
Well, that issue suggests *both* courses of action.  I did just finished pushing the other half of it, which removes the link.

But I don't think it hurts to have a PR template here?

fdncred:
We can land this, but I personally detest all the `<!--` mess on PR templates. It's a hassle on the nushell repo IMO.

NotTheDr01ds:
I'm not tied to it.  Let's wait until we have a better use-case.

NotTheDr01ds:
And yes, contributors really should delete the boilerplate before submitting (I do, at least).

fdncred:
i delete the first part of it in the nushell repo, which is why it's a hassle.

PR #1729, opened 2025-01-01
Remove "Edit this page" and contributors list from Command docs

NotTheDr01ds:
* Completes #1722 by adding frontmatter to Nushell Commands and Categories pages to remove the *"Edit this page in GitHub"* link.

* Also removes the *"Contributors"* list since it is not accurate for these pages.

* And implements a workaround for https://github.com/vuejs/core/issues/12088 by placing table heading rows inside a `<thead>` and table body rows inside a `<tbody>`.  This shouldn't be necessary, but without it VuePress will now generate ugly warnings for each of the Category pages.

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -75,6 +75,8 @@ def make_docs [
 #   Various commands for working with bits.
 # usage: |
 #   Various commands for working with bits.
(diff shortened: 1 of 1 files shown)


fdncred (approved):
Much better, thanks!

PR #1735, opened 2025-01-06
fix typo in configuration.md `($nu.data-dir)/vendor/autoloads`

maxim-uvarov:
It seems like the `s` at the end is a typo.


hustcer:
Thanks

PR #1737, opened 2025-01-06
Update operators.md

jesper-olsen:
logical not - higher precedence than and/xor/or, comparison.

### DIFF
--- a/book/operators.md
+++ b/book/operators.md
@@ -48,6 +48,7 @@ Presented in descending order of precedence, the article details the operations 
 - Multiply (`*`), Divide (`/`), Integer/Floor Division (`//`), and Modulo (`mod`)
 - Add (`+`) and Subtract (`-`)
 - Bit shifting (`bit-shl`, `bit-shr`)
+- Logical not (`not`)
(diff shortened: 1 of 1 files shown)


fdncred:
Either I'm not following what this PR is supposed to do, or it's just not right. If you do `help operators | sort-by precedence -r` the order is correct in our current documentation.

jesper-olsen:
Take the expression "not true and false"

If "not" has lower precedence than "and" then it is equivalent to 

not (true and false) = not false = true

if "not" has higher precedence than and, then it it is equivalent to 

(not true) and false = false and false = false

The shell - like most other languages - implements the latter case. So the documentation
is wrong. "not" has the highest precedence of all the boolean operators.

fdncred:
ok, but then our `help operators` command and code that generates it must be wrong too. I'm not sure if `0` was hard coded for `not` or if it is system generated. I haven't looked.

sholderbach:
Related issue already on https://github.com/nushell/nushell/issues/13675

NotTheDr01ds:
For some reason I thought this has been merged already - I checked last week and must have missed it still being open.

@jesper-olsen The operator precedence has changed slightly in 0.103, and it looks like your update needs an update.  Can you move *"Logical not"* directly above *"Logical and"*?

PR #1740, opened 2025-01-07
fix: zh-CN/book/moving_around

Chen1Plus:
remove an incorrect warning for `cp` command

`cp` always overwrites existing files


hustcer:
Thanks

PR #1741, opened 2025-01-08
Remove deprecated/removed split-by command from Map table

NotTheDr01ds:
Removes the `split-by` command from the doc.  This is the only non-blog/command-help reference I could find in the English docs.

Also removes the reference to 0.94 in that page.

Also checked for `date to-table` and `date to-record` as they have been deprecated/removed - There wasn't any use of those commands that I could find.


fdncred:
Thanks

PR #1744, opened 2025-01-12
Describe difference between `get` and `select`

tbrinkman3:


lang-guide/chapters/filters/select-get.md
```diff
@@ -1,3 +1,33 @@
...
+| get name
+| describe
+# => string
+```
+
+### Select
+
+Select only these columns or rows from the input. Opposite of `reject`.
+
+This differs from `get` in that, rather than accessing the given value in the data structure, it removes all non-selected values from the structure.
```
fdncred, line 23: This statement is a little confusing. I always thought of `get` and `select` like this.
- `get` gets the value out of the structure it is in
- `select` selects the value maintaining the structure it is in

NotTheDr01ds:
There's already [Using `get` and `select`](https://www.nushell.sh/book/navigating_structured_data.html#using-get-and-select) in the Book.  While I realize this part of the Language Guide had a "TODO" on it, I'm not sure we need to repeat this.

lang-guide/chapters/filters/select-get.md
tbrinkman3: That makes sense.  I just grabbed the language from the implementation, `scope commands | where name == select | get extra_description.0`.  I didn't realize it was already documented in the Book, so I will close this PR.

tbrinkman3:
@NotTheDr01ds I will close this if it is already documented in the Book.  Thanks!

fdncred:
We could as least put pointers to where it is documents. The language guide is there for a reason and leaving it blank isn't a great idea in my opinion. There should probably be some minimal information, even if it's copied from where it is document in the book.

PR #1747, opened 2025-01-13
Remove select get

tbrinkman3:
Already documented in the Book.  Removing duplicate location. See: https://github.com/nushell/nushell.github.io/pull/1744

### DIFF
--- a/.vuepress/configs/sidebar/en.ts
+++ b/.vuepress/configs/sidebar/en.ts
@@ -252,7 +252,6 @@ export const sidebarEn: SidebarConfig = {
             '/lang-guide/chapters/filters/each-par-each.md',
             '/lang-guide/chapters/filters/selecting_data.md',
             '/lang-guide/chapters/filters/where-filter.md',
-            '/lang-guide/chapters/filters/select-get.md',
           ],
         },
         '/lang-guide/chapters/custom_commands.md',
--- a/lang-guide/chapters/filters/select-get.md
+++ b/lang-guide/chapters/filters/select-get.md
@@ -1,3 +1,0 @@
-# Understanding the difference between `get` and `select`
-
-TODO


fdncred:
I don't think it's a good idea to remove sections from the language guide because they're documented in the book. The language guide performs a slightly different task than the book and things like this should be minimally documented in the guide.

NotTheDr01ds:
@fdncred To be fair, I added that page and TODO in the Language Guide a while back, then added the `select`/`get` info to the Book but forgot to remove it from the Language Guide.  If there's something specific we need in the language guide regarding this, then I'm okay with leaving it, but it's my "fault" it's in there in the first place, so I'm okay with it coming out if you are.

fdncred:
I think the language guide should every language related item in it. Even if it's just a high-level blurb that points to more in-depth documentation.

NotTheDr01ds:
@fdncred (Delayed response) That's fair, but considering this is *only* a page with a "TODO" on it, I think we should remove it for now and add something in when we have real text, right?  There are literally hundreds of potential TODO pages still to be written in the Language Guide. :-)

fdncred:
@NotTheDr01ds I'm up for whatever, but imo, i'd rather see TODOs because it shows potential contributors what's missing. If there are no TODOs, one would think that it's complete, which it is far from.

NotTheDr01ds:
No worries - I don't have strong feelings on it either; it's just that we don't have many other TODOs today.  Perhaps the Language Guide should be an exception there, since it still needs a *lot* of content.

NotTheDr01ds:
I guess the question is - Should we land some version of #1744 then?  Or what would you see as the "Language Guide" version of this section?

fdncred:
Ya, I'd just have a small-ish blurb like that PR for get/set.

PR #1748, opened 2025-01-14
add pipeline examples

WindSoilder:
It's a copy paste from https://hackmd.io/R-r2khDpRvenRg9GndTAgw#Stdout-File-Redirect-ogt.  Additionally with more examples about custom commands.

I think it's good to add to lang-guide, as a follow up of https://github.com/nushell/nushell/pull/14821

### DIFF
--- a/lang-guide/chapters/pipelines.md
+++ b/lang-guide/chapters/pipelines.md
@@ -99,6 +99,8 @@ Pipeline and redirection behavior can be hard to follow when they are used with 
 
 - (^cmd1 | ^cmd2; ^cmd3 | ^cmd4) | ^cmd5
 
+It runs `(^cmd1 | ^cmd2; ^cmd3 | ^cmd4)` first, then pipes *stdout* to `^cmd5`, where both stdout and stderr are directed to the Terminal.
+
 | Command | Stdout     | Stderr     |
 | ------- | --------   | ---------- |
 | cmd1    | Piped      | Terminal   |
@@ -108,6 +110,8 @@ Pipeline and redirection behavior can be hard to follow when they are used with 
 
 - (^cmd1 | ^cmd2; ^cmd3 | ^cmd4) e>| ^cmd5
 
+It runs `(^cmd1 | ^cmd2; ^cmd3 | ^cmd4)` first, then pipes *stderr* to `^cmd5`, where both stdout and stderr are directed to the Terminal.
+
 | Command | Stdout   | Stderr   |
 | ------- | -------- | -------- |
 | cmd1    | Piped    | Terminal |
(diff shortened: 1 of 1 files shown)


lang-guide/chapters/pipelines.md
```diff
@@ -81,3 +81,149 @@ nu demo.nu o> file.txt o+e>| str upcase
 ```
 
 Also note that `complete` is special, it doesn't work with `e>|`, `o+e>|`.
+
+## Stdio and redirection behavior examples
+
+Pipeline and redirection behavior can be hard to follow when they are used with subexpression, or custom commands.  Here are some examples to show stdio behavior
```
fdncred, line 87:
```suggestion
Pipeline and redirection behavior can be hard to follow when they are used with subexpressions, or custom commands.  Here are some examples that show intended stdio behavior.
```

lang-guide/chapters/pipelines.md
```diff
@@ -81,3 +81,149 @@ nu demo.nu o> file.txt o+e>| str upcase
...
+- (^cmd1 | ^cmd2; ^cmd3 | ^cmd4)
+
+| Command | Stdout     | Stderr     |
+| ------- | --------   | ---------- |
+| cmd1    | Piped      | Terminal   |
+| cmd2    | *Terminal* | Terminal   |
+| cmd3    | Piped      | Terminal   |
+| cmd4    | Terminal   | Terminal   |
+
+- (^cmd1 | ^cmd2; ^cmd3 | ^cmd4) | ^cmd5
```
fdncred, line 100: I wonder if it might be good to show where the output of cmd5 goes as well but I'm not sure you can tell that without understanding what cmd5 does. e.g. does it write to file, stdout, or stderr? 

Maybe if each table had a short description saying what it demonstrates then the user would know that cmd5 writes to stdout (or whatever the example shows) and it wouldn't be needed in the table?

lang-guide/chapters/pipelines.md
```diff
@@ -81,3 +81,149 @@ nu demo.nu o> file.txt o+e>| str upcase
...
+Given the following custom commands
+
+```nushell
+def custom-cmd [] {
+    ^cmd1 | ^cmd2
+    ^cmd3 | ^cmd4
+}
+```
+
+Here are some examples to show stdio behavior.  Actually the stio behavior will be the same to previous session.
```
fdncred, line 164:
```suggestion
The custom command stdio behavior is the same as the previous section.
```

lang-guide/chapters/pipelines.md
```diff
@@ -81,3 +81,149 @@ nu demo.nu o> file.txt o+e>| str upcase
...
+
+You can think the body of `custom-cmd` like `(^cmd1 | ^cmd2; ^cmd3 | ^cmd4).
```
fdncred, line 166:
```suggestion
In the examples below the body of `custom-cmd` is `(^cmd1 | ^cmd2; ^cmd3 | ^cmd4).
```

fdncred (changes requested):
This looks good and helpful. I just tried to tweak the text and grammar a bit. Thanks!

lang-guide/chapters/pipelines.md
WindSoilder: > Maybe if each table had a short description saying what it demonstrates

I'd like to suggest leaving it there, the example command themselves only care about the behavior of cmd(1/2/3/4).  The behavior and output of cmd5 is not important.

fdncred: I think it would be better to have a short description above each table for clarity, but we can start with what we have here if you like.

WindSoilder: Sure.  I updated a short description for these commands

fdncred:
thanks

PR #1750, opened 2025-01-17
Show link for save command

kai687:
The link to the save command wasn't shown correctly,  because the whole markup was wrapped in backticks.


fdncred:
Thanks

PR #1755, opened 2025-01-22
chore: add draft front matter for todo/placeholder pages

vyadh:
As per [discord discussion](https://discord.com/channels/601130461678272522/729071784321876040/1331412437714468884) to help exclude pages that shouldn't show up on search because they have no real content.


132ikl:
thanks so much!

PR #1756, opened 2025-01-23
feat: expanded the testing section to add nutest and made slightly more cohesive

vyadh:
I hope you don't mind me plugging Nutest here. It seemed worth highlighting all the options given it's the most comprehensive solution. I also cleaned up the wording a bit and made the different options flow a bit better in terms of why you might choose each.

### DIFF
--- a/book/testing.md
+++ b/book/testing.md
@@ -106,16 +106,18 @@ Error:
    ╰────
 ```
 
-## Running the Tests
+## Running Tests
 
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
Sorry for the delay here.  While I think this section of the doc needs work, I'm not sure about having a project in a non-Nushell repo is the right way to go.  There's also the `toolkit.nu` testing that could probably be mentioned here?

vyadh:
Yeah, I don't mind, I just thought it might be helpful. Can toolkit.nu be used outside of nushell?

PR #1759, opened 2025-01-27
Document file completion fallback for custom/external completers

ysthakur:
Document the fact that custom/external completers can return `null` to fall back to file completions (implemented in https://github.com/nushell/nushell/pull/14781).


fdncred:
Thanks

PR #1761, opened 2025-01-30
Draft state TWiN Configuration

NotTheDr01ds:
This is just a WIP which:

* Creates a new top level navbar item for "News"
* Moves "Blog" under "News"
* Creates a new "This Week in Nushell" entry under News
* Adds the supporting `TwinPosts` component and other changes to support the change

At the moment, this is just a PoC.  I'm going to post the TWiN items as Blog posts instead for now.

If we wanted to move this to production, I think the only remaining TODO would be to figure out how to either:

* Consolidate the Blog and TWiN RSS/Atom feeds
* Have separate RSS/Atom feeds for each

### DIFF
--- a/twin/README.md
+++ b/twin/README.md
@@ -1,0 +1,5 @@
+# This Week in Nushell
+
+[RSS](/rss.xml) / [Atom](/atom.xml)
+
+<TwinPosts />
--- a/.vuepress/client.js
+++ b/.vuepress/client.js
@@ -6,9 +6,11 @@
 
 import { defineClientConfig } from 'vuepress/client';
 import BlogPosts from './components/BlogPosts.vue';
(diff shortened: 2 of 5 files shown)


twin/README.md
```diff
@@ -0,0 +1,5 @@
+# This Week in Nushell
+
+[RSS](/rss.xml) / [Atom](/atom.xml)
```
Kissaki, line 3: What's your intention and suggestion regarding feeds? The feed plugin config was not adjusted so would not include twin posts, but this twin index still links to the feeds.

If we split twin from feeds, I would expect separate feeds too.

Kissaki:
#1765 merged the twin posts as blog posts.

Does that mean this draft PR is obsolete?

PR #1768, opened 2025-02-01
Drop invalid links in translated content

Kissaki:
On translated pages, rather than point to non-existing pages resulting in a 404 not found page being shown to the user, do not link anywhere.

---

Before this change, launching vuepress reports

```
warning Broken links found in fr/book/README.md: dataframes.md
warning Broken links found in pt-BR/book/README.md: shells_em_shells.md
```

The live pages with the dead links:

* https://www.nushell.sh/fr/book/
* https://www.nushell.sh/pt-BR/book/


hustcer:
Thanks

PR #1769, opened 2025-02-01
Fix feed, fix feed item sorting

### LINKED ISSUE #1767: Blog feed (RSS and Atom) only lists earliest posts with wrong date
The [RSS feed](https://www.nushell.sh/rss.xml) and [atom feed](https://www.nushell.sh/atom.xml) advertised at the top of the [Blog page](https://www.nushell.sh/blog/) have items with a publication date that matches the last build date, making them practically unusable as feeds.

Right now, the feeds list *only* "This week in Nu #" 1 - 30, which are the *oldest* blog posts rather than listing the newest.

I suspect this is because the blog posts do not have a date defined in the front matter. Opening the source of a few of them, this seems to be the case for all blog post items, not just the "this week in Nu" posts.

Kissaki:
66c88881f3a1205679adb8a45a2e61b97183a809 (#1765) added old blog posts but with a current author and commit date.

Blog items should be dated to their creation and publication rather than when this git source move took place.

Because the feed plugin used git dates, only the 30 oldest items ended up in the feed.

Falling back to front-matter date should resolve the issue.

Blog posts showing up correctly dated on `/blog/` indicates that the page date is set correctly from the file name, even without explicit front matter within the post source file.

Untested; please verify.

Hopefully resolves #1767

An alternative approach would be committing the post files individually with git commits dated back to their publication dates. But the whole git integration seems error-prone to me, so this commit seems like a better solution with less effort.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -250,14 +250,7 @@ export default defineUserConfig({
         );
       },
       sorter: (a, b) => {
-        return compareDate(
-          a.data.git?.createdTime
-            ? new Date(a.data.git?.createdTime)
-            : a.frontmatter.date,
-          b.data.git?.createdTime
-            ? new Date(b.data.git?.createdTime)
-            : b.frontmatter.date,
-        );
+        return compareDate(a.frontmatter.date, b.frontmatter.date);
       },
     }),
     sitemapPlugin({


NotTheDr01ds:
Were you able to test this locally?  The main reason I didn't test was that RSS feeds aren't working in the local dev server for me.

NotTheDr01ds:
Thanks! Unfortunately doesn't seem to work. There's no frontmatter date for most (if not all) of the content.

I think I'm going to need to either add YALM frontmatter to the older content at least.  Or possible change the sorter to get the date from the filename.

Note for testing: In the `feedPlugin` settings, set `devServer: true,` to enable the feed in the devserver.

PR #1780, opened 2025-02-03
Sort feeds based on filename date first

NotTheDr01ds:
Partial fix for #1765.

This at least should get the newer Blog articles back at the top of the RSS/Atom feeds, but there are still problems (and could be more):

* Feed date still shows the GitHub commit date
* Might be issues when the post filename does not have a valid date - I haven't had a chance to test yet, but will soon.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -239,6 +239,7 @@ export default defineUserConfig({
       apiKey: 'dd6a8f770a42efaed5befa429d167232',
     }),
     feedPlugin({
+      devServer: true,
       rss: true,
       json: true,
       atom: true,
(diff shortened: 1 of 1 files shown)


Kissaki:
For reference; issue #1767 "Blog feed (RSS and Atom) only lists earliest posts with wrong date"

Kissaki:
Note: The feed plugin configured filter `frontmatter.feed === true` includes content that has feed set, which may not be in `/blog/` nor begin with a date (as you noted).

Doesn't look like it's set for any content though, so we probably don't currently hit it.

fdncred:
Couldn't we also just write some script that tacks on whatever is missing from the frontmatter so that it sorts it right?

NotTheDr01ds:
@Kissaki and @fdncred Good points.

As for a script to add the frontmatter, yes - I almost did that the first time around, but since everything *looked* good, I skipped it.

Ultimately, I'd like to make sure that dates work by:

* Using frontmatter if it exists
* Falling back to parsing the filename for a leading-date otherwise
* Falling back to the Git date as a last resort

And to use the same date both for sorting and for the feed "publish date".

PR #1784, opened 2025-02-05
Fix inline code format fencing

Kissaki:
The inline code formatted too much as code. "command" is part of the sentence, not the code/command.


sholderbach (approved):
Thanks!

Issue #1788, opened 2025-02-07
nushell 0.102.0 blog post does not actually list plugins released


mgattozzi:
The post mentions that:

> As part of this release, we also publish a set of optional plugins you can install and use with Nu. To install, use cargo install nu_plugin_<plugin name>.

However, there is no mention of the actual plugins released in the post that I could find and looking around the main repository it was unclear if there were any actually released.

ysthakur:
Ah, looks like we forgot to include a link to the plugins page in the latest blog post. The last did have that. Here it is: https://www.nushell.sh/book/plugins.html#core-plugins.

fdncred:
we could also update the 102 blog if someone wants to submit a pr.

NotTheDr01ds:
Yes, I planned to do this.

I actually *fixed* it in the 0.102 release notes while I was working on them, then accidentally undid the change.

I also *thought* I had updated the template back in December to fix this ... 

I'll get both of these.

mgattozzi:
Thanks so much y'all! 😄

PR #1789, opened 2025-02-07
Include link to plugins

ysthakur:
As pointed out in https://github.com/nushell/nushell.github.io/issues/1788, the 0.102 release notes don't link to the list of core plugins. I threw in a link to https://www.nushell.sh/book/plugins.html#core-plugins


fdncred:
Good point. thanks

PR #1793, opened 2025-02-12
Fix Japanese README typo

massongit:
`Nu はでは` in the Japanese README is incorrect Japanese.
Therefore, I amend it to `Nu は`.


hustcer:
Thanks

PR #1800, opened 2025-02-14
remove mention of `build-string` from coloring_and_theming.md

Bahex:
`build-string` command was apparently removed years ago and somehow survived in the docs until now. Brought to attention in https://github.com/nushell/nushell/issues/15116


hustcer:
Thanks

PR #1805, opened 2025-02-14
Fix outdated `config path` code block

Kissaki:
Although outdated, it's better to be *more* correct.

*Correct me if I'm missing something here, but I don't see what else the previous code block was meant to show.*


fdncred:
Most of the non-english languages are very out of date.

PR #1806, opened 2025-02-14
Clean up code blocks; drop prompt, format ouptut

Kissaki:
* Drop prompt user working dirs
* Drop prompt char
* Format output


Kissaki:
Looking at the diff, I see now that the prompts are probably intentional and useful here.

PR #1807, opened 2025-02-14
Consistently use `nu` as code block lang

Kissaki:
CONTRIBUTING.md defines a preference of `nu` over `nushell`, which overwhelmingly matches our code blocks.

These 66 occurrences were the outliers.

### DIFF
--- a/book/configuration.md
+++ b/book/configuration.md
@@ -136,7 +136,7 @@ The Standard Library also includes a helper command. The default `path add` beha
 a directory so that it has higher precedence than the rest of the path. For example, the following can be
 added to your startup config:
 
-```nushell
+```nu
 use std/util "path add"
 path add "~/.local/bin"
 path add ($env.CARGO_HOME | path join "bin")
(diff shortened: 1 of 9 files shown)


fdncred:
I think the difference may be that github only recognizes nushell whereas our documentation recognizes nu. Thanks.

Kissaki:
I wasn't aware that GitHub recognizes `nushell`; interesting.

I was thinking of the difference between nu as the programming language and interactive REPL examples as well, and whether - if both are supported - it would be worth it to distinguish between the two.

Either way, currently, the docs only use `nu`. Would be nice if GitHub were to support that too.

NotTheDr01ds:
@Kissaki thanks for the cleanup/consistency here.  Greatly appreciated!

The GitHub problem is apparently that `nu` is already assigned to an [earlier, unrelated language](https://en.wikipedia.org/wiki/Nu_(programming_language)).

In VuePress, we can configure it to use our syntax, but we can't change that on GitHub, I don't believe?

I spent far too long using `nu` on GitHub and not understanding why it wasn't highlighting properly :-)

PR #1811, opened 2025-02-20
Update ssh_agent.md

ZennoZenith:
Fix parsing keychain environment variable

### DIFF
--- a/cookbook/ssh_agent.md
+++ b/cookbook/ssh_agent.md
@@ -70,7 +70,7 @@ do --env {
 keychain --eval --quiet <your ssh keys, eg. id_ed25519>
     | lines
     | where not ($it | is-empty)
-    | parse "{k}={v}; export {k2};"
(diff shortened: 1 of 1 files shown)


cookbook/ssh_agent.md
```diff
@@ -70,7 +70,7 @@ do --env {
 keychain --eval --quiet <your ssh keys, eg. id_ed25519>
     | lines
     | where not ($it | is-empty)
-    | parse "{k}={v}; export {k2};"
+    | parse "set -e {k}; set -x -U {k} {v};"
```
fdncred, line 73: Um, this makes it dependent on a particular OS I think. We strive to be as cross platform as possible. The setenv above also isn't cross platform friendly.

ZennoZenith: Maybe you could a little note like 'change parse according to platform'

PR #1814, opened 2025-02-22
Replace usages of use std

### LINKED ISSUE #1813: Update `use std log` reference in Book
Mentioned [in Discord today](https://discord.com/channels/601130461678272522/614593951969574961/1342903968111788054) - `use std log` is not recommended - Need to update to `use std/log`.  There are probably also additional areas that we need to clean up.

https://www.nushell.sh/book/stdout_stderr_exit_codes.html#echo-print-and-log-commands

132ikl:
Fix #1813

### DIFF
--- a/book/dataframes.md
+++ b/book/dataframes.md
@@ -94,7 +94,7 @@ $df_0 | polars schema
 To output more statistically correct timings, let's load and use the `std bench` command.
 
 ```nu
-use std bench
+use std/bench
 ```
 
 We are going to group the data by year, and sum the column `geo_count`.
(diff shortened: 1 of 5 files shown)


vojta001:
This is kind of unfortunate. I get it that this is the preferred way to write the code, but at the same time, the new one does not work in Nu v0.100.0 (current stable NixOS). The documentation does not seem to be versioned at the moment, so anybody lagging behind upstream may potentially be unable to use the manual.

In this particular case, can we possible leave there a note about using `use std assert` instead of `use std/assert` etc. in the past and that it might be required on older versions of Nu?

132ikl:
this functionality has been implemented since 0.99, so you shouldn't have any issues on 0.100. could you share more details about what specifically doesn't work? i also think having versioned documentation for the past however many versions would be something good to look into

vojta001:
Oh right, I am sorry. This is what happens if you run `use std/assert` on one machine and `nu --version` on another. It was actually `0.93.0`, so NixOS 24.05 (security fixes only at the moment). The documentation issue is therefore much less serious, but I would still confider leaving a note there

PR #1815, opened 2025-02-22
fix(setup.md): Quote the EDITOR value

choan:
help: the parsing of assignments was changed in 0.97.0, and this would have previously been treated as a string. 

Alternatively, quote the string with single or double
        quotes to avoid it being interpreted as a command name. This restriction may be removed in a future release.

### DIFF
--- a/cookbook/setup.md
+++ b/cookbook/setup.md
@@ -49,7 +49,7 @@ Note: if you've never used `vim` before and you want to leave typing `:q!` will 
 Go to the end of the file and add
 
 ```nu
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Good catch, we forgot that here. Thanks!

fdncred:
Related/Unrelated, in the nushell config you can have parameters. I'm not sure if that works with env vars though.
```nushell
$env.config.buffer_editor = ["emacsclient", "-s", "light", "-t"]
```

PR #1819, opened 2025-02-24
Make it clearer about `each` behavior toward `null`

hongquan:
- Clearly mention what `each` does with the `null` items. I kept search around Nu docs for a thing like `filter_map`, until I tried `each` with `null`. I wish the docs were more obvious from the start.
- Suggest a better method to iterate over keys / values in a record.
- Add links to related commands.

### DIFF
--- a/commands/docs/each.md
+++ b/commands/docs/each.md
@@ -60,7 +60,7 @@ Produce a list of values in the record, converted to string
 
 ```
 
-Produce a list that has "two" for each 2 in the input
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks, but as per the robot, we can't change this stuff here. It has to be changed in the rust command itself. It's also probably not a good idea to have links since this help shows when you do `each --help` or `help each`.

hongquan:
Thank you. Will head to the Rust source instead.

PR #1825, opened 2025-03-03
Update thinking_in_nu.md

gisle:
Make the jq-example have similar behaviour to the Nushell counterpart.


fdncred:
Ya, that makes sense. Thanks.

PR #1826, opened 2025-03-03
Add docs for jobs

cosineblast:
This adds a book entry for background job handling in nushell, with regard to how it is currently implemented.

Quick question: How do I generate the `.md` files for new commands such as `job spawn`?


fdncred:
> Quick question: How do I generate the .md files for new commands such as job spawn?

You don't need to do that. @hustcer generates all the commands shortly after each release.

132ikl:
i'll take a look when i get the chance

book/background_task.md
```diff
@@ -1,34 +1,77 @@
-# Background Tasks with Nu
+# Background Jobs
 
-Currently, Nushell doesn't have built-in background task management feature, but you can make it "support" background task with some tools, here are some examples:
+## Spawning Jobs
 
-1. Using a third-party task management tools, like [pueue](https://github.com/Nukesor/pueue)
-2. Using a terminal multiplexer, like [tmux](https://github.com/tmux/tmux/wiki) or [zellij](https://zellij.dev/)
+Nushell currently presents experimental support for thread-based background jobs.
```
132ikl, line 5: could you move this line to above the "Spawning Jobs" header?

also, nitpick, but not sure about "presents" here, I think just "has" works fine

book/background_task.md
```diff
@@ -1,34 +1,77 @@
...
-5. Restart Nushell.
+open status.txt
+## => i am inevitable
+```
 
-Then you will get some commands to schedule background tasks. (e.g: `task spawn`, `task status`, `task log`)
+::: tip Note
+Unlike many other shells, Nushell jobs are **not** separate processes, and are instead implemented
+as background threads. An important side effect of that, is that all background jobs terminate once the shell process exits.
+:::
```
132ikl, line 29: could you expand this into a proper (but brief) section? things I think we should specifically cover:

- interactive vs. non-interactive job killing behavior on Nushell exit
- explain that this is the rationale behind no `disown` / eventually `job dispatch`

book/background_task.md
```diff
@@ -1,34 +1,77 @@
...
 
-Cons: It spawns a new Nushell interpreter to execute every single task, so it doesn't inherit current scope's variables, custom commands, alias definition.
-It only inherits environment variables whose value can be converted to a string.
-Therefore, if you want to use custom commands or variables, you have to [`use`](/commands/docs/use.md) or [`def`](/commands/docs/def.md) them within the given block.
+## Listing and Killing jobs
 
-## Using Nu With Terminal Multiplexer
+Active jobs can be queried with the [`job list`](/commands/docs/job_list.md) command, which returns a table with the information of the jobs which are currently executing.
+Jobs can also be kiled/interrupted by using the [`job kill`](/commands/docs/job_kill.md) command, which interrupts the job's thread and kills all of the job's child processes:
```
132ikl, line 34:
```suggestion
Jobs can also be killed/interrupted by using the [`job kill`](/commands/docs/job_kill.md) command, which interrupts the job's thread and kills all of the job's child processes:
```

book/background_task.md
```diff
@@ -1,34 +1,77 @@
...
+# => ┣━━━╋━━━━╋━━━━━━━━╋━━━━━━━━━━━━━━━━┫
+# => ┃ 0 ┃  1 ┃ frozen ┃ [list 1 items] ┃
+# => ┗━━━┻━━━━┻━━━━━━━━┻━━━━━━━━━━━━━━━━┛
+
+job unfreeze 1
+# (process is brought back where it stopped)
+```
+
+If no job id is provided to `job unfreeze`, it will unfreeze the job id of the most recently frozen job.
+Therefore, one can use `alias fg = job unfreeze` to achieve a behavior similar to the one of existing unix shells.
```
132ikl, line 77:
````suggestion
## Job suspension

On Unix targets, such as Linux and macOS, Nushell also supports suspending external commands using <kbd>Ctrl</kbd>+<kbd>Z</kbd>. When a running process is suspended, it is turned into a "frozen" background job: 

```nu
long_running_process # this starts running, then Ctrl+Z is pressed
# => Job 1 is frozen

job list
# => ┏━━━┳━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━┓
# => ┃ # ┃ id ┃  type  ┃      pids      ┃
# => ┣━━━╋━━━━╋━━━━━━━━╋━━━━━━━━━━━━━━━━┫
# => ┃ 0 ┃  1 ┃ frozen ┃ [list 1 items] ┃
# => ┗━━━┻━━━━┻━━━━━━━━┻━━━━━━━━━━━━━━━━┛
```


A frozen job can be brought back into foreground with the [`job unfreeze`](/commands/docs/job_unfreeze.md) command:


```nu
job unfreeze
# process is brought back where it stopped
```

::: tip Tip
For those familiar with other Unix shells, you can create an alias to emulate the behavior of the `fg` command:
```nu
alias fg = job unfreeze
```
:::

By default, `job unfreeze` will unfreeze the most recently frozen job. However, you can also specify the id of a specific job to unfreeze:

```nu
vim
# => Job 1 is frozen

long_running_process
# => Job 2 is frozen

job unfreeze 1
# we're back in vim
```
````

I tweaked this section to make it flow better, let me know what you think. I'm not sure if we need to specify SIGTSTP specifically, I thought of moving it to a note/tip but I couldn't think of what that would look like.

File: book/background_task.md
Comment (132ikl) [inline book/background_task.md:1]: could we rename the file to `background_jobs.md` also?

File: book/background_task.md [hunk 5]
```diff
@@ -1,34 +1,77 @@
...
+Jobs can be can be spawned using [`job spawn`](/commands/docs/job_spawn.md), which receives a closure and starts its execution in a background thread, returning
+an unique integer id for such job:
 
-The module borrows the power of [pueue](https://github.com/Nukesor/pueue), it is possible to schedule background tasks to pueue, and manage those tasks (such as viewing logs, killing tasks, or getting the running status of all tasks, creating groups, pausing tasks etc etc)
+```nu
+'i am' | save status.txt
 
-Unlike terminal multiplexer, you don't need to attach to multiple tmux sessions, and get task status easily.
+job spawn { sleep 10sec; ' inevitable' | save --append status.txt }
+## => 1
```
132ikl, line 14:
```suggestion
# => 1
```

book/background_task.md
```diff
@@ -1,34 +1,77 @@
...
 
-Here we provide a [nushell module](https://github.com/nushell/nu_scripts/tree/main/modules/background_task) that makes working with pueue easier.
+open status.txt
+## => i am
```
132ikl, line 17:
```suggestion
# => i am
```

book/background_task.md
```diff
@@ -1,34 +1,77 @@
...
+# wait for 10 seconds
+sleep 10sec
 
-1. Install pueue
-2. run `pueued`, you can refer to [start-the-daemon page](https://github.com/Nukesor/pueue/wiki/Get-started#start-the-daemon) for more information.
-3. Put the [task.nu](https://github.com/nushell/nu_scripts/blob/main/modules/background_task/task.nu) file under `$env.NU_LIB_DIRS`.
-4. Add a line to the `$nu.config-path` file: `use task.nu`
-5. Restart Nushell.
+open status.txt
+## => i am inevitable
```
132ikl, line 23:
```suggestion
# => i am inevitable
```

132ikl (changes requested):
thanks for writing this up! sorry I just got around to this, I completely forgot about this one :sweat_smile: 

i have a few suggestions, let me know what you think

book/background_task.md
cosineblast: that looks good to me! (and I'm not even the reviewer)

cosineblast:
i think this may have broken the link to the background tasks section sidebar

image

any ideas on how to solve this?

sholderbach:
The sidebar config is under `.vuepress/configs/sidebar` which is sadly hidden from ripgrep by default. There the entries are named. (Due to the renaming the existing entry failed)

cosineblast:
>  which is sadly hidden from ripgrep by default
Oh, that explains a lot, thx

PR #1831, opened 2025-03-09
Add `pwd` command

tkr-sh:
- Adds `pwd` in the cook book when coming from bash.
- Refactor the chinese table.


NotTheDr01ds:
Thanks!

PR #1834, opened 2025-03-15
Correct the tip for changing completion prompt marker

yo-goto:
I've updated the documentation to clarify how to change the prompt marker for the completion menu, specifically, modified the instructions to reflect that the `marker` value in the `$env.config.menus` list should be changed where the `name` key is `completion_menu`.


fdncred:
Thanks

PR #1837, opened 2025-03-19
Fix 0.103.0 release notes help command reference

Kissaki:
`find -f` does not exist. It should be `help -f`.

I decided to primarily use `--find` instead of `-f` to make it clear what the command parameter is for.


hustcer:
Thanks

PR #1838, opened 2025-03-19
Improve German landing page

Kissaki:
PR #779 suggested a wording change that lead to some discussion about the wording. IMO the suggestion is neither a fixup nor an improvement.

As per discussion there, this PR suggests a significant improvement to the wording. With an emphasis on data being structured.

Replaces #779

### DIFF
--- a/de/README.md
+++ b/de/README.md
@@ -8,7 +8,7 @@ actionLink: /de/book/
 features:
   - title: Pipelines um jedes Betriebssystem zu steuern
     details: Nu läuft auf Linux, macOS und Windows. Einmal lernen - überall verwenden.
-  - title: Alles sind Daten.
+  - title: Überall strukturierte Daten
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Even though this was my recommendation I'm now unsure if we want "Überall strukturierte Daten" or "Strukturierte Daten Überall". The latter may sound a bit better.

Kissaki:
> Even though this was my recommendation I'm now unsure if we want "Überall strukturierte Daten" or "Strukturierte Daten Überall". The latter may sound a bit better.

I 100% prefer "Überall strukturierte Daten" over "Strukturierte Daten Überall".

It can be read as or in a more active and defining voice and feel. The latter may be interpreted more passively. Nushell is the active part in making it so though, and it is positive introduction. The latter areas more like an assessment of state than intentional feature. Not fully, but I hope you get what I mean, and feel the same way about it.

NotTheDr01ds:
Thanks!  We might should update the English to match!

cptpiepmatz:
> Thanks! We might should update the English to match!

I don't think so, "Everything is data" is pretty nice. But it just doesn't translate well but I think it doesn't have to directly.

PR #1842, opened 2025-03-22
Correct light theme instructions to use module

antonio-caceres:
The instructions for using Nushell with the light theme used defunct variables.

The updated instructions import the values from `config` instead, which works on my machine. 

- I'm new to Nushell; I think my instructions just put the `light-theme` and `dark-theme` variables in scope, but I'm not sure. 
- I have not tested the website since I do not think my changes are just to content and I'm not confident doing the setup.


NotTheDr01ds:
Thanks!

PR #1844, opened 2025-03-23
COOKBOOK/EXTERNAL_COMPLETERS.MD Add filepath escape troubleshooting

Beinsezii:
See discussions on https://github.com/nushell/nushell/issues/10690

Demo tested nu 0.103.0

I gave an example by manually checking if the value is path, but this means for super fat completions like docker or git you'll experience lag from the sheer amount of FS calls.

Alternatively, a person could probably do something like 

```nu
| update value { str replace "\"" "\\\"" | $"($in)" }
```
but that's going to fail in edge cases.

### DIFF
--- a/cookbook/external_completers.md
+++ b/cookbook/external_completers.md
@@ -120,6 +120,20 @@ let carapace_completer = {|spans: list<string>|
 }
 ```
 
+### File path escapes
+External completers with the exception of [carapace](https://github.com/carapace-sh/carapace) use POSIX escape sequences for files, rather than nu-compatible quoting.
+
+While there's no built-in mechanism to process posix escapes, this can be imperfectly worked around by manually un-escaping files.
+
+For example
+```nu
+let fish_completer_escaped = {|spans|
+    do $fish_completer $spans | update value {
+        if ($in | path exists) {$'"($in | str replace "\"" "\\\"" )"'} else {$in}
+    }
+}
+```
+
 ## Putting it all together
 
 This is an example of how an external completer definition might look like:
--- a/cookbook/external_completers.md
+++ b/cookbook/external_completers.md
@@ -23,6 +23,9 @@ let fish_completer = {|spans|
     fish --command $'complete "--do-complete=($spans | str join " ")"'
     | from tsv --flexible --noheaders --no-infer
     | rename value description
+    | update value {
+        if ($in | path exists) {$'"($in | str replace "\"" "\\\"" )"'} else {$in}
+    }
 }
 ```
 
@@ -31,6 +34,7 @@ A couple of things to note on this command:
 - The fish completer will return lines of text, each one holding the `value` and `description` separated by a tab. The `description` can be missing, and in that case there won't be a tab after the `value`. If that happens, `from tsv` will fail, so we add the `--flexible` flag.
 - The output of the fish completer does not contain a header (name of the columns), so we add `--noheaders` to prevent `from tsv` from treating the first row as headers and later give the columns their names using `rename`.
 - `--no-infer` is optional. `from tsv` will infer the data type of the result, so a numeric value like some git hashes will be inferred as a number. `--no-infer` will keep everything as a string. It doesn't make a difference in practice but it will print a more consistent output if the completer is ran on it's own.
+- Since fish only supports POSIX style escapes for file paths (`file\ name.txt`, etc.), file paths completed by fish will not be quoted or escaped properly on external commands. Nushell does not have a way to natively parse POSIX escapes, so we need to do this conversion manually such as by testing if the items are valid paths as shown in the example. This simple approach is imperfect, but it should cover the 99.9%.
 
 ### Zoxide completer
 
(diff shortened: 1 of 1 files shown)


cookbook/external_completers.md
```diff
@@ -120,6 +120,20 @@ let carapace_completer = {|spans: list<string>|
 }
 ```
 
+### File path escapes
```
ysthakur, line 123: Imperfect though this may be, I think it would be good to include your unescaping code right in the `$fish_completer` definition above, rather than in the troubleshooting section, especially if this is a common issue Fish completer users run into.

ysthakur (commented):
Thanks for improving the experience for other users.

cookbook/external_completers.md
Beinsezii: It really applies to all completers that don't directly support nushell syntax, not just fish. Fish is just the common victim because its the most popular

ysthakur: I see. I'd still rather have the Fish completer above include the unescaping logic (also the Zoxide completer), so people don't have to come back here to figure out why their completer isn't escaping stuff correctly.

And in addition to that, you can leave this bit in the troubleshooting section in addition to that, in case people use another completer. I'd rename `fish_completer_escaped` to `completer_escaped` and so on in that case, though.

Beinsezii: ah if zoxide is good and its just fish sure

Beinsezii:
Updated to incorporate the filepath testing directly in the fish completer as discussed. I tried with `git -` on my local sys and saw no perf impact, but I also run entirely on NVME drives.

cookbook/external_completers.md
```diff
@@ -23,6 +23,9 @@ let fish_completer = {|spans|
     fish --command $'complete "--do-complete=($spans | str join " ")"'
     | from tsv --flexible --noheaders --no-infer
     | rename value description
+    | update value {
+        if ($in | path exists) {$'"($in | str replace "\"" "\\\"" )"'} else {$in}
```
ysthakur, line 27: Not a huge deal but using single quotes helps avoid the escapes
```suggestion
        if ($in | path exists) { $'"($in | str replace '"' '\"' )"' } else { $in }
```

cookbook/external_completers.md
```diff
@@ -31,6 +34,7 @@ A couple of things to note on this command:
...
 - The output of the fish completer does not contain a header (name of the columns), so we add `--noheaders` to prevent `from tsv` from treating the first row as headers and later give the columns their names using `rename`.
 - `--no-infer` is optional. `from tsv` will infer the data type of the result, so a numeric value like some git hashes will be inferred as a number. `--no-infer` will keep everything as a string. It doesn't make a difference in practice but it will print a more consistent output if the completer is ran on it's own.
+- Since fish only supports POSIX style escapes for file paths (`file\ name.txt`, etc.), file paths completed by fish will not be quoted or escaped properly on external commands. Nushell does not have a way to natively parse POSIX escapes, so we need to do this conversion manually such as by testing if the items are valid paths as shown in the example. This simple approach is imperfect, but it should cover the 99.9%.
```
ysthakur, line 37: Minor wording change (as far as I know, we don't plan to handle `\ `, so it's more that Nushell doesn't do it than that it hasn't been implemented yet)
```suggestion
- Since fish only supports POSIX style escapes for file paths (`file\ name.txt`, etc.), file paths completed by fish will not be quoted or escaped properly on external commands. Nushell does not parse POSIX escapes, so we need to do this conversion manually such as by testing if the items are valid paths as shown in the example. This simple approach is imperfect, but it should cover 99.9% of use cases.
```

cookbook/external_completers.md
Beinsezii: it uses single quotes for the substitute `$'` so that won't work

ysthakur: Whoops, didn't realize that

cookbook/external_completers.md
Beinsezii: lgtm

cookbook/external_completers.md
Beinsezii: yeah. it would be really nice if 
```
$` () `
```
was a thing

ysthakur:
Thanks

youwen5:
```nu
| update value { str replace "\"" "\\\"" | $"($in)" }
```
This seems to not work properly. If you try to tab complete and fish returns file paths, it'll insert one double quote into the line and `fish --complete` will error out. You have to manually fix the quote. For example, if I'm trying to open a PDF with zathura, and I type
```
zathura (there is a space here)
```
and hit tab, it'll insert a quote
```
zathura "
```
and then `fish --complete` will complain about the unpaired quote. If you manually add the quote

```
zathura ""
```
Then completion works as expected. Or if you already begin typing a file path
```
zathura ~/Documents
```
It will work too. Only when you type a command, a space right after, and then hit tab, it'll insert the single unpaired quote and break.

This snippet that I got from a comment in [this issue](https://github.com/nushell/nushell/issues/10690#issuecomment-2299916524) has worked much better for me so far.

```nu
let fish_completer = {|spans|
    fish --command $'complete --escape "--do-complete=($spans | str join " ")"'
    | $"value(char tab)description(char newline)" + $in
    | from tsv --flexible --no-infer
    | each {|i|
        if '\' in $i.value {
            $i | merge {'value': $"\"($i.value | str replace -a '\' '')\""}
        } else {$i}
    }
}
```

PR #1849, opened 2025-03-26
Add mention of new operators

LemmusLemmus:
Simply adds a mention of the new operators `has`  and `not-has` in the book.

### DIFF
--- a/book/operators.md
+++ b/book/operators.md
@@ -2,37 +2,39 @@
 
 Nushell supports the following operators for common math, logic, and string operations:
 
-| Operator      | Description                                             |
-| ------------- | ------------------------------------------------------- |
-| `+`           | add                                                     |
-| `-`           | subtract                                                |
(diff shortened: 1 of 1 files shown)


fdncred:
While you're in here, would you mind adding `like` and `not-like`? `like` is a synonym for `=~` and `not-like` is a synonym for `!~`.

LemmusLemmus:
Sure!

Does my change look OK or would you prefer listing the synonyms on separate table rows?

<sub>
BTW, `like` and `not-like` do not seem to be listed when running `help operators`.<sub/>

fdncred:
> Does my change look OK

I think it's fine. Thanks for adding it. If someone doesn't like it, they can submit a PR to change it. The key is that it's documented now.

> BTW, `like` and `not-like` do not seem to be listed when running `help operators`.

Yup, that's a bug too. I wasn't quite sure how to add them.

PR #1859, opened 2025-04-02
make positional parameters more searchable

sunnyzanchi:
just a small change to try and make this page more searchable.  :slightly_smiling_face:  i was searching for "positional arguments" and "positional parameters" and this page wasn't in the results, but it's exactly what i was looking for

### DIFF
--- a/book/scripts.md
+++ b/book/scripts.md
@@ -56,7 +56,7 @@ When this script is run, Nushell will first run the `a` command to completion an
 
 ## Parameterizing Scripts
 
-Script files can optionally contain a special "main" command. `main` will be run after any other Nu code, and is primarily used to add parameters to scripts. You can pass arguments to scripts after the script name (`nu <script name> <script args>`).
+Script files can optionally contain a special "main" command. `main` will be run after any other Nu code, and is primarily used to add parameters to scripts, like POSIX positional parameters. You can pass arguments to scripts after the script name (`nu <script name> <script args>`).
 
 For example:
 


book/scripts.md
```diff
@@ -56,7 +56,7 @@ When this script is run, Nushell will first run the `a` command to completion an
 
 ## Parameterizing Scripts
 
-Script files can optionally contain a special "main" command. `main` will be run after any other Nu code, and is primarily used to add parameters to scripts. You can pass arguments to scripts after the script name (`nu <script name> <script args>`).
+Script files can optionally contain a special "main" command. `main` will be run after any other Nu code, and is primarily used to add parameters to scripts, like POSIX positional parameters. You can pass arguments to scripts after the script name (`nu <script name> <script args>`).
```
sholderbach, line 59: Thanks for tweaking this, I think there is nothing particularly POSIX about how we handle arguments so I would say 

`like positional parameters or flags`

NotTheDr01ds, line 59:
```suggestion
Script files can optionally contain a special "main" command. `main` will be run after any other Nu code, and is primarily used to allow positional parameters and flags in scripts. You can pass arguments to scripts after the script name (`nu <script name> <script args>`).
```

NotTheDr01ds (changes requested):
Agreed with @sholderbach and have provided a suggested change that still gets that "searchability" without mentioning POSIX.

sunnyzanchi:
thanks for the replies! and fair point about POSIX. i committed your suggestion @NotTheDr01ds, should be good to go!

NotTheDr01ds (approved):
Thanks!

Issue #1843, opened 2025-03-23
Documentation: A "Coming from Perl" section


incansvl:
### Related problem

The "Coming from <other language>" sections of the Nushell book are a great idea, but there is an obvious "other language" missing- Perl.

As a solution that aims to bridge the gap between shells and programming languages the Pathologically Eclectic Rubbish Lister (PERL) seems to me the natural ancestor/competitor of Nushell. Perl may be long in the tooth but it is still a very widely used tool on Linux / Unix based systems, and has a lot of features designed specifically to make it work well as a "glue" language.

It's not obvious (at least to me as a user new to Nushell) how some of the things Perl can do can be accomplished in Nushell, some specific "getting started" help would be useful.

### Describe the solution you'd like

A documentation section on "Coming from Perl" could help open up Nushell to areas where Perl has been the "swiss army knife" of system automation for years.

As an example, it would be particularly helpful to have idiomatic Nushell alternatives for the Perl "special variables" including-

```
$0    $PROGRAM_NAME        Pathname of the current script
$$    $PID                 The pid of the process running the current script
$^T   $BASETIME            Timestamp when the current process started
$<    $UID                 Real user ID (uid) of this process
$>    $EUID                Effective user ID of this process
$(    $GID                 Real group ID
$)    $EGID                Effective group ID
$!    $ERRNO / $OS_ERROR   Number or text of the error from the last system call
$?    $CHILD_ERROR         Status returned by the last child process
```

NotTheDr01ds:
I think that's a great idea - I've moved this over to our Doc repo where the PR will need to be as well.

I also think that you have the right idea - this would be better as a page of its own rather than adding it to the rather unwieldy [monolithic mapping table](https://www.nushell.sh/book/nushell_map.html).

Our docs are all user contributions, so hopefully someone can chime in who has the expertise in both Perl and Nushell to help out.

If you'd like to start a page and ask in Discord (or possibly [Stack Exchange](https://discord.com/channels/601130461678272522/1342157458231263242)) about specific items, perhaps we can collectively create something.

For the items you asked about ... Well, I was answering them but had a reboot and lost it :-/

I'll try to fill something in in a bit.

incansvl:
I found a couple of answers in the responses (quite possibly yours?) to https://github.com/nushell/nushell/issues/4564

```
Perl / Bash        Nushell
-----------        -------
$$                 $nu.pid
$?                 $env.LAST_EXIT_CODE
[...]

PR #1860, opened 2025-04-02
New chapter page collecting the list of special variables

NotTheDr01ds:
Added a "Special Variables" page to the "Nu Fundamentals" chapter.  How many did I miss?

### DIFF
--- a/book/special_variables.md
+++ b/book/special_variables.md
@@ -103,6 +103,13 @@ the file resides. Note that this value is also available as a constant through:
 path self | path dirname
 ```
 
+### `$env.PROCESS_PATH`
+
+When _executing a script_, this variable represents the name and relative path of the script. Unlike the two variables
+above, it is not present when sourcing a file or importing a module.
+
+Note: Also unlike the two variables above, the exact path (including symlinks) that was used to _invoke_ the file is returned.
+
 ### `$env.NU_LIB_DIRS`
 
 A list of directories which will be searched when using the `source`, `use`, or `overlay use` commands. See also:
--- a/book/special_variables.md
+++ b/book/special_variables.md
@@ -1,0 +1,159 @@
+# Special Variables
+
+Nushell makes available and uses a number of special variables and constants. Many of these are mentioned or documented in other places in this Book, but this page
+should include _all_ variables for reference.
+
+[[toc]]
+
+## `$nu`
+
+The `$nu` constant is a record containing several useful values:
+
+- `default-config-dir`: The directory where the configuration files are stored and read.
+- `config-path`: The path of the main Nushell config file, normally `config.nu` in the config directory.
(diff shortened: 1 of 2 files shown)


book/special_variables.md
```diff
@@ -0,0 +1,159 @@
...
+- `home-path`: The user's home directory which can be accessed using the shorthand `~`.
+- `data-dir`: The data directory for Nushell, which includes the `./vendor/autoload` directories loaded at startup and other user data.
+- `cache-dir`: A directory for non-essential (cached) data.
+- `vendor-autoload-dirs`: A list of directories where third-party applications should install configuration files that will be auto-loaded during startup.
+- `user-autoload-dirs`: A list of directories where the user may create additional configuration files which will be auto-loaded during startup.
+- `temp-path`: A path for temporary files that should be writeable by the user.
+- `pid`: The PID of the currently running Nushell process.
+- `os-info`: Information about the host operating system.
+- `startup-time`: The amount of time (in duration) that it took for Nushell to start and process all configuration files.
+- `is-interactive`: A boolean indicating whether Nushell was started as an interactive shell (`true`) or is running a script or command-string. For example:
```
132ikl, line 27: could you add a comment to the Nushell source code where `$nu` is generating telling people to update this section of the docs if they update `$nu`?

132ikl:
you missed `$env.PROCESS_PATH` 😄 

also, maybe it would be good to make a table of equivalent special variables in bash, perl, etc. for quick reference?

NotTheDr01ds:
> you missed `$env.PROCESS_PATH` 😄

Ah, yes - Thanks! I tested it and thought it had been removed, but I tested via `source` and `use`, but apparently not via script execution where it's valid.

> also, maybe it would be good to make a table of equivalent special variables in bash, perl, etc. for quick reference?

Probably, but [those tables](https://www.nushell.sh/book/nushell_map.html) are spread across several different pages today, and are starting to get (a) unwieldy and (b) outdated themselves.  We also don't even have a Perl comparison today, but someone expressed interest in #1843.  I'm not sure what the best way to handle it is.

book/special_variables.md
NotTheDr01ds: Good point - https://github.com/nushell/nushell/pull/15481

132ikl:
thanks for taking care of this!

kkoang:
Missing `$env.NU_LOG_LEVEL`
ref: [Stdout, Stderr, and Exit Codes | Nushell](https://www.nushell.sh/book/stdout_stderr_exit_codes.html#echo-print-and-log-commands)

Kissaki:
> Missing `$env.NU_LOG_LEVEL` ref: [Stdout, Stderr, and Exit Codes | Nushell](https://www.nushell.sh/book/stdout_stderr_exit_codes.html#echo-print-and-log-commands)

I created #1976

PR #1867, opened 2025-04-08
Special variables comparisons

NotTheDr01ds:
At @132ikl's suggestion in #1860, I've added a few comparisons to other languages. I've done this as a VuePress *Tab* rather thana  table so that we can use language-highlighting/code-blocks.

@132ikl - What do you think about this format?

### DIFF
--- a/book/special_variables.md
+++ b/book/special_variables.md
@@ -47,6 +47,54 @@ The `$nu` constant is a record containing several useful values:
 
 `$env` is a special mutable variable containing the current environment variables. As with any process, the initial environment is inherited from the parent process which started `nu`.
 
+::: details **_Comparison with other languages:_**
+
+::: tabs
+@tab Nu
+
+```nu
+# Set an environment variable
+$env.VARNAME = "Content"
+
+# Access an environment variable
+$env.VARNAME
+# => Content
+
+# View all environment variables
+$env
+```
+
+@tab Bash/POSIX
+
+````bash:no-line-numbers
+# Set an environment variable
+VARNAME="Content"
+export VARNAME
+# or
+export VARNAME="Content"
+
+# Access an environment variable
+echo $VARNAME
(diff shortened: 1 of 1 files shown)


132ikl:
I'm not sure if the reduced ctrl-f-ability is made up for by a little syntax highlighting. The only one where the syntax highlighting does anything significant is the `$env` example. It would be difficult to find something like the equivalent to `$0` since you can't ctrl-f for `$0`, so people might have to actual read through/scroll through in order to find what they're looking for. Not totally sure the best way to handle it is

NotTheDr01ds:
Right - The <kbd>Ctrl</kbd>+<kbd>F</kbd> searchability is a problem, but I think most folks would figure that out from the index at the beginning.

The bigger problem, with either a table or tabs (but more so with a table) is that there's just no "1-for-1 mapping" in these.

For instance, `$0` in Bash is *not* the same as *any* of our constructs.

In Bash:

* `$0` in a `source` file will report the *path to the interpreter*.
* But running as a *script* works similar to our `$env.PROCESS_PATH`:
* And to add insult to injury, there's `BASH_SOURCE` (non-POSIX) which is an array where `BASH_SOURCE[0]` is closest to our `PROCESS_PATH` in *both* `source` and scripts.
 
Of course, we don't use an array to show the `PROCESS_PATH` when there's a parent caller (or callers) like `BASH_SOURCE`, but we probably should.  But that means that neither `?0` nor `BASH_SOURCE` have a direct equivalent in Nushell.

PowerShell is equally incongruous to Nushell in these constructs.

And, for that reason, I think I'm going to give up trying to actually document this in any "succinct" form :-)

NotTheDr01ds:
@fdncred I closed it because what I had in here was just *wrong* in most respects and, regardless, needs a fresh start.

PR #1868, opened 2025-04-08
Update Config chapter for recent Nushell changes (e.g., user autoloads)

NotTheDr01ds:
Updates the Configuration Chapter of the Book for recent updates, mainly user autoload directories.


blindFS:
I have a question about this feature: should those files in autoload files loaded with `nu --config minimal_config.nu`? Or is there any flag to disable them?

Actually I want to keep a minimal config file for the lsp server, that's the only reason keeps me from switching to this fancy setup.

PR #1873, opened 2025-04-11
release notes 0.104.0: datetime handling highlights

LoicRiegel:
Tell me if it's alright or if there is something I need to change.
Especially I wasn't sure if/where I should include "breaking change" block


fdncred:
Thanks

PR #1876, opened 2025-04-13
Add note about issue #15326

132ikl:
A couple people have made issues about this broken example, so adding a note here that it's broken until nushell/nushell#15326 is fixed

Not sure if this is bad form, but seems better than removing the examples for now and hoping someone remembers to re-add them (cc @NotTheDr01ds)

Rel: https://github.com/nushell/nushell/issues/15561, https://github.com/nushell/nushell/issues/15416


NotTheDr01ds:
I agree - Now let's just hope someone remembers to remove the comment once it's fixed :-)

PR #1877, opened 2025-04-14
chore: move 'job' from strings category to experimental

0x4D5352:
seems like the job command doc was copied from something in strings; this moves it over to experimental with the rest of the job commands.

the other change is from my markdown linter.


0x4D5352:
oh crap, wrong repo.

PR #1878, opened 2025-04-14
Update Cookbook direnv instructions for new std helper

NotTheDr01ds:
As mentioned in https://github.com/nushell/nushell/pull/15569 - This updates the `direnv` formula in the Cookbook to utilize the new `std/config env-conversions` helper.

### DIFF
--- a/cookbook/direnv.md
+++ b/cookbook/direnv.md
@@ -11,23 +11,21 @@ Configuring direnv to work with nushell requires nushell version 0.66 or later.
 
 ### Configuring direnv
 
-To make direnv work with nushell the way it does with other shells, we can use the "hooks" functionality:
+To make direnv work with nushell the way it does with other shells, we can use the "hooks" functionality along with:
+
+- [`load-env`](/commands/docs/load-env.md)
+- An `env-conversion` helper for the `PATH` from the [Standard Library](/book/standard_library.md)
 
 ```nu
-$env.config = {
-  hooks: {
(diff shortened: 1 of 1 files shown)


LemmusLemmus:
Nice!

What is to be done with this  tip on the page?
> you can follow the [`nu_scripts` of Nushell](https://github.com/nushell/nu_scripts/blob/main/nu-hooks/nu-hooks/direnv/config.nu) for the always up-to-date version of the hook above

nu-scripts uses a different implementation. In fact, 2 days ago a [separate fix](https://github.com/nushell/nu_scripts/pull/1083) was introduced for the nu-scripts implementation.

Currently it would be confusing for users whether to use the solution in the cookbook or the one in nu_scripts. Maybe it would be smart to keep only a single implementation around, considering that this is not the first time that the implementation in the cookbook and nu-scripts have diverged?

NotTheDr01ds:
Put in https://github.com/nushell/nu_scripts/pull/1084 to remove the nu_scripts version. Also updated verbage.

LemmusLemmus:
Can this also be merged now, since nushell/nu_scripts#1084 has been merged?

ofek:
It would be great if this could be merged. The tip that this removes contains a link to a file that was removed by the aforementioned PR and I was about to open a fix when I noticed that this change was pending.

cc the most recent commiters @cptpiepmatz @fdncred

Issue #1881, opened 2025-04-16
Fish complete snippet in cookbook breaks when completing file paths


youwen5:
```nu
| update value { str replace "\"" "\\\"" | $"($in)" }
```
This seems to not work properly. If you try to tab complete and fish returns file paths, it'll insert one double quote into the line and `fish --complete` will error out. You have to manually fix the quote. For example, if I'm trying to open a PDF with zathura, and I type

```
zathura (there is a space here)
```
and hit tab, it'll insert a quote
```
zathura "
```
and then `fish --complete` will complain about the unpaired quote. If you manually add the quote

```
zathura ""
```
Then completion works as expected. Or if you already begin typing a file path
```
zathura ~/Documents
```
It will work too. Only when you type a command, a space right after, and then hit tab, it'll insert the single unpaired quote and break.

This snippet that I got from a comment in [this issue](https://github.com/nushell/nushell/issues/10690#issuecomment-2299916524) has worked much better for me so far.

```nu
let fish_completer = {|spans|
    fish --command $'complete --escape "--do-complete=($spans | str join " ")"'
    | $"value(char tab)description(char newline)" + $in
    | from tsv --flexible --no-infer
    | each {|i|
        if '\' in $i.value {
            $i | merge {'value': $"\"($i.value | str replace -a '\' '')\""}
        } else {$i}
    }
}
```

_Originally posted by @youwen5 in https://github.com/nushell/nushell.github.io/issues/1844#issuecomment-2810432615_

ysthakur:
@Beinsezii Could you look into this?

Beinsezii:
Ah, that's goofy. It's easy to fix though by just inverting the quotes

`fish --command $"complete '--do-complete=($spans | str join ' ')'"`

that way the manual path checks still work instead of relying on the shady posix unescape

ysthakur:
@Beinsezii Ah, nice. If you have time, could you put up another PR to change that in the docs?

Beinsezii:
Will be resolved with #1883

PR #1883, opened 2025-04-19
Update external_completers.md - Fix fish quotes for ISS1881

Beinsezii:
Invert fish quotes to prevent issues where the output is a single double quote, breaking the string substitution

https://github.com/nushell/nushell.github.io/issues/1881

the TL;DR above is that if you do something like `cat <TAB>` instead of `cat scri<TAB>` the string substitution would break from some funky nushell quoting issues.

cc @ysthakur


fdncred:
Thanks

ysthakur:
Thanks for the quick fix!

ysthakur:
Thanks for the quick fix!

PR #1888, opened 2025-04-26
Document pre_execution use of commandline command

### LINKED ISSUE #1827: Document how to receive commandline as part of `pre_execution` hook
See https://github.com/nushell/nushell/issues/15174

Kissaki:
Resolves #1827


sholderbach (approved):
It could be argued that this is close to some cookbook example but I think each chapter should contain some practical nugget so I like this

fdncred:
Looks good as a primer. Thanks!

PR #1889, opened 2025-04-26
Fix hooks.md basic hooks examples and hooks list

### LINKED ISSUE #1879: In doc Basic Hooks  config error
Image

Image

**Valid syntax**

```
$env.config.hooks = {
    pre_prompt: [{ print "pre prompt hook" }]      # 使用列表包裹闭包
    pre_execution: [{ print "pre exec hook" }]     # 使用列表包裹闭包
    env_change: {
        PWD: [{ |before, after| print $"changing directory from ($before) to ($after)" }]  # 使用列表包裹闭包
    }
}
```

Kissaki:
Fix hooks syntax; they are a list of values. The code as it was did not run.

Nushell has been using an inherent configuration and the user configuration only replacing parts. For the hooks it makes sense to set only the hooks value.

Replace "you can also use a list", which is the default and used above already, with a "you can append additional hooks" example.

For zh-CN, at least add the missing available hook docs in English instead of missing them altogether.

For zh-CN, add the new example replacement with EN text rather than keeping the old outdated one, or missing it altogether.

Resolves #1879

### DIFF
--- a/book/hooks.md
+++ b/book/hooks.md
@@ -28,15 +28,11 @@ The steps to evaluate one line in the REPL mode are as follows:
 To enable hooks, define them in your [config](configuration.md):
 
 ```nu
-$env.config = {
-    # ...other config...
-
-    hooks: {
-        pre_prompt: { print "pre prompt hook" }
-        pre_execution: { print "pre exec hook" }
-        env_change: {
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #1890, opened 2025-04-27
Use new code font generally to fix Android monospace

### LINKED ISSUE #83: Seeing issues with tables not aligning properly in some browsers
Here's a picture from one screenshot:

Eq4Q2IdVEAEqyqM

Asking around on twitter, I was told:

"Ok so I assumed this was ligatures from ---- not the actual characters for those lines. The misalignment on Android comes from the droid sans mono font not having those characters, causing it to fall back to noto sans symbols, which has different widths.

Which means `font-variant-ligatures: none` will have no effect, but using a web font to force it to use one of the other fonts in your list before it falls back to monospace will fix this for you"

So it sounds like we need to set better fallback fonts so that Android (and other browsers with the same issue) don't fallback to the wrong font.

cc @ibraheemdev

Kissaki:
Using a specific font, and hosting it ourselves, allows us to ensure consistent behavior and a code font that covers the Nushell table border drawings characters.

Adds FiraCode font, which is under OFL (SIL Open Font license).

https://github.com/tonsky/FiraCode

Introduces a 100 kb font file download, which should happen only once through adequate cashing.

---

I can't test locally; if someone could do that, or we test it live.

Should resolve #83.


cptpiepmatz:
I'd like it better if we could just use a google font instead of rehosting a widely used font

Kissaki:
Using Google fonts has some implications regarding data sharing.

While the [Google Fonts Privacy FAQ](https://developers.google.com/fonts/faq/privacy) declares no data is used for profiling or targeted advertising. Despite that, there's a [German court ruling](https://openjur.de/u/2384915.html) that sharing an IP address with Google and consequently into the USA constitutes sharing of personal data.

The original benefits of CDNs often do not apply anyway. Different versions of fonts mean different websites will do their own requests. CDN delivery is often not or not significantly faster than delivering it directly. The font used here is not *that* popular on regular websites. And browsers have implemented resource isolation where fonts loaded from cross-origin are isolated to that context.

What do you dislike about hosting it?

cptpiepmatz:
Yeah sure, but please include the license of it:
> 2) Original or Modified Versions of the Font Software may be bundled,
redistributed and/or sold with any software, provided that each copy
contains the above copyright notice and this license.

fdncred:
Sounds good @Kissaki. Thanks for the info. I agree with @cptpiepmatz that if we can get that license in the code, I will land it. Appreciate the support.

fdncred:
Thanks

Issue #1734, opened 2025-01-05
contributor-book/plugins


jesper-olsen:
This page is not up to date - I read it recently for the first time and tried to follow the examples:
https://www.nushell.sh/contributor-book/plugins.html#creating-a-plugin-in-rust

The examples reference v. 0.90.2 and nushell is now at (0.101.1) - I think there have been a lot of changes to the plugin interface, in particular the streaming interface.

For SimplePluginCommand, it is enough to rename fn usage() to fn description() - then it compiles and seems to work ok. For PluginCommand() it is more complicated.

In section "under the hood" there is this example:

```
Putting that together, it looks like this:

$ ./target/release/nu_plugin_len --stdio
json{"Hello":{"protocol":"nu-plugin","version":"0.90.2","features":[]}}
{"Hello":{"protocol":"nu-plugin","version":"0.90.2","features":[]}}
{"Call":[0,"Signature"]}
{"CallResponse":[0, {"Signature":[{"sig":{"name":"len","usage":"calculates the length of its input","extra_usage":"","search_terms":[],"required_positional":[],"optional_positional":[],"rest_positional":null,"vectorizes_over_list":false,"named":[{"long":"help","short":"h","arg":null,"required":false,"desc":"Display the help message for this command","var_id":null,"default_value":null}],"input_type":"String","output_type":"Int","input_output_types":[],"allow_variants_without_examples":false,"is_filter":false,"creates_scope":false,"allows_unknown_args":false,"category":"Default"},"examples":[]}]}]}
```

I does work - with the SimplePluginCommand implementation - but I think it should be re-written using piped input. Input and output in this version are interleaved and it is not possible - at least in my terminal - to interactively input the example; Plugin waits for enter and input terminates after enter.

There is also this example in the section:
```
$ echo '{"Hello":{"protocol":"nu-plugin","version":"0.90.2","features":[]}}{"Call":[0,{"Run":{"name":"len","call":{"head":{"start":100953,"end":100957},"positional":[],"named":[]},"input":{"Value":{"String":{"val":"hello","span":{"start":100953,"end":100957}}}}}}]}' | target/release/nu_plugin_len --stdio
json{"Hello":{"protocol":"nu-plugin","version":"0.90.2","features":[]}}
{"PipelineData":{"Value":{"Int":{"val":5,"span":{"start":100953,"end":100957}}}}}
```

It doesn't work as listed any more and in any case, it would be good to have an explanation of where 100953 comes from.
[...]

PR #1895, opened 2025-04-27
Update Plugin example version numbers

Kissaki:
Relates #1734


fdncred:
Does the example still compile after changing version numbers?

Kissaki:
I went through the SimplePlugin example, yes.

As mentioned in #1734 the follow-up block with the streaming implementation does not - but I'm not sure if that has anything to do with the version increase. I chose to see that as a separate issue.

fdncred:
Thanks

PR #1897, opened 2025-04-28
Don't mention positional on custom completer page

ysthakur:
Since https://github.com/nushell/nushell/pull/15511 deprecates `positional`, this PR removes mentions of that option from the page on custom completions. Also adds that `completion_algorithm` can be set to `substring`.


ysthakur:
I'll just merge this since it's not a huge change.

PR #1898, opened 2025-04-28
Fix release notes `$nu.vendor-autoload-dirs`

Kissaki:
0.95.0 mentions `$nu.vendor-autoload-dir` future-facing (will be introduced in the future).
But it was introduced as `$nu.vendor-autoload-dirs` (plural) in 0.96.0.

0.96.0 mentions PRs #13217 and #13382.
The first introduced it in singular, but the second changed it to plural. The plural name `$nu.vendor-autoload-dirs` is still the correct name in the current version 0.103.0.


fdncred:
Thanks

PR #1906, opened 2025-04-30
Fix vuepress config

Kissaki:
Regression introduced in 2ec6143edbf6b9412a2e8a951fb90a91d355bb1a

Missing comma is missing on the first, newly introduced array element, and somehow that does not lead to build errors, but probably to runtime issues.


fdncred:
Thanks

Issue #1066, opened 2023-09-19
Rust example in plugins chapter of contributers book does not compile in recent versions of Nushell


edhowland:
The Rust example plugin in [https://www.nushell.sh/contributor-book/plugins.html](https://www.nushell.sh/contributor-book/plugins.html)
does not compile with later versions of Nushell. Following the instructions
on the page the Cargo.toml pins the crates to 0.84.0 or whatever your current
version of Nushell is.  The sample Cargo.toml on the page pins the crates at
0.80.0. If you replace the Cargo.toml with the example, the project compiles.
The change occurred between 0.80.0 and 0.81.0 in the nu-protocol crate.

The lines in the file: './src/main.rs'  from the plugins.md source for the above
web page are:

./src/main.rs:16,17

```rust
            .input_type(Type::String)
            .output_type(Type::Int)
```

The change occurred in this commit:

(I've elided some of the non-relevant details of the commit message)

```
commit df94052180f45685b2a04ba332da734aa8aa701f
Author: Dan Davison <dandavison7@gmail.com>
Date:   Wed Nov 9 16:55:05 2022 -0500

    Declare input and output types of commands (#6796)
    
...
    * Make it possible to declare input and output types for commands
    
    - Enforce them in tests
    
    * Declare input and output types of commands
    

    * Revert SyntaxShape::Table -> Type::Table change

    Co-authored-by: JT <547158+jntrnr@users.noreply.github.com>
```

The source lines that changed are in the main nushell repository:

'./crates/nu-protocol/src/signature.rs' (around lines 443 or so)

Changing the lines of the example main.rs to:

```rust
            .input_output_type(Type::String, Type::Int)
```

Allows the project to build.

Additionally,  the sample Cargo.toml should be changed to pin both the 
'nu-plugin' and 'nu-protocol' crates to at least 0.84.0.

A corresponding PR has also been submitted.

LoicRiegel:
It looks like it's good now. I haven't tried to actually compile the rust plugin example, but in the [creating-a-plugin-in-rust](https://www.nushell.sh/contributor-book/plugins.html#creating-a-plugin-in-rust) chapter, the correct ``input_output_type`` is used and nu-protocol and nu-plugin crates are pinned to 0.92.0
So I wonder if we can close this issue

fdncred:
I don't want to close it without it being tested. It also needs to be updated to 103.

LoicRiegel:
Okay I can test it when I have time

[...]

PR #1908, opened 2025-05-01
Fix plugins streaming sample

### LINKED ISSUE #1734: contributor-book/plugins

Kissaki:
Please verify as I am not very familiar with Rust.

Tested working with a `seq` stream, a string, and a list of values as I would expect from the implementation.
Tested `help len` and unexpected type in as well.

Content that follows on the page has not been tested or adjusted.

Resolves #1734 and #1066

### DIFF
--- a/contributor-book/plugins.md
+++ b/contributor-book/plugins.md
@@ -328,27 +328,20 @@ impl PluginCommand for Len {
                     Value::int(length as i64, call.head).into_pipeline_data()
                 )
             },
-            input => {
-                // Handle a string
-                let span = input.span().unwrap_or(call.head);
-                let value = input.into_value(span);
-                match &value {
-                    Value::String { val, .. } => Ok(
-                        Value::int(val.len() as i64, value.span()).into_pipeline_data()
+            PipelineData::Value(Value::String { val, .. }, _) => {
+                Ok(Value::int(val.len() as i64, call.head).into_pipeline_data())
+            },
+            _ => Err(
+                LabeledError::new(
+                    "Expected String or iterable input from pipeline",
+                ).with_label(
+                    format!(
+                        "requires string or iterable input; got {}",
+                        input.get_type(),
                     ),
-                    _ => Err(
-                        LabeledError::new(
-                            "Expected String or iterable input from pipeline",
-                        ).with_label(
-                            format!(
-                                "requires string or iterable input; got {}",
-                                value.get_type(),
-                            ),
-                            call.head,
-                        )
-                    ),
-                }
-            }
+                    call.head,
+                )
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #1909, opened 2025-05-01
Update Plugins with v0.104.0 dependencies

Kissaki:
Plugins compiled with v0.103.0 do not run in Nushell v0.104.0.

This sample has been compiled in v0.104.0 and runs and behaves like expected.


fdncred:
Thanks

PR #1910, opened 2025-05-01
Fix code font family declaration

Kissaki:
VuePress puts our `:root` declaration before their own, meaning the declaration we intended as an override was overridden instead.

Use higher specificity (`body`) to make sure it takes precedence.

This is a follow-up to #1907


Kissaki:
Oooh, looks good on my Android phone now 🎉

fdncred:
Looks good on my mac. Thanks.

PR #1925, opened 2025-05-27
Fix: color theme examples

levitte:
The color theme examples in book/coloring_and_theming.md weren't quite right.

### DIFF
--- a/book/coloring_and_theming.md
+++ b/book/coloring_and_theming.md
@@ -602,11 +602,11 @@ If you are working on a light background terminal, you can apply the light theme
 
 ```nu
 # in $nu.config-path
-use config light-theme   # add this line to load the theme into scope
+use std config light-theme   # add this line to load the theme into scope
 
 $env.config = {
   # ...
-  color_config: (light_theme)   # if you want a light theme, replace `$dark_theme` to `$light_theme`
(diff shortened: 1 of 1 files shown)


book/coloring_and_theming.md
```diff
@@ -602,11 +602,11 @@ If you are working on a light background terminal, you can apply the light theme
 
 ```nu
 # in $nu.config-path
-use config light-theme   # add this line to load the theme into scope
+use std config light-theme   # add this line to load the theme into scope
 
 $env.config = {
   # ...
-  color_config: (light_theme)   # if you want a light theme, replace `$dark_theme` to `$light_theme`
+  color_config: (light-theme)   # if you want a light theme, replace `$dark_theme` to `$light_theme`
```
NailFec, line 609: I think this comment is misleading. It appears to be a leftover from [nushell PR #5804](https://github.com/nushell/nushell/pull/5804/), which modified `default_config.nu`. However, that file has since been deprecated. I suggest removing this comment or changing it, as it no longer reflects the current state of the configuration.

```suggestion
  color_config: (light-theme)
```

fdncred: I think `(light-theme)` or `(dark-theme)` will work as long as they're brought into scope with use.

PR #1928, opened 2025-06-03
Add simple example uses in `closures.md` (nushell/issues/15845)

raoulkent:
See; https://github.com/nushell/nushell.github.io/issues/1929


fdncred:
Thanks

PR #1930, opened 2025-06-05
feature(make_docs.nu): adjust the input/output section generation

raoulkent:
Now uses the data directly from `help {command}` to ensure that the table
is fully populated by all input/output combinations.

Is meant to address https://github.com/nushell/nushell.github.io/issues/1931

However, it is lacking in that it now is not clear which of the IO pairs is the default etc (details found when inspecting with `scope commands`. However, `scope commands` instead hides details when applying deduping logic
```scope.rs
sigs.dedup_by(|(k1, _), (k2, _)| k1 == k2);
```

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -1,4 +1,3 @@
-
 def plugin-paths [ nu_path?: path ] {
     const PLUGINS = [
         nu_plugin_inc,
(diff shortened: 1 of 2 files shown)


fdncred:
> However, it is lacking in that it now is not clear which of the IO pairs is the default 

I don't think there is a default. Just a list.

raoulkent:
@fdncred @132ikl cheers. Reopened the pr

make_docs.nu
```diff
@@ -243,10 +242,13 @@ $"## Notes
         # FIXME: Parentheses are required here to mutate $input_output, otherwise it won't work, maybe a bug?
         $input_output = ($input_output | append [[input output]; [$input $output]])
     }
-    let in_out = if ($input_output | length) > 0 {
-        let markdown = ($input_output | sort-by input | to md --pretty | str replace -a '<' '\<' | str replace -a '>' '\>')
-        ['', '## Input/output types:', '', $markdown, ''] | str join (char newline)
-    } else { '' }
+    # Input/output types: use help commands
+    let input_output_table = help commands | where name == $command.name | get input_output | first | to md
```
132ikl, line 246:
```suggestion
    let input_output_table = (
        help commands
        | where name == $command.name
        | get input_output
        | first
        | to md --pretty
        | str replace -a '<' '&lt;'
        | str replace -a '>' '&gt;'
    )
```

This replicates the previous behavior (old line 247), and also switches the `<`/`>` to their HTML entity counterparts instead of just escaping them with `\`

132ikl (changes requested):
almost there, thank you!

132ikl:
thanks!

PR #1942, opened 2025-06-11
Fix broken link in control_flow.md

axelcool1234:
This currently sends you to "https://www.nushell.sh/book/book/thinking_in_nu.html" instead of "https://www.nushell.sh/book/thinking_in_nu.html".


fdncred:
Thanks

PR #1943, opened 2025-06-11
Remove awkward newline from scripts.md

axelcool1234:
Removed a newline from a code example in order to match its counterparts in this file (every other code example has "# myscript.nu" directly above the code, without an additional newline in between them). I apologize if my pedantism is a little much here! If it is, let me know and I'll stop.


fdncred:
Thanks

PR #1945, opened 2025-06-12
Fix broken link in creating_modules.md

axelcool1234:
The link as of now sends you to https://www.nushell.sh/book/modules/scripts.html#parameterizing-scripts instead of https://www.nushell.sh/book/scripts.html#parameterizing-scripts


sholderbach:
Thanks!

PR #1948, opened 2025-06-21
feat: update available table modes

echasnovski:
This is a follow up of nushell/nushell#16013. Its PR template says to update this documentation after the PR is merged.

This also adds missing new-ish `'single'` table mode.

One issue is that I am not sure if this is appropriate to merge right away, as `'double'` mode is for yet-not-released 0.106.0 version. If you want, I can split this PR into two: one adding `'single'` (which can be merged now) and one for `'double'` (for later).

### DIFF
--- a/book/coloring_and_theming.md
+++ b/book/coloring_and_theming.md
@@ -21,6 +21,7 @@ The options for `$env.config.table.mode` can be listed with `table --list`:
 - `compact`
 - `default`
 - `dots`
+- `double`
 - `heavy`
 - `light`
 - `markdown`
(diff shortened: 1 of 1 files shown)


fdncred:
I usually land this type of thing. maybe I'm wrong about doing that?

echasnovski:
I wouldn't say wrong per se, but it can result in user confusion. Like, if new users now decide to adjust table mode and excitedly tries `double` after reading about it on official website, they will be mighty confused about the error.

With frequent releases (as what is seems like with Nushell right now) probably not a big deal though.

PR #1957, opened 2025-06-26
book: merge in settings, instead of suggesting to override them

Axlefublr:
the indentation was somewhat inconsistent, I picked 4 spaces as that is probably the most visible/clear


fdncred:
I like 2 myself, but whatev. Thanks

PR #1958, opened 2025-06-26
fish_completer: escape all singlequotes

scarlet-storm:
Fixes an oopsie in PR #1955


fdncred:
Thanks for the cleanup

PR #1963, opened 2025-07-03
optimize fish_completer path lookups

scarlet-storm:
Check if string needs escaping before doing a path lookup. This avoids looking up the path, unless the completion contains special characters that need escaping


fdncred:
Thanks. I'd like the $v to be a bit more verbose in name and also, it would be good to explain what's going on in the section below that describes what's happening. The cookbook needs to be able to explain itself.

scarlet-storm:
> Thanks. I'd like the $v to be a bit more verbose in name and also, it would be good to explain what's going on in the section below that describes what's happening. The cookbook needs to be able to explain itself.

I modified the section to add a short exlanation on the changes

PR #1972, opened 2025-07-12
Improve cookbook setup env setup

### LINKED ISSUE #1926: Remove refernece to manually splitting path by `char esep` in setup/cookbook.md
[Context](https://www.nushell.sh/cookbook/setup.html#configure-your-path-and-other-environment-variables)

I think we should avoid suggesting this directly as it's probably more confusing than it needs to be with path's ENV_CONVERSIONS being default now. We should switch this to link to https://www.nushell.sh/book/configuration.html#path-configuration probably

Kissaki:
Resolves #1926

### DIFF
--- a/cookbook/setup.md
+++ b/cookbook/setup.md
@@ -11,15 +11,35 @@ There are other ways to view these values and variables, however setting up your
 
 ### Configure your path and other environment variables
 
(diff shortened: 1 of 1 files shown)


fdncred:
ok, thanks.

PR #1974, opened 2025-07-12
Fix swapped closure parameters

Kissaki:
As per [`reduce` command documentation examples][reduce], the reduce closure has parameters `|it, acc|`.

This accidental swap was part of the introduction of the example in #1928.

[reduce]: https://www.nushell.sh/commands/docs/reduce.html


fdncred:
Good catch. Thanks

Issue #1929, opened 2025-05-29
Better Closure Docs


Android789515:
### Related problem

The docs have very little information on each subject.  Just one tiny example.

Going through the closures section:

[Common commands that can be used with a closure](https://www.nushell.sh/lang-guide/chapters/types/basic_types/closure.html#common-commands-that-can-be-used-with-a-closure)

    all
    any
    collect
    do
    each
    explain
    filter
    group-by
    interleave
    items
    par-each
    reduce
    skip until
    skip while
    take until
    tee
    update
    upsert
    zip

None of these methods are explained in further detail.  I have no idea what items does, what par-each does, group-by, explain, etc.

### Describe the solution you'd like

Document each method as well as possible usages.

sholderbach:
I actually kind of disagree with adding this to the *Language Reference manual*. The purpose of the reference manual primarily is to properly explain each mentioned concept cleanly and correctly and only use examples to explain particular important concepts (e.g. that a closure captures values by value and closes over them so they aren't affect by the outer scope anymore)
Explaining the use of the commands would be better suited to the command reference and generally helpful examples should probably move to the larger book where more than just the core patterns and rules you need to know can be explained. If the reference is long but trivial it defeats the purpose as a quick and useful reference to understand concepts and give a refresher to the intermediate users.

@NotTheDr01ds added those references to commands that interact with certain types from the `scope commands` set to better build connections and discover concepts worth explaining.

I think in the language reference we should cross link those to the command reference but I would recommend collections of examples to rather be explained in the book or if its just examples (that are helpful) the cookbook

raoulkent:
@sholderbach That's fair. Would it be better to then link to the book for examples then, and move the examples there?

I'm thinking from the new users perspective, they may not know where to look if different types of information can be found if it's not super clearly specced out early in the discovery process.

Kissaki:
I agree with @raoulkent, the language reference *(although the URL/dir slug is called "guide")* should not describe other commands with examples - especially not when it does not demonstrate a language concept, but more how specific commands work/use them.

The current docs have examples for `each`, `where`, `sort-by`, `reduce` - added by #1928. I checked all of them, and they all have examples in the command documentation. I think that's enough. No need to duplicate that elsewhere.

[...]

PR #1975, opened 2025-07-12
Drop examples from closure language reference

Kissaki:
Relates to #1929

The language reference should be focused on the language concepts, not describe how various commands make use of it, and especially not in the form of examples.

All these commands have closure examples in the docs comments. All listed commands link to their respective docs.

Dropping the examples increases focus of the *reference* doc, and reduces risks of doc duplication and of command changes leading to outdated lang ref docs.


sholderbach:
> I like examples as a way to help explain things.

We are talking about the Language Guide/Reference not the general book. Examples for concepts / distinct specialties of closures yes, stamp collecting of commands that happen to use closures no.

Kissaki:
@fdncred Where do you think would be a fitting place for these kinds of or these examples specifically?

Closures are documented in the Book in the Types of data section. But as a relatively concise block. https://www.nushell.sh/book/types_of_data.html#closures

---

The language reference documents that closures exist.

The fact that they can different forms and parameters is a consequence of what they are. I don't think a list of specific command examples is what should be documented here.

If we want to show that there is variance, beyond just saying that there is, we can do so with two or three examples, with or without specific commands. I would certainly want to evade command headlines as if we were talking about those commands here.

---

In terms of examples in general, those commands already have closure examples for how they are used in those commands.

fdncred:
I'm not suggesting that it be entirely duplicated but having a couple examples and then a "to see more information on closures see _some_link_" would probably be helpful.

Kissaki:
While going through the page I also created #1981

I can see now how the examples can be useful here too. in #1982 I reduced them into something - in my eyes - more appropriate for the lang reference.

PR #1976, opened 2025-07-12
Add `$env.NU_LOG_LEVEL` to Special Variables

Kissaki:


book/special_variables.md
```diff
@@ -118,6 +118,37 @@ A list of directories which will be searched when using the `source`, `use`, or
...
+# => 2025-07-12T21:27:30.080|DBG|1111
+# => 9
+
+nu -c '1 | print; use std/log; $env.NU_LOG_LEVEL = "debug"; log debug 1111; 9 | print'
+# => 1
+# => 2025-07-12T21:27:57.888|DBG|1111
+# => 9
+```
+
+Note that `$env.NU_LIB_DIRS` is different from `nu --log-level`, which sets the log level for Nushell commands. It does not influence the `std/log` logging used in custom commands and scripts.
```
Kissaki, line 141: I was think about whether there is a good, fitting qualifier for Nushell's own commands. In the end I chose not to add any qualifier, and only call the custom commands custom commands as a differentiation (like they are called in the docs).

```suggestion
Note that `$env.NU_LIB_DIRS` is different from `nu --log-level`, which sets the log level for native Nushell commands. It does not influence the `std/log` logging used in custom commands and scripts.
```

```suggestion
Note that `$env.NU_LIB_DIRS` is different from `nu --log-level`, which sets the log level for (native) Nushell commands. It does not influence the `std/log` logging used in custom commands and scripts.
```

book/special_variables.md
```diff
@@ -118,6 +118,37 @@ A list of directories which will be searched when using the `source`, `use`, or
 - [Module Path](./modules/using_modules.md#module-path)
 - [Configuration - `$NU_LIB_DIRS`](./configuration.md#nu-lib-dirs-constant)
 
+### `$env.NU_LOG_LEVEL`
+
+The [standard library](/book/standard_library.md) offers logging in `std/log`. The `NU_LOG_LEVEL` environment variable is used to define the log level being used.
```
Kissaki, line 123: Is this - with the mention and reference of standard library - sufficient to clarify the conditional nature of it? That it is not available in a fresh shell session?

Kissaki (commented):
Is it fitting to add a standard library special variable to the list of special variables?

I think it is.

It's a good cross-reference to the availability of the standard library and this submodule in particular.

In this case, it's also a good cross-discoverability for `--log-level`.

book/special_variables.md
Kissaki:
```suggestion
Note that `$env.NU_LOG_LEVEL` is different from `nu --log-level`, which sets the log level for Nushell commands. It does not influence the `std/log` logging used in custom commands and scripts.
```

book/special_variables.md
fdncred: I don't think it's specific enough.

```suggestion
The [standard library](/book/standard_library.md) offers logging in `std/log`. The `NU_LOG_LEVEL` environment variable is used to define the log level being used for custom commands, modules, and scripts.
```

PR #1985, opened 2025-07-16
Manually link page next-prev across child borders

### LINKED ISSUE #1818: Book navigation Next link skips chapter contents resulting in confusing navigation.
The book's page footer that provides the _Prev_ and _Next_ navigation seems to pick the next chapter/page at the same level instead of any subpages that are children of that page. 

This works OK for single page chapters like _"Introduction"_  but is problematic for multi-page chapters like [Getting Started](https://www.nushell.sh/book/getting_started.html) The [next link at the bottom of the page](https://www.nushell.sh/book/getting_started.html) links to "Nu Fundamentals" (the next chapter) skipping over all the pages in the Getting Started chapter.  It seems like next should be descending into the chapter  linking to the first page which is [Quick Tour](https://www.nushell.sh/book/quick_tour.html).  

Once on the Quick Tour page, it points to the next page at the same level so the navigation makes sense.

But the same problem is see on the [Modules page](https://www.nushell.sh/book/modules.html) where _Next_ links to _Overlays_ instead of _Using Modules_.

Kissaki:
The previous and next navigation links at the bottom of the page get filled automatically by vuepress, sourced from the (locale-specific) sidebar configuration.

We define the sidebar through sections - which may be single page sections or a section with child pages.

Previously, the website did not guide the user through our book.

* Section prev always linked to previous section, never previous sections last child if available
* Section next never linked to first child of section, if available
* First child did not offer a prev link
* Last child did not offer a next link

I did not see documentation about customizing the auto-discover behavior, nor the sidebar structure dependency of the automatic determination. But page frontmatter can override the prev and next links.

As a fixup, overriding the automatic discovery from the sidebar, we manually set the prev and next links on section child borders. The link text matches the headline/sidebar, and the links the sidebar structure. Given that this is manual configuration, there's a risk of it eventually being missed and becoming outdated or wrong when structure changes are made.

I did not test. I don't have a local nodejs env.

Resolves #1818

* https://v1.vuepress.vuejs.org/theme/default-theme-config.html#prev-next-links
* https://v1.vuepress.vuejs.org/guide/i18n.html#default-theme-i18n-config
* https://ecosystem.vuejs.press/themes/default/frontmatter.html#prev

### DIFF
--- a/book/advanced.md
+++ b/book/advanced.md
@@ -1,3 +1,11 @@
+---
+prev:
+  text: How Nushell Code Gets Run
+  link: /book/how_nushell_code_gets_run.md
+next:
+  text: Standard Library (Preview)
+  link: /book/standard_library.md
+---
 # (Not so) Advanced
 
 While the "Advanced" title might sound daunting and you might be tempted to skip this chapter, in fact, some of the most interesting and powerful features can be found here.
--- a/book/background_jobs.md
+++ b/book/background_jobs.md
@@ -1,3 +1,8 @@
+---
+next:
+  text: Coming to Nu
+  link: /book/coming_to_nu.md
+---
 # Background Jobs
 
 Nushell currently has experimental support for thread-based background jobs.
--- a/book/cheat_sheet.md
+++ b/book/cheat_sheet.md
@@ -1,3 +1,8 @@
+---
+next:
+  text: Nu Fundamentals
+  link: /book/nu_fundamentals.md
+---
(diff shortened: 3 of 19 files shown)


Kissaki:
@hustcer Book looks fine to me with generated link though because it has no children!?

https://www.nushell.sh/book/ has a Next link to Installation.

I see that install has a default shell sub page. Seems like I missed that? I will take a look.

PR #1990, opened 2025-07-23
Fix `:::` blocks not closing correctly

ropeladder:
There might be more of these problems with `:::` blocks, but this was the page I was reading carefully enough to notice them on.


hustcer:
Thanks!

PR #1994, opened 2025-07-27
Update docs for `save.md`

TheBearodactyl:
Depends on [nushell/nushell#16277](https://github.com/nushell/nushell/pull/16277)


TheBearodactyl:
oh didn't know that
nvm then

PR #1996, opened 2025-07-28
Use null character ending with env

meghprkh:
Line splitting was less safe as some scripts inject `BASH_FUNC_...` env variables which have newlines in them


meghprkh:
(VSCode had this issue too https://github.com/microsoft/vscode-remote-release/issues/9457 )

fdncred:
Thanks

PR #1997, opened 2025-07-29
book: write Control key as just «Ctrl», not «Ctrl>»

das-g:
Also to make this consistent with other parts of the same page, such as
https://github.com/nushell/nushell.github.io/blob/6b4312b1a329498756af4d02c45a03b78a279056/book/line_editor.md?plain=1#L122-L124


fdncred:
Thanks

PR #1999, opened 2025-07-29
Release notes for `0.107.0`

sholderbach:
Please add your new features and breaking changes to the release notes
by opening PRs against the `release-notes-0.107.0` branch.

## TODO
- [x] PRs that need to land before the release, e.g. [deprecations](https://github.com/nushell/nushell/labels/deprecation) or [removals](https://github.com/nushell/nushell/pulls?q=is%3Apr+is%3Aopen+label%3Aremoval-after-deprecation)
- [x] add the full changelog
- [x] categorize each PR
- [x] write all the sections and complete all the `TODO`s

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -90,7 +90,16 @@ export default defineUserConfig({
     },
   },
   head: [
-    ["link", { rel: "preload", href: "/fonts/FiraCode-Regular.woff2", as: "font", type: "font/woff2", crossorigin: "anonymous" }],
+    [
+      'link',
+      {
+        rel: 'preload',
+        href: '/fonts/FiraCode-Regular.woff2',
+        as: 'font',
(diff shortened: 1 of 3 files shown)


ItsHarper:
The "`http` subcommands now keep track of redirects" section seems to have a copy/paste that included some control codes or something

132ikl:
> I assume these will get rendered correctly on the website? It's quite unfortunate that it makes the code snippets unreadable with the GitHub viewer (especially since there's no website preview). What was the goal of this change?

It's so that we can get nice syntax highlighting on the blog, including for results where the stock syntax highlighting doesn't cut it. This is especially when we would otherwise have to use an image, like this example which needs real syntax highlighting to make sense:

image


but it's still a lot nicer for things where we would otherwise have no syntax highlighting at all, like errors:

image

Issue #2003, opened 2025-08-04
Spread Operator not properly spread in example


AlOwain:
_This is an issue, and not a pull request because I am new to Nushell and would like to confirm that this is in fact not proper behavior._

In [Programming In Nu#Operators:262](https://github.com/nushell/nushell.github.io/blob/22957e0164847926d9ea0b6f83d97ddd9626f73b/book/operators.md?plain=1#L262) (shown below)

```nu
foo "bar" "baz" ...[1 2 3]
# => [false, bar, baz, [1, 2, 3]]
foo "bar" "baz" [1 2 3]
# => [false, bar, baz, [[1, 2, 3]]]
```

The example **(and many others in the same file)** seem to indicate in the comments, that the default behavior of appending a list to another is that it is encapsulated as another list, and spreading it only appends the list as-is. My assumption would be that it wouldn't, and trying it out in the Shell _seems_ to indicate so, for example:

```nu
~> ["bar" "baz" ...[1 2 3]] | to nuon
[bar, baz, 1, 2, 3]
~> ["bar" "baz" [1 2 3]] | to nuon
[bar, baz, [1, 2, 3]]
```
> The absence of `foo` is intentional, as it does not change the demonstrability of the issue.

> As a precursor, I had attempted to check for issues, or pull requests that mention this, it does not seem to have been addressed. The `git blame` result is a commit in the pull request #1692, which has many changes indicating that it could have just slipped by.

AlOwain:
Ah, thank you very much. I apologize for the hassle, I overlooked it.

AlOwain:
You are definitely right though, having it be a record makes it clearer, perhaps while reading the documentation, I'll try reviewing other examples for similar patterns, I'll try to avoid verbosity.

Finally, thank you very much, the documentation is superb, it's on me that I missed it.

PR #2004, opened 2025-08-05
Use a record instead of a list to clarify arguments in Programming in Nu#Operators examples.

AlOwain:
In [Programming in Nu#Operators:262](https://github.com/nushell/nushell.github.io/blob/22957e0164847926d9ea0b6f83d97ddd9626f73b/book/operators.md?plain=1#L262) inferring the arguments from the list position could introduce some misunderstanding.

This PR is a result of my own misunderstanding (see issue #2003) in how a list was spread. I recognize that my misunderstanding was not a result of a lack of clarity, but that I simply did not understand how it operated, and had not properly scrutinized the following examples.

However, I do believe that this makes things clearer, at the risk of being needlessly tedious. This is left for your consideration.

### DIFF
--- a/book/operators.md
+++ b/book/operators.md
@@ -247,7 +247,9 @@ external command.
 Here is an example custom command that has a rest parameter:
 
 ```nu
-def foo [ --flag req opt? ...args ] { [$flag, $req, $opt, $args] | to nuon }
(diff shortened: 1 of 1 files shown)


ysthakur (approved):
LGTM, thanks for the quick PR

PR #2014, opened 2025-08-20
Add hint for generating \0 character to strings guide

rktjmp:
The Working with Strings guide lists supported escape characters (eg: `\n`, `\t`) but `\0` is not supported by double quote strings.

Include an example of generating `NUL` via `\u{0}`, as well as pointing to combining string interpolation with `chars --integer 0`.

### DIFF
--- a/book/working_with_strings.md
+++ b/book/working_with_strings.md
@@ -58,8 +58,7 @@ Nushell currently supports the following escape characters:
 - `\u{X...}` - a single unicode character, where X... is 1-6 hex digits (0-9, A-F)
  
 To create a `\0` (`NUL`) character, you may use `\u{0}` or
-[string interpolation](#string-interpolation) with [`char`](/commands/docs/char.md),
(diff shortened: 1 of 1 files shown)


book/working_with_strings.md
```diff
@@ -56,6 +56,10 @@ Nushell currently supports the following escape characters:
 - `\n` - newline (line feed)
 - `\t` - tab
 - `\u{X...}` - a single unicode character, where X... is 1-6 hex digits (0-9, A-F)
+ 
+To create a `\0` (`NUL`) character, you may use `\u{0}` or
+[string interpolation](#string-interpolation) with [`char`](/commands/docs/char.md),
+e.g. `(char --integer 0)`.
```
fdncred, line 62: or `char nul`, `char null_byte`, `char zero_byte` as seen in `char --list`. It may also be worth nothing that things like `\u{0}` only work when surrounded by double quotes because in nushell double quotes interpret escapes.

PR #2017, opened 2025-08-23
Report and forgive missing plugins in make_docs.nu

### LINKED ISSUE #2010: `make_docs.nu` does not take kindly to being run without plugins installed
Here's the abbreviated error output. In reality, I had to use `head` (forgive me for not having the `nu` equivalent memorized yet 😆) to capture the start of the error, because it was so long that it filled my terminal's buffer.

```nu
> nu make_docs.nu
Error: nu::shell::io::not_found

  × Not found
  help: '/usr/bin/nu_plugin_inc' does not exist

Error: nu::shell::unsupported_input

  x Unsupported input
     ,-[/home/harper/nushell.github.io/make_docs.nu:446:11]
 445 |         scope commands
 446 |         | join (command-names) name
     :           ^^|^|
     :             | `-- (List { vals: [Record { val: Record { inner: [("name", String { val: "alias", internal_span: Span { start: 144140, end: 144154 } }), 

<It goes on like this for ~forever>

144138, end: 144152 } }], internal_span: Span { start: 144138, end: 144152 } }, Nothing, String String)
     ·             ╰── (PipelineData<table>, table, string, string)
 447 │         | sort-by category
     ╰────

Error: nu::shell::non_zero_exit_code

  × External command had a non-zero exit code
   ╭─[entry #36:1:1]
 1 │ nu make_docs.nu
   · ─┬
   ·  ╰── exited with code 1
   ╰────
```

I see there's code that attempts to print a nice error in this case, but it's clearly not working.

It does behave much better when using the wrapper (though even that doesn't display the error message written in the script):

```nu
> source make_docs.nu
> make_docs
Error: nu::shell::io::not_found

  x Not found
  help: '/usr/bin/nu_plugin_inc' does not exist

Error: nu::shell::non_zero_exit_code
```

Kissaki:
Example screenshot of constructed error case:

image

Resolves #2010

---

I would have preferred an error exit, but I'm not sure how to do that in a way that it works for this script both when calling it as a script via `nu` as well as when running it via `source` then `make_docs` command.

Checking for invalid paths, reporting them and skipping them seems like a reasonable improvement, especially in the context of #2010 (runtime errors further down the script and command pipeline).

### DIFF
--- a/make_docs.nu
+++ b/make_docs.nu
@@ -14,12 +14,17 @@ def plugin-paths [ nu_path?: path ] {
         _ => ($nu_path | path dirname)
     }
 
-    $PLUGINS | each {|plugin|
+    let plugin_paths = $PLUGINS | each {|plugin|
         match (sys host | get name) {
             'Windows' => $'($nu_dir | path join $plugin).exe'
             _ => $'($nu_dir | path join $plugin)'
         }
     }
+
+    let decorated = $plugin_paths | wrap path | insert exists {|x| $x.path | path exists }
+    $decorated | where not exists | each {|x| print $"(ansi red)No plugin under path '($x.path)', will be skipped...(ansi reset)" }
+
+    $decorated | where exists | get path
 }
 
 # get all command names from a clean scope


hustcer:
Thanks, I can accept this change, and I won't create the PR if any error happens next time.

PR #2026, opened 2025-08-30
docs(zh-CN): Revoke code examples to English

jexjws:
This commit updates the code examples in the Chinese documentation for data types (`types_of_data.md`).

Variable names, string literals, and comments within the code blocks have been revoked from Chinese to English. This change makes the examples more idiomatic and consistent with Nushell's English-based syntax, while keeping the surrounding explanatory text in Chinese.


hustcer:
Thanks

PR #2030, opened 2025-09-02
fix: fix incorrect winget install cmd

rmoraes92:
the previous version of the recommended command line was throwing an error.

this new proposed one will actually install the correct nushell from the winget store.


fdncred:
why is it necessary to use `--source winget`?

fdncred:
`winget install nu` seems to work for me.

PR #2041, opened 2025-09-20
Fix references to renamed range command

nome:
Since nu 0.102.0, the range command has been renamed to slice. Two pages in the book still referenced the old command name (with corresponding broken hyperlinks).


sholderbach (approved):
Thanks I forgot to follow up on that

PR #2047, opened 2025-09-30
docs: fix syntax

bf:
Add missing `)` to docs.

### DIFF
--- a/book/hooks.md
+++ b/book/hooks.md
@@ -192,7 +192,7 @@ $env.config = ($env.config | upsert hooks {
             code: 'print $"changing directory from ($before) to ($after)"'
         }
     }
-}
+})
 ```
 
 ## Examples


fdncred:
Are the parentheses even needed at all? This works but it will also overwrite your existing env_change.PWD if it already exists.
```nushell
$env.config = $env.config | upsert hooks {
    env_change: {
        PWD: [{
            code: 'print $"changing directory from ($before) to ($after)"'
        }]
    }
}
```

wsx7524999 (approved):

Bahex:
We should probably go further and update these to use the "modern" style:
```nushell
# Make sure the `PWD` hook list exists
$env.config.hooks.env_change.PWD = $env.config.hooks.env_change.PWD? | default []

# Add hook(s)
$env.config.hooks.env_change.PWD ++= [
    {
        condition: {|_, after| $after == /some/path/to/directory }
        code: 'def foo [] { print "foo" }'
    }
    {
        condition: {|before, _| $before == /some/path/to/directory }
        code: 'hide foo'
    }
]

$env.config.hooks.env_change.PWD ++= [
    {
        code: 'print $"changing directory from ($before) to ($after)"'
    }
]
```

wsx7524999:
I like to merged this pull request but how do I get access to the repository?

PR #2050, opened 2025-10-02
fix error USERNAME not exist in ssh-agent workaround script

ValouBambou:
I tried the script on this page and was surprised to have it failed with the following error.

```
Error: nu::shell::column_not_found

  × Cannot find column 'USERNAME'
    ╭─[/home/user/.config/nushell/env.nu:27:69]
 26 │     let ssh_agent_file = (
 27 │         $nu.temp-path | path join $"ssh-agent-($env.USER? | default $env.USERNAME).nuon"
    ·                                                                     ──────┬──────┬
    ·                                                                           │      ╰── value originates here
    ·                                                                           ╰── cannot find column 'USERNAME'
 28 │     )
    ╰────
 ```

I fixed it by removing the part of the code where it defaults to the non-existing USERNAME. Maybe this var is needed in some other way, but in any case, I think this page needs to be updated. 

PS: Obviously if this introduces error for other users ping me and I will update this PR


ValouBambou:
I confirm it works perfectly. I will modify my PR.

fdncred:
Thanks

PR #2058, opened 2025-10-18
Fix typo/wording in v108 release notes

Kissaki:
While I assume it's a word typo of 'show' -> 'should', the two closed bug ticket samples work now, so I would find it more appropriate to make it a definite statement. 'should' would unnecessarily leave open why it may still not be working.


fdncred:
sure. thanks

PR #2078, opened 2025-11-22
Update vuepress plugins (fix CMD/Ctrl+K!)

xav-ie:
Hooray! cmd/ctrl+k finally work. There was previously annoying bug that caused it not to work and it is finally fixed now in vuepress. 

I just bumped all docsearch deps to latest since they are managed and released together. I was getting conflicting peer deps issues otherwise.

Basically, everything still works, and I would expect this because the mantainer of vue-press, Mister-Hope, is very careful. I don't think this many package updates is concerning.

Tested locally to verify
- feeds
- shiki highlights
- searching 
- translations
- building

Still work. I am sure all the other plugins have great updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,65 +10,141 @@
       "hasInstallScript": true,
       "license": "MIT",
       "devDependencies": {
-        "@vuepress/bundler-vite": "2.0.0-rc.24",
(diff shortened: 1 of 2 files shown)


hustcer:
Thanks

PR #2100, opened 2026-01-02
Book: Fix outdated substring output

Drazape:
In the *NuShell book/Nu Fundamentals/Working with strings/The `str` command/Substrings* (`book/working_with_strings.md`),
the command `'Hello World!' | str substring 4..8'`'s output is given as `o Wo`—which would be correct if `8` wasn't included and the output would be of 4 characters, as given in the book.

Running the command internally has shown that 8—in fact inculded in the substring so the correct output shown in the comment should be `o Wor` (of 5 characters) as it is in the cheat sheet and not `o Wo`


fdncred:
Thanks

PR #2104, opened 2026-01-04
Non-conflicting .list filename in install instructions

pcrockett:
`fury.list` is an example filename that comes from Gemfury's APT repo documentation. However the name is entirely too generic, and can easily conflict with other projects who also host their APT repos on Gemfury.

Making the name Nushell-specific makes it less likely that users will blow away existing Gemfury repo configurations by accident.


hustcer:
Thanks

PR #2124, opened 2026-02-25
Update piping alias section for clarity

c-git:
Clarify that aliases do not support multiple commands and provide examples for defining commands without parameters.


c-git:
Still ended up making another change, so will try again another time to read it over. 

Once I'm finally done, do I need to squash the commits or will you squash on merge?

PR #2131, opened 2026-02-28
Minor typo fixes

c-git:
Found another one. I think I'm done reading for today. What frequency should I make the pull requests? I don't want to do them too often but I also don't want them to get too stale. What frequency seems good to you. I'll just set a reminder to send them at that frequency instead of when I find them.

### DIFF
--- a/book/stdout_stderr_exit_codes.md
+++ b/book/stdout_stderr_exit_codes.md
@@ -20,7 +20,7 @@ Without the pipeline, Nushell will not do any redirection, allowing it to print 
 
 Another common stream that external applications often use to print error messages is stderr. By default, Nushell does not do any redirection of stderr, which means that by default it will print to the screen.
 
-But you can do pass stderr to a command or a file if you want to:
+But you can pass stderr to a command or a file if you want to:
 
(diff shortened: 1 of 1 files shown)


LikeLakers2:
> What frequency should I make the pull requests? I don't want to do them too often but I also don't want them to get too stale. What frequency seems good to you.

I'm not a maintainer, so please don't take my word as gospel, or as a replacement for the maintainer's word.

However, for something like this, I think it'd make sense to just add new commits as you find things to fix. If a PR gets merged, and you find more things to fix, then you make a new PR.

That said, I also get if you don't want to keep opening new PRs - maybe it's annoying to make new branches, or you feel like it's bothering the maintainers.

So I think what you've done here is equally as valid - marking the PR as a draft until you've had time to go through everything.

Again, I'm not a maintainer of nushell, so it's possible the maintainers may disagree... but I still hope my viewpoint helps.

c-git:
Life has been pulling me in other directions. Haven't found anymore typos since. I'll cap the number of PRs to at most one per week. Does that seem ok?

PR #2140, opened 2026-03-23
completions: Document filter option and span and display_override fields

ysthakur:
Document recently-added completion features:
- `filter` option for disabling filtering: https://github.com/nushell/nushell/pull/17253
- `span` suggestion record field for replacing specific span instead of just the last argument: https://github.com/nushell/nushell/pull/15888
- `display_override` suggestion record field for customizing suggestion display value: https://github.com/nushell/nushell/pull/17330 (behavior updated in https://github.com/nushell/nushell/pull/17060)

Here's what the code block for showing the results of setting `style` and `display_override` looks like:

image

### DIFF
--- a/book/custom_completions.md
+++ b/book/custom_completions.md
@@ -4,7 +4,7 @@ Custom completions allow you to mix together two features of Nushell: custom com
 
 A completion is defined in two steps:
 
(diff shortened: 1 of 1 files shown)


fdncred:
Should we land this?

PR #2143, opened 2026-03-28
Implement page search URL + Opensearch XML

### LINKED ISSUE #1093: FR: Generate search URL
This might be a niche, but it would be really nice to generate a search URL instead of showing directly the results in the Algolia search.

Screenshot 2023-10-03 at 12 58 34

for example, in this case it could be https://www.nushell.sh/search/TEST, and it could be used for queries with https://www.nushell.sh/search/<query>

Then it could be used to generate quick searches from the browser using `!nu` for example.

I can understand that this is not priority, but it would be really cool giving that the shell is a daily use tool and nushell is much different than bash or fish.

ian-h-chamberlain:
Closes #1093 

I tend to create a lot of [shortcuts to doc searches](https://kb.mozillazine.org/Using_keyword_searches) in my browser, so I wanted the same thing for Nushell! I decided to take a crack at implementing the search URL logic I've seen used by many other docs websites.

https://github.com/user-attachments/assets/fc48510c-64a7-4645-83b7-ac123e76d5a1

Also set up [Opensearch XML](https://developer.mozilla.org/en-US/docs/Web/XML/Guides/OpenSearch#overview) so some browsers can more easily add a search engine:

image

## Implementation

- Wrap the `DocSearch` component provided by the `docsearchPlugin`, and override its registration with our custom component

In the new `URLDocSearch` component:

- Prefill and execute search when a page loads with the `search` param set, at `onMounted`

- Add an `input` event handler, which updates the URL bar when the search bar is updated. Use a 500ms debounce to avoid flooding the user's browser history with partial searches

### DIFF
--- a/.vuepress/components/URLDocSearch.vue
+++ b/.vuepress/components/URLDocSearch.vue
@@ -1,5 +1,5 @@
 <template>
-  <DocSearch :options="docsearchOptions" />
+  <DocSearch />
 </template>
 
 <script setup lang="ts">
@@ -18,7 +18,6 @@ const SEARCH_INPUT_ID = 'docsearch-input';
 const router = useRouter();
 const route = useRoute();
 
-const docsearchOptions = ref<DocSearchOptions>({});
 const inputElement = ref<HTMLInputElement>();
 const isNavigating = ref(false);
 
--- a/.vuepress/components/URLDocSearch.vue
+++ b/.vuepress/components/URLDocSearch.vue
@@ -1,0 +1,46 @@
+<template>
+  <DocSearch :options="docsearchOptions" />
+</template>
+
+<script setup lang="ts">
+import { DocSearch } from '@vuepress/plugin-docsearch/client';
+import { useDebounceFn, useEventListener } from '@vueuse/core';
+import { onMounted, onUnmounted, ref } from 'vue';
+import { useRouter, useRoute } from 'vue-router';
+
+const SEARCH_KEY = 'search';
+const docsearchOptions = ref({});
+const router = useRouter();
+const route = useRoute();
+
+// Handle initial search query, if one is set
+onMounted(() => {
+  const query = new URL(window.location.href).searchParams.get(SEARCH_KEY);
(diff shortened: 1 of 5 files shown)


fdncred:
sounds cool!

.vuepress/components/URLDocSearch.vue
```diff
@@ -0,0 +1,46 @@
...
+// There's some debounce builtin to docsearch, this mimics that and should
+// help prevent browser history from getting filled with partial queries.
+const inputHandler = useDebounceFn(handleSearchInput, 500);
+useEventListener('input', (event) => inputHandler(event as InputEvent));
+
+// Update the URL bar when the search input changes.
+function handleSearchInput(event: InputEvent) {
+  const target = event.target as HTMLInputElement | undefined;
+  const searchQuery = target?.value;
+  if (target?.id !== 'docsearch-input' || !searchQuery) {
```
hustcer: I think If the input value is empty it's better to delete the search param here. Returning early when the input is cleared means the existing `?search=` parameter is not removed. As a result, after the user clears the search and closes the modal, the URL may still retains the old query.

ian-h-chamberlain: Hmm, good point, I think that probably makes sense. I'm not sure how easy it will be to detect the modal being dismissed (since that logic is implemented in a nested child), but it's probably doable with some kind of global event listener at least.

I will try to get to implementing this next week when I should have some time!

ian-h-chamberlain: Thanks for the feedback, I ended up rewriting a lot of the implementation to cover a few different cases:

- Modal is dismissed: clear query param
- Search form is reset ("Clear" button): clear query param
- Navigate to search result: builtin navigation clears query param automatically

hustcer:
Hi @ian-h-chamberlain , sorry it took me so long to circle back on this — thanks for your patience. The latest revision looks great. I pushed a couple of tiny polish commits on top; happy to merge 👍

PR #2145, opened 2026-04-01
doc: change filesize unit tip to warning for clarity

CAESIUS-TIM:
Change from tip to warning to highlight potential pitfall: case does not distinguish between bits and bytes.


Juhan280:
Can you explain  what you are talking about

CAESIUS-TIM:
In general, an uppercase B stands for byte, while a lowercase b represents bit.
Therefore, if file size unit is case-insensitive, this comment had better be marked as a warning to avoid potential pitfalls

fdncred:
I think it's clear enough how it's already written. It says they're insensitive and equivalent.

PR #2146, opened 2026-04-09
Update `metadata set` example of the book

Juhan280:
Also the commands page `metadata set` is also outdated. Not sure how to update them. Are they synced up using a script?


fdncred:
command help gets updated with a script that hustcer runs after a release. everything else is manually updated.

Juhan280:
Ci is not failing in my commit

fdncred:
Thanks

PR #2156, opened 2026-04-19
fix: keychain command

istar410:
The keychain command provided in the doc has a wrong pattern for matching environment variables output by "keychain --eval", there is an extra ";" at the end and the socket path was still in quotes.


fdncred:
Thanks

Issue #1487, opened 2024-08-01
Determine how to handle Vuepress code snippets in docs


NotTheDr01ds:
I just noticed that the Book uses some Vuepress code snippets in places (but not everywhere).  There are some advantages and disadvantages to these:

Advantages:

* (Minor) Can be reused if needed
* (Nice) Indention is much easier if the code-block is nested under a list item

Disadvantages:

* (Major) All of the existing snippets are `.sh` files.  These need to be updated to `.nu` files in order to obtain proper syntax highlighting, of course.
* (Major) Renaming or moving a snippet requires updates to all language translations.
* When a code snippet is updated, it is propagated across all language translations.  This sounds nice, but that may not be desirable if the text around it needs to be updated as well, which will likely be the case.
* Probably won't be portable if we change backends in the future

When I first saw these, I liked the concept, but after dealing with the ramifications, it seems like they might be more problematic than helpful.

At the least, all of them have to be reworked to get syntax highlighting.  Should this just be a single "eliminate Vuepress snippets" enhancement?

fdncred:
I'm fine with standardizing without tiny files. I never really understood why we had separate files instead of just putting them in code fences.

NotTheDr01ds:
When I get a chance, I'll add this guidance to `CONTRIBUTING.md` and then close this out.

PR #2162, opened 2026-05-12
support extracting nu code snippets into standalone .nu files for better downstream trackability

orthonormalremy:
There are many useful recipes that exist in the cookbook and there is no way to track them directly in a script-friendly way (markdown file parsing is not ideal obviously).

This PR adds support for useful standalone code snippets (e.g. snippets that someone might want to put in their `autoload/` dir) to be extracted out into their own .nu files so it can remain the source of truth for the cookbook while also being programmatically fetch-able.

Note: this PR uses the lightweight @vuepress/plugin-markdown-include plugin which is an official VuePress plugin from the same @vuepress/* family already used in this repo

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -6,6 +6,7 @@ import { shikiPlugin } from '@vuepress/plugin-shiki';
 import { defaultTheme } from '@vuepress/theme-default';
 import { sitemapPlugin } from '@vuepress/plugin-sitemap';
 import { docsearchPlugin } from '@vuepress/plugin-docsearch';
+import { markdownIncludePlugin } from '@vuepress/plugin-markdown-include';
 
 import {
   navbarDe,
@@ -211,6 +212,7 @@ export default defineUserConfig({
   plugins: [
     // Note: gitPlugin, backToTopPlugin, mediumZoomPlugin, and copyCodePlugin
     // are already included in defaultTheme, so we don't need to add them here
+    markdownIncludePlugin({}),
     shikiPlugin({
       themes: {
         light: 'dark-plus',
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -70,6 +70,8 @@ ls | where type == file | sort-by -r size | first 5
 
 The preferred form for consistency is `` ```nu ``.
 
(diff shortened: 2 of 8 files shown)


orthonormalremy:
the people yearn for programmatically trackable cookbook recipes

fdncred:
what people?

fdncred:
i'm ok with this, although there are more scripts than the few you have in this pr probably. @hustcer usually makes most decisions on this repo so i'll leave it up to him.

orthonormalremy:
> what people?

just a meme format lol (see: "my people yearn for freedom")

hustcer:
Thanks, I’m fine with this change.

NotTheDr01ds:
Apologies for missing this, but ...  We used to do this, and we decided to *remove* snippets.  It was a lot of work to remove them across all of the files and translations, and I really don't want to re-introduce this "feature."

Translations being one of the bigger issues.  I know that this is limited to the Cookbook at the moment, and I don't think we have any Cookbook translations.

It's probably my fault for never finishing up #1487 and adding this guidance to Contributing.md ;-)

Also note that the support is already there (see https://github.com/nushell/nushell.github.io/commit/61adb5cf760bd94d6b38e4a2ee884cf66692f3fa), so no need to add a new plugin for it, I don't think?

But regardless, I'd like to consider moving these back to being embedded in the md files themselves.  It's far easier on maintainers to edit all in one doc.  The disadvantages of this change, IMHO, far outweigh any advantages.  The "Copy" button on each code block is what most users will use to make use of the snippet anyway.

Do we really foresee people cloning the entire doc repository just to be able to copy over a few files?  I seriously doubt it.

fdncred:
ya, i forgot that i agreed with @NotTheDr01ds. handling those extra files is a pain. here's the revert https://github.com/nushell/nushell.github.io/pull/2171

orthonormalremy:
Ah sorry I tired to find some prior art around this but I should've looked harder, no worries!

NotTheDr01ds:
@orthonormalremy Not a problem at all!

Again, I consider this my fault for not catching your PR earlier (as well as for not updating the Contribution Guide as I had promised).  I really do appreciate your attempt at improving Nushell here!  Looking forward to future suggestions!

PR #2167, opened 2026-05-20
Document the percent sigil

### LINKED ISSUE #2166: Percent sigil is undocumented
[One](https://www.nushell.sh/blog/2026-05-08-twin0350.html) of the TWiN mentioned an improvement percent sigil, but searching for `%` or `percent` in nushell docs does not yield anything relevant.

samoylovfp:
Fixes https://github.com/nushell/nushell.github.io/issues/2166

Is the example too convoluted? Maybe just
```nu
let cmd = "is-empty"
[ 1 ] | %$cmd
```
?

### DIFF
--- a/book/aliases.md
+++ b/book/aliases.md
@@ -125,3 +125,12 @@ def ls [
     ) | sort-by type name -i
 }
 ```
+
+To call the underlying built-in command you can use a percent sigil `%`, e.g. 
+```nu
+def ls [] {
+    "something else"
+}
+
+%ls # <- calls the original ls
+```
--- a/lang-guide/chapters/strings_and_text.md
+++ b/lang-guide/chapters/strings_and_text.md
@@ -44,7 +44,10 @@ The percent sigin `%` can be used to call the built-in command (without argument
 
 ```nu
 let cmd = "ls"
+# call internal `ls` command
 %$cmd
+# call external `ls` command
+^$cmd
 ```
 
 ## String Quoting
--- a/lang-guide/chapters/strings_and_text.md
+++ b/lang-guide/chapters/strings_and_text.md
@@ -40,6 +40,13 @@ let arguments = ["arg1", "-a", "arg2"]
 
 The caret `^` before the string interpolation symbol `$` allows that external command to be executed.
 
+The percent sigin `%` can be used to call the built-in command (without arguments).
+
+```nu
+let cmd = "ls"
+%$cmd
+```
+
 ## String Quoting
 
 ### Double quotes


Juhan280:
Primary use case for `%` is to call builtins that are overriden by other custom command.
```nushell
alias cd = __zoxide_z
%cd some_dir # can be used to run builtin cd
```
This can also be used for wrapping a builtin command with the same name
```nushell
def ls [] {
  %ls | sort-by size
}
```

lang-guide/chapters/strings_and_text.md
```diff
@@ -40,6 +40,25 @@ let arguments = ["arg1", "-a", "arg2"]
...
 The caret `^` before the string interpolation symbol `$` allows that external command to be executed.
 
+The percent sigil `%` allows running dynamically defined built-in commands.
+
+#### Example
+
+```nu
+
+[[cmd]; ["any"] ["all"]] 
+| insert result {|row| [1 2 3] | %$row.cmd {$in == 2}}
```
fdncred: This isn't a great example of why % exists. Maybe one like Juhan suggests would be better here.

samoylovfp:
Now that I understood the intention, I think I found a better place for this piece of information.

book/aliases.md
```diff
@@ -63,35 +63,9 @@ displaying all listed files and folders in a grid.
 ## Replacing Existing Commands Using Aliases
 
 ::: warning Caution!
-When replacing commands it is best to "back up" the command first and avoid recursion error.
+When replacing commands use the percent sigil `%` to call the original command and avoid recursion error.
```
Juhan280:
```suggestion
When replacing commands, use the percent sigil `%` to call the builtin command and avoid recursion error.
```

% only runs builtins. It doesn't run custom command.

```nushell
def test [a] {
  echo $a
}

def test [] {
  %test b # doesn't work
}
```

samoylovfp: Right. Also returned the line about using aliases for commands as opposed to a builtin.

fdncred:
I think it's fine here but the it shouldn't be the only place because I didn't write the code for aliases. I wrote it to be able to run built-ins from anywhere. So, either that's another special section in another area or it's around where `^` is documented.

samoylovfp:
@fdncred 
I've added mentions of the percent sigil to both the section about overriding commands and one about executing strings.
Does this look better?

lang-guide/chapters/strings_and_text.md
```diff
@@ -40,6 +40,13 @@ let arguments = ["arg1", "-a", "arg2"]
 
 The caret `^` before the string interpolation symbol `$` allows that external command to be executed.
 
+The percent sigin `%` can be used to call the built-in command (without arguments).
+
+```nu
+let cmd = "ls"
+%$cmd
```
fdncred, line 48: I think it's fine but I'd do something like this just to drive the point home.
```nushell
# call internal `ls` command
%$cmd
# call external `ls` command
^$cmd
```

lang-guide/chapters/strings_and_text.md
```diff
@@ -40,6 +40,13 @@ let arguments = ["arg1", "-a", "arg2"]
 
 The caret `^` before the string interpolation symbol `$` allows that external command to be executed.
```
Juhan280, line 41: I know this is not a part of this PR, but since you are already modifying the file, I might as well add this.

I don't think the `$` symbol here is considered string interpolation in this context. But rather, it's a part of the variable. e.g. `$cmd`

Juhan280: Nvm, I didn't expand the diff, and just assumed the wrong thing

lang-guide/chapters/strings_and_text.md
samoylovfp, line 48:
```suggestion
# call internal `ls` command
%$cmd
# call external `ls` command
^$cmd
```

fdncred:
Good enough to begin with. Thanks!

PR #2179, opened 2026-06-08
Update plugin list docs: replace is_running with status field

guluo2016:
The output format of `plugin list` has changed in PR https://github.com/nushell/nushell/pull/14085.  
It now uses the `status` field instead of the `is_running` field.

So, we should update the corresponding documentation accordingly to avoid confusing users who refer to it.


ko/book/plugins.md
```diff
@@ -225,31 +227,31 @@ scope commands | where type == "plugin"
 
 플러그인은 사용 중인 동안 계속 실행되며, 비활성 상태가 일정 기간 지속되면 기본적으로 자동으로 중지됩니다. 이 동작은 [플러그인 가비지 수집기](#plugin-garbage-collector)에 의해 관리됩니다. 플러그인을 수동으로 중지하려면 해당 이름으로 `plugin stop`을 호출하십시오.
 
-예를 들어, 해당 플러그인에서 `gstat` 명령을 실행한 다음 `is_running` 상태를 확인하십시오.
+예를 들어, 해당 플러그인에서 `query` 명령을 실행한 다음 `status` 상태를 확인하십시오.
```
guluo2016, line 230: I updated this document with the help of the AI translation tool because I don't know Korean.  
I checked it again after the update and couldn't find any issues.

fdncred:
Thanks

PR #2183, opened 2026-06-16
FIX: Nushell + Starship example updated to work with latest Nu syntax

TheRealBunbury:
The Nushell overlays + Starship prompt example required changes to work properly with the latest version of Nushell.

The recommended fix has been tested and now works properly.

### DIFF
--- a/cookbook/setup.md
+++ b/cookbook/setup.md
@@ -106,7 +106,7 @@ currently activated.
 ```nu
 # set NU_OVERLAYS with overlay list, useful for starship prompt
 $env.config.hooks.pre_prompt = ($env.config.hooks.pre_prompt | append {||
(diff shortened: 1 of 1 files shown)


cookbook/setup.md
```diff
@@ -106,7 +106,7 @@ currently activated.
 ```nu
 # set NU_OVERLAYS with overlay list, useful for starship prompt
 $env.config.hooks.pre_prompt = ($env.config.hooks.pre_prompt | append {||
-  let overlays = overlay list | slice 1..
+  let overlays = overlay list | where ($it.active == true) | slice 1.. | each {|overlay| $overlay.name }
```
fdncred, line 109: I don't think the parentheses are necessary, are they?

PR #2185, opened 2026-06-18
Document match_description option for custom completions

kiil:
Custom completers can now opt into matching the user's prefix against suggestion descriptions in addition to values, by setting `match_description: true` in the returned `options` record. The inserted completion is still the suggestion's value.

This PR documents the new option in `book/custom_completions.md`:
- Adds `match_description` to the list of completion `options`.
- Adds a "Matching against descriptions" example showing how a value (an email address) can be completed by matching against its description (a person's name).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kiil:
The failing check spelling error is not in my patch.

The spelling error is in another part of the book: 

https://www.nushell.sh/book/line_editor.html#abbreviations

fdncred:
Thanks

PR #2187, opened 2026-06-19
Fix comparison for pwsh and csharp

sharpchen:
There are mistakes for comparison of pwsh and csharp, some are fundamentally wrong.
Sorry I'm not going to list the changes here, please use ai to summarize if you find it hard to review.


book/nushell_operator_map.md
```diff
@@ -3,28 +3,29 @@ next:
...
+| Nushell | SQL      | Python             | .NET (C#)                     | PowerShell            | Bash               |
+| ------- | -------- | ------------------ | ----------------------------- | --------------------- | ------------------ |
+| ==      | =        | ==                 | ==                            | -eq                   | -eq                |
+| !=      | !=, <>   | !=                 | !=                            | -ne                   | -ne                |
+| <       | <        | <                  | <                             | -lt                   | -lt                |
+| <=      | <=       | <=                 | <=                            | -le                   | -le                |
+| >       | >        | >                  | >                             | -gt                   | -gt                |
+| >=      | >=       | >=                 | >=                            | -ge                   | -ge                |
+| =~      | like     | re, in, startswith | Regex.IsMatch                 | -cmatch               | =~                 |
```
fdncred: I don't really care for all your changes here because you removed the LINQ items entirely. There's a reason LINQ was in the column header.

fdncred (changes requested):
Most of the changes seem good. However, I don't like removing the LINQ parts.

book/nushell_operator_map.md
sharpchen: I removed it because it had nothing to do with linq, these comparisons are mistakes

fdncred: I disagree. If you use c# enough with datasets, you'll probably be using linq. so, i think it's helpful for some to understand it here. maybe there's a different way to include both. i'm not sure what it is though.

fdncred: you'll also need to fix the typo and rebase on main

sharpchen: The only relevant here is `Except` which is still wrong because `Except` returns enumerable not boolean, making linq irrelevant here.
`Contains`, `StartsWith`, `Power` and all these operators none of them is of linq convention

fdncred: in c# with an ienumerable you can call methods like `Contains` below and many other methods.
```csharp
string[] fruits = { "apple", "banana", "mango", "orange", "passionfruit", "grape" };
string fruit = "mango";
bool hasMango = fruits.Contains(fruit);
// or
var result = fruits.Where(f => f.StartsWith("Ap")); 
```

sharpchen: How is it relevant to linq? It can have arbitrary predicate. We should mention `str.StartsWith` instead, and the original doc doesn't even mentioned `Where`.
Although my change does includes `Any` as well, but most in the list are not collection operation except `in` and `not-in`, why is linq considered worthy here in the first place?
I know csharp pretty well, my first glance to this doc is just confusing, even misleading.
Maybe just get rid of the whole linq column would be better

fdncred: If you look at linq and how nushell works with filters there are quite a few similarities. Certain parts of nushell are inspired by c# and linq. In fact, the reason why string interpolation exists as it does now is because of c#.

sharpchen: Interesting story. If to appreciate the inspiration you can have a dedicated page like `Coming from C#` under filter chapter or simply note on the filter doc.
I understand all these similarities, it's definitely not a proper place to put linq in `nu_operator_map` as the listed are mostly binary operation, not collection wise.
It's not helpful(and wrong) for csharp dev getting into nushell at all.

PR #2195, opened 2026-07-07
Update 0.114 release notes for structured MCP output

Alb-O:
Replace the highlighted nu-mcp escaping note by fdncred with the structured evaluate output change from [#18499](https://github.com/nushell/nushell/pull/18499) which supersedes it


fdncred:
thanks

PR #2199, opened 2026-07-11
Fix path type code demonstration

### LINKED ISSUE #2155: Documentation on type path doesn't match behaviour
# String won't be expanded into a fully pathname when passed as an arguments with type of path

## Document

Doc's here [Nushell Type Path](https://www.nushell.sh/lang-guide/chapters/types/other_types/path.html#additional-language-notes)

```nu
def show_difference [
 p: path
 s: string
] {
 print $"The path is expanded: ($p)"
 print $"The string is not: ($s)"
}

# Results
cd ~/testing
show_difference . .
# => The path is expanded: /home/username/testing
# => The string is not: .
```

## Actually

```nu
show_difference . .
# => The path is expanded: .
# => The string is not: .

nu --version
# => 0.112.2
```

You wouldn't be surprised. Code changes fast around here. Just a reminder
I like nushell, thanks for all works of yours

Kissaki:
`.` does not get expanded

tested in current 0.114.0 and reported in #2155 for 0.112.2

Resolves #2155

On Windows `...` gets expanded to `..\..`. I have not tested on Linux, so I leave the `...` example as-is.

### DIFF
--- a/lang-guide/chapters/types/other_types/path.md
+++ b/lang-guide/chapters/types/other_types/path.md
@@ -12,7 +12,7 @@
 
 1. `path` is technically a "syntax shape" rather than a full "type".
    It is used for annotating strings that should be treated as a path to a filename or directory.
-   `~` and `.` characters in the string will automatically be expanded treated as a `path`.
+   `~` and `...` characters in the string will automatically be expanded treated as a `path`.
 
    Example:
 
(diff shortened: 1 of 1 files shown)


fdncred:
I wonder if it's a bug that the path isn't expanded with `.` and `path`

Juhan280:
`.` not expanded is an intended behavior, nushell used to expand them. But a lot of programs makes a distinction between relative and absolute paths, so the behavior was changed
- https://github.com/nushell/nushell/pull/15878

fdncred:
According to that PR `~, ., and ..+ are still expanded`, which is not the case any longer

Juhan280:
It expands, but keeps the relative paths relative, unless I am reading it wrong

PR #2204, opened 2026-07-11
Fix code fencing

Kissaki:


fdncred:
conflict here. maybe save up a bunch of these for 1 PR instead of a PR for each small change.

Kissaki:
IIRC stuff gets squashed here which is why I prefer individual specific ones. Specifically for separate concerns, also to streamline reviews.

Kissaki:
I was aware that the conflict would show up. It's a separate line, though. My diff editor auto-resolved it.

I rebased it. The conflict is resolved.

fdncred:
Thanks

PR #2206, opened 2026-07-16
Fix pubDate not being added to RSS entries

### LINKED ISSUE #2103: Feed repeatedly lists old items as new
For many weeks now, the [RSS feed](https://www.nushell.sh/rss.xml) repeatedly reports a bunch of old items as new, presumably when a new post is published.

_Some exploration, context, references:_

PR #1780 fixed an issue early last year regarding the filename inferred date sorting.

What changed?

Looking into recent config changes, #2075 dropped plugin references from the config, presumably fixing redundant declarations. Could that be the issue, causing the plugin configuration from the site rather than theme no longer being applied?

ninjack-dev:
Resolves #2103. I went for the simplest possible fix which just uses the parsed date from the file name; since this lacks an actual time, the RSS feed shows new entries as being published at midnight. A more granular fix would be somewhat hacky (e.g. parsing the Git commit metadata on a per-file basis, or requiring that article writers specify `date` in the frontmatter), so I'd just leave it as-is for now.

Confirmed with:
```nu
def getTitlePubDate [] {
  $in | get content.content.0
      | where tag == "item"
      | each {|it|
          {
            title: ($it.content | where tag == "title" | first 1 | get content.0.content.0)
            pubDate: ($it.content | where tag == "pubDate" | first 1 | get content.0.content.0)
          }
        } | first 3 | to md 
}
```
Running `open .vuepress/dist/rss.xml | getTitlePubDate`:

| title | pubDate |
| --- | --- |
| Nushell 0.114.1 | Sat, 11 Jul 2026 00:00:00 GMT |
| Nushell 0.114.0 | Sat, 04 Jul 2026 00:00:00 GMT |
| This week in Nushell #356 | Fri, 19 Jun 2026 00:00:00 GMT |

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -279,6 +279,9 @@ export default defineUserConfig({
           frontmatter.feed === true || filePathRelative?.indexOf('blog/') >= 0
         );
       },
+      getter: {
+        publishDate: (page) => new Date(page.date),
+      },
       sorter: (a, b) => {
         const pathDateA = new Date(
           a.path.replace('/blog/', '').substring(0, 10),


fdncred:
Thanks

PR #2207, opened 2026-07-17
Fix incorrect update date overriding publish date in Atom feed

ninjack-dev:
Addendum to #2206. While that fix did correctly set `pubDate` in RSS feeds, it didn't set the `update` property for Atom entries, only `published`, which is not used by feeds for ordering.


fdncred:
one more try. thanks

PR #2211, opened 2026-07-24
Document passing arguments to external commands

### LINKED ISSUE #527: Document passing arguments to externals better
I've seen a lot of issues about external arguments, like this one: https://github.com/nushell/nushell/issues/5947.

We should have a page that documents all the ways you can call external with arguments (i.e., explaining that you can pass a list to an external as a list of arguments and that one long string with spaces will be considered as one argument).

ColumbusLabs:
## What

Document how Nushell passes arguments to external commands.

## Why

The existing page explains how to select an external command, but not how quoted strings, lists, or dynamic executable names map to the invocation.

Fixes #527

## Details

- explains that quoted spaces stay inside a single argument
- makes the direct-list error explicit and shows `...$paths` for separate arguments
- documents executable names stored in variables with `^$program`
- links the canonical spread, string-command, and `extern` documentation

## Testing

- `npm ci`
- `npm run build` (1,628 pages rendered)
- `npx prettier --check book/running_externals.md`
- rendered-link and anchor inspection
- `git diff --check`

### DIFF
--- a/book/running_externals.md
+++ b/book/running_externals.md
@@ -18,6 +18,32 @@ External command (typically `/usr/bin/ls`):
 On Windows, `ls` is a PowerShell _alias_ by default, so `^ls` will not find a matching system _command_.
 :::
 
+## Passing Arguments
+
+External arguments are separated by Nushell syntax, not by spaces inside a quoted string. Quote a string to keep its spaces inside one argument:
+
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #2216, opened 2026-07-30
Add llms.txt for AI/LLM discoverability

abyworkings-coder:
## What

Adds an `llms.txt` file at the site root (served from `.vuepress/public/llms.txt`, following the same static-passthrough convention as `opensearch.xml` and `icon.png`), giving AI tools and LLM-based assistants a curated, low-noise index of the documentation.

## Why

[llms.txt](https://llmstxt.org/) is an emerging convention (like `robots.txt` or `sitemap.xml`) for giving LLM-based tools a curated entry point into a site's docs, rather than requiring them to crawl or guess through the full site. Given nushell.github.io has 700+ individual command pages and 450+ blog posts, an exhaustive dump wouldn't be useful — this file instead curates the highest-value ~80 pages: getting-started guides, core language concepts, scripting/extensibility docs, configuration, the commands index and category pages, the cookbook, and the language guide.

## What's included

- Getting Started (installation, quick tour, fundamentals)
- Migration guides (coming from bash/PowerShell/CMD)
- Core language concepts (types, variables, control flow, tables, records, lists, strings)
- Scripting & extensibility (hooks, aliases, completions, plugins, dataframes, testing)
- Configuration & environment
- Commands reference (index + key categories, not all 700+ individual pages)
- Cookbook and language guide indexes
- Optional: contributor book, contributing guide

Verified every linked URL returns 200 on the live site before opening this PR.

## Test plan

- [x] Confirmed no existing `llms.txt` (404 before this change)
- [x] Confirmed `.vuepress/public/` is the static-passthrough directory used for other root-level static files (`opensearch.xml`, `icon.png`), both live at 200
- [x] Spot-checked all major linked paths (book, commands, cookbook, lang-guide, blog indexes, category pages) return 200 live


fdncred:
I generally don't have a problem with this but I look at it and think it's just one more thing to keep updated when the site changes. Hopefully someone can keep up with that.

PR #2223, opened 2026-08-13
convert the direnv export PATH to a list before loading it into env

orthonormalremy:
I think it's cleaner to never load it into the env as a string in the first place


fdncred:
Thanks

PR #2228, opened 2026-08-16
Fix to yaml examples 0115

maxim-uvarov:
Fix two to yaml examples in the 0.115.0 release notes
Both snippets do not match what Nushell 0.115.0 actually does.

`--non-roundtrip` takes a string, so a bare `null` is parsed as the
`null` literal and the line fails to parse:

    Error: nu::parser::parse_mismatch_with_full_string_msg
      x Parse mismatch: expected string.

Quoting it makes the example run and print `null` as documented.

The `--indent 4 --quote double` example showed output with 2-space
indentation and single quotes. The real output on 0.115.0 is:

    outer:
        inner: "value"

which is also what the two flags are there to demonstrate.


cptpiepmatz (approved):
Lol, I was writing these examples by hand. Thanks for catching these.

PR #2237, opened 2026-09-04
docs: fix broken heading anchors

wahidrizka:
Five internal links point at heading anchors that do not exist, so they land at the top of the page instead of the intended section. Each target below was confirmed against the deployed site.

| File | Link anchor | Target heading | Correct anchor |
| --- | --- | --- | --- |
| `book/navigating_structured_data.md` | `#keycolumn-names-with-spaces` | `## Key/Column names with spaces` | `#key-column-names-with-spaces` |
| `lang-guide/.../basic_types/float.md` | `#floatsdecimals` | `### Floats/Decimals` | `#floats-decimals` |
| `lang-guide/.../basic_types/string.md` | `#textstrings` | `### Text/Strings` | `#text-strings` |
| `book/control_flow.md` | `#variables-are-immutable` | `## Variables are Immutable by Default` | `#variables-are-immutable-by-default` |
| `book/coming_from_bash.md` | `#editing-mode` | `## Setting the Editing Mode` | `#setting-the-editing-mode` |

The first three are slug mismatches: a heading with a `/` produces a `-` in the generated slug (for example `Key/Column` becomes `key-column`), but the link dropped that dash. The last two point at headings that were later renamed. Each target page has exactly one matching section, so the intended destination is unambiguous.


fdncred:
Thanks

PR #2239, opened 2026-09-09
docs: fix duplicated words

wahidrizka:
Four spots in the docs repeat a word by accident. This removes the extra word in each:

- `book/custom_completions.md`: "only the the custom command" to "only the custom command"
- `book/dataframes.md`: "the unique unique values" to "the unique values"
- `book/dataframes.md`: "into into nushell value" to "into nushell value" (the `polars into-nu` row)
- `lang-guide/chapters/types/type_signatures.md`: "the same as the the existence" to "the same as the existence"

Text only, no behavior or command output changes.


fdncred:
Thanks

PR #2244, opened 2026-09-17
Include `$ans` in Special Variables

pimgit0:
Added a short section to the `Special Variables` page about the recently-added `$ans` variable, re:
> Many of these are mentioned or documented in other places in this Book, but this page should include all variables for reference.


fdncred:
We may need to look at this again. It would be nicer to "just work" with externals but we can proceed with this for now. Thanks!

pimgit0:
I imagine it would be hard to balance between what kind of external command outputs should and shouldn't overwrite in a way that feels natural to the end-user.

fdncred:
indeed. when writing it i remember there being issues of things like fzf hanging and fighting over who owns the terminal. not sure it can be easily fixed.

